PR #1, opened 2020-08-13
adding application architecture adr

boecklic:


ltclm:
thank you @boecklic this kind of metadata should be mandatory for all our github repositories.

adr/2020_08_13_application_architecture.md
```diff
@@ -0,0 +1,20 @@
...
+- Programming Language **Python**: Python is used in most of the backend services and is the programming language that's best known within devs, so no reason to change here.
+- Application Framework **Django**: Django is used as application framework. Django is very majure and has a wide user community. It comes with excellent documentation and a powerfull ORM. Futhermore, there are well-supported and maintained extensions, e.g. for designing REST API's that can considerably reduce the amount of boilerplate code needed for serializing, authentication, ....
+- Asset Storage **S3**: Since this service runs on AWS assets will be stored in S3 object store.
+- Metadata Storage **PostGIS**: Since Metadata can contain arbitrary geoJSON-supported geometries, Postgres along with the PostGIS extension is used as storage for metadata.
+
+The application architecture does initially only involve syncronous operations. Async tasks will be reconsider once certain write operations don't meet performance requirements anymore.
+
+## Consequences
+Developers not familiar with Django will have to have through the Django tutorial before they get started with development.
```
hansmannj, line 20: just two minor comments:
"have to have the ..." could be changed to "have to take the ..."

And how about adding a link to the Django tutorial, even if it may change from time to time?
e.g. https://docs.djangoproject.com/en/3.1/intro/

boecklic:
> thank you @boecklic this kind of metadata should be mandatory for all our github repositories.

thanks for the comment. I think so too, therefore I included it in the guidelines: https://github.com/geoadmin/doc-guidelines/blob/master/ARCHITECTURE.md#architecture-decision-records-adr 
Global decisions should go there, app-specific decision in the repo. Next global decision will be regarding Aurora/RDS I guess

rebert (approved):

PR #3, opened 2020-09-07
Added 'checker' endpoint, changed landing page to '/api/stac/v0.9'

hansmannj:
- Changed the landing page to be found under "api/stac/v0.9"
- added a simple "checker"-endpoint

### DIFF
--- a/project/app/urls.py
+++ b/project/app/urls.py
@@ -4,4 +4,5 @@ from . import views
 
 urlpatterns = [
     path('', views.index, name='index'),
+    path('checker/', views.checker, name='checker'),
 ]
--- a/project/app/urls.py
+++ b/project/app/urls.py
@@ -3,6 +3,6 @@ from django.urls import path
 from . import views
 
 urlpatterns = [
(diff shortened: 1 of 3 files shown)


ltshb (commented):
Not sure about the checker endpoint should not it be `data.geo.admin.ch/checker` instead of `data.geo.admin.ch/api/stac/v0.9/checker` ? It has nothing to do with the STAC therefore I would no add it to the STAC path. @procrastinatio, @rebert, @boecklic, @ltkum what are your opinion on this ?

rebert (changes requested):
I agree with @ltshb. Beside LGTM

hansmannj:
makes sense, I've moved the checker endpoint to the root

hansmannj:
something went wrong, my changes were overwritten. I'll push again, soon

ltshb (approved):
We might want in future to split the views but for the moment I think it is fine to have it in one file as it only have the checker.

rebert (approved):

PR #5, opened 2020-09-08
Added missing dependencies in docker file

ltshb:
Django project requires some apt package dependencies in order to run.
See https://docs.djangoproject.com/en/3.1/ref/contrib/gis/install/geolibs/


rebert (approved):
shall we add it to the make setup as well? or can it be considered to be a part of a standard dev machine

PR #7, opened 2020-09-09
DEBUG and ALLOWED_HOSTS configuration + minor makefile improvement

ltshb:
Made the `DEBUG` and `ALLOWED_HOSTS` flags configurable via environment variables. 
The `DEBUG` flag is per default to `False` and set to `True` in the makefile for local development.
The `ALLOWED_HOSTS` will be then set to the DEV, INT or PROD host name by k8s.

Also improved the makefile with usefull targets for local development.

### DIFF
--- a/project/project/settings.py
+++ b/project/project/settings.py
@@ -27,6 +27,10 @@ SECRET_KEY = '%5+eq2851!d7qi^sze(nv2g#kt8v$7)4ck3cq*e!5c2rx%13p+'
 DEBUG = bool(strtobool(os.getenv('DEBUG', 'False')))
 
 ALLOWED_HOSTS = []
+if DEBUG:
+    # When the debug flag is set allow local host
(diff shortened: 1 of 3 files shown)


project/project/settings.py
```diff
@@ -23,9 +24,13 @@
 SECRET_KEY = '%5+eq2851!d7qi^sze(nv2g#kt8v$7)4ck3cq*e!5c2rx%13p+'
 
 # SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = True
+DEBUG = bool(strtobool(os.getenv('DEBUG', 'False')))
 
 ALLOWED_HOSTS = []
+if DEBUG:
```
rebert, line 30: Isn't that the default behaviour? https://docs.djangoproject.com/en/3.1/ref/settings/#allowed-hosts

rebert (approved):
@ltshb I tested the Makefile and it seems to work

PR #12, opened 2020-09-16
Changed folder structure

ltshb:
Renamed the `project/project` to `project/config` which makes more sense
as it contains configuration files. Also moved the logging-cfg-*.yml
files in the new config folder.


hansmannj (approved):
Hi Brice

I see your point in renaming the inner `project` folder to `config`.
The Django definition of `project`-folder is:  
>**Project**\
>*A Python package – i.e. a directory of code – that contains all the settings for an instance of Django. This would include database configuration, Django-specific options and application-specific settings.*  

So basically, the folder, that we want to rename to config, would meet this criteria (just to have mentioned that :wink:.
  
So, in order to create the folder that is now going to be named `config` from scratch, one would have to use:\
`$ django-admin startproject **config**`, which would lead to:  

```
**config**/
    manage.py
    **config**/
        __init__.py
        settings.py
        urls.py
        asgi.py
        wsgi.py
```  


The way this project was started, was using:\
`$ django-admin startproject **project**`.
```
**project**/
    manage.py
    **project**/
        __init__.py
        settings.py
        urls.py
        asgi.py
        wsgi.py
```  


In both cases, the name of the outer and inner folder are not the ideal combination. I think, your suggestion makes sense. An outer **project** and an inner **config**.
One could think of renaming the outer **project** to something more meaningful, as its name does not matter to Django:
> "ts name doesn’t matter to Django; you can rename it to anything you like.")  

Same for the **app** folder: I think it makes sense to rename that to something like "stac-API", as we discussed this morning. We could then add more apps, if needed, e.g. "stac-browser", et al.

ltshb:
@hansmannj  would it make sense to rename also the `app` folder to `stac-api` within this PR ?

hansmannj:
@ltshb yes, that's a good idea!

PR #13, opened 2020-09-17
BGDIINF_SB-1317: corrected the landing page

ltshb:
The `links` attribute was missing in the landing page. The description has
also been updated as required by management.

Also renamed the landing page function called `index` to `landing_page`
which is the name of the page as defined in the ticket and spec.


rebert (approved):
I like it

hansmannj:
makes things a bit clearer

PR #14, opened 2020-09-21
Added django migrate and django check in makefile

ltshb:
It seems that the `manage.py migrate` command will be often used,
therefore add it as makefile target to simplify its uses.

Also added the `manage.py check` command to the lint target. This
command is kind of a django linting command as it does some static code
checking (see https://docs.djangoproject.com/en/3.1/ref/checks/). To
start with the WARNING level but this could change in the future as we
gain experience with django.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -109,7 +109,8 @@ format: $(DEV_REQUIREMENTS_TIMESTAMP)
 
 
 .PHONY: lint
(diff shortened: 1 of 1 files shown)


hansmannj (approved):
Thanks Brice :+1:

PR #16, opened 2020-09-25
Added isort to make format target

ltshb:
Added isort as mentioned in our geoadmin/doc-guideline and formated the
code to follow the import sorting.


hansmannj (approved):
:+1:
Thanks Brice!

PR #17, opened 2020-09-25
model definitions for collections, items and assets

hansmannj:
This is pretty much work in progress. Did not really test anything so far, so I intentionally did not add reviewers to this PR yet.
Things need to further evolve first.
Just wanted to have a backup of my local dev work on the one hand, and on the other hand also make it possible for someone else to continue working on it ;-)


project/stac_api/models.py
```diff
@@ -1,28 +1,307 @@
...
+import re
 
 # pylint: disable=fixme
 # TODO remove this pylint disable once this is done
 
 # st_geometry bbox ch
 BBOX_CH = 'POLYGON((2317000 913000,3057000 913000,3057000 1413000,2317000 1413000,2317000 913000))'
 
+# I allowed myself to make excessive use of comments below, as this is still work in progress.
+# all the comments can be deleted later on
```
boecklic, line 14: NOPE! I like excessive comments!! We should have more of them in general. It makes collaboration it much easier, specially when you look at older code

project/stac_api/models.py
```diff
@@ -1,28 +1,307 @@
...
+
+# For Collections and Items: No primary key will be defined, so that the auto-generated ones
+# will be used by Django. For assets, a primary key is defined as "PositiveBigInteger" due the
```
boecklic, line 17: as mentioned: default case in django is not to explicitly handle the primary key and leave this to django. In case `models.AutoField` is not large enough (large number of objects excpected), `id = models.BigAutoField(primary_key=True)` can be defined

project/stac_api/models.py
```diff
@@ -1,28 +1,307 @@
...
+    crs = models.URLField(default=["http://www.opengis.net/def/crs/OGC/1.3/CRS84"])  # [string]
+    created = models.DateTimeField(auto_now_add=True)  # datetime
+    updated = models.DateTimeField(auto_now=True)  # datetime
+    description = models.TextField()  # string  / intentionally TextField and
+    # not CharField to provide more space for the text.
+    # TODO: ""description" is required in radiantearth spec, not required in our spec
+
+    # temporal extent will be auto-populated on every item update inside this collection:
+    # start_date will be set to oldest date, end_date to most current date
+    start_date = models.DateTimeField(blank=True)  # will be automatically populated
```
boecklic, line 31: it's probably necessary to have `null=True` as well, you have to create a collection before you add any items, in that phase the start_date and end_date will be `null`

project/stac_api/models.py
```diff
@@ -1,28 +1,307 @@
...
+    SW = ArrayField(models.FloatField(), blank=True)  # [Float], auto-populated
+    NE = ArrayField(models.FloatField(), blank=True)  # [Float], auto-populated
+
+    collection_name = models.CharField(unique=True, default="")  # string
+    # collection_name is what is simply only called "id" in here:
+    # http://ltboc.infra.bgdi.ch/static/products/data.geo.admin.ch/apitransactional.html#operation/createCollection
+    itemType = models.CharField(default="Feature")  # string
+
+    # The field formerly known as "Keywords" has been "sourced out" to a
+    # separate Keyword class with many-to-many relations to the Collection objects
```
boecklic, line 59: Imho it's clearer to define the manytomany field here:
```
class Keywords(models.Model):
    name = models.CharField(max_length=64)

class Collection(models.Model):
    ...
    keywords = models.ManyToMany(Keywords)
```
this would allow to reuse the keyword model for M2M in e.g. an Item model

project/stac_api/models.py
```diff
@@ -1,28 +1,307 @@
...
+    stac_extension = ArrayField(
+        models.CharField(),
+        default=[
+            'eo',
+            'proj',
+            'view',
+            'https://data.geo.admin.ch/stac/geoadmin-extension/1.0/schema.json'
+        ],
+        editable=False
+    )
```
boecklic, line 81: nice!

project/stac_api/models.py
```diff
@@ -1,28 +1,307 @@
...
+        # TODO: move this check to the items save()
+        if self.start_date is None and self.end_date is None:
+            raise ValidationError(_('At least a start date or an end date has to be defined.'))
+
+        # very simple validation, raises error when geoadmin_variant strings contain special
+        # characters or umlaut.
+        for variant in self.geoadmin_variant:
+            if not (bool(re.search('^[a-zA-Z0-9]*$', variant)) == True):
+                raise ValidationError(_('Property geoadmin:variant not correctly specified.'))
+
```
boecklic, line 111: just for Info: cleaning of individual fields (e.g. the `geoadmin_variant`) can also be done like this
```
def clean_geoadmin_variant(self):
    ....
```

project/stac_api/models.py
```diff
@@ -1,28 +1,307 @@
...
+            raise ValidationError(_('Bounding box incorrectly defined.'))
+
+    def save(self, *args, **kwargs):
+        #TODO: on every item update:
+        # 1. re-calculate the envelope over all bboxes of all items inside this collection
+        #    and update the collections' bbox accordingly
+        #
+        # 2. loop over all items inside the collection and set the "start_date" of
+        #    the collection to the oldest date found in item dates. Same for
+        #    collections' end_date (set to most recent date found in items)
```
boecklic, line 193: Not sure it's necessary to always loop over all items (this will be a performance nightmare): Imho it's sufficient to just compare the current item with the values stored on the collection level, and adapt if necessary (only thing to consider is to do proper row locking on the collection object (or use `update()`)).

project/stac_api/models.py
```diff
@@ -1,28 +1,307 @@
...
+    # using BigIntegerField as primary_key to deal with the expected large number of assets.
+
+    # alter save-function, so that the corresponding collection of the parent item of the asset
+    # is saved, too.
+    def save(self, *args, **kwargs):
+        self.collection = self.feature.collection
+        super().save(*args, **kwargs)
+
+    # using "_name" instead of "_id", as "_id" has a default meaning in django
+    asset_name = models.CharField(unique=True, blank=False)
```
boecklic, line 220: you'll have to define `max_length` for char fields

project/stac_api/models.py
```diff
@@ -1,28 +1,307 @@
...
+    def clean(self):
+        # very simple validation, raises error when geoadmin_variant strings contain special
+        # characters or umlaut.
+        for variant in self.geoadmin_variant:
+            if not (bool(re.search('^[a-zA-Z0-9]*$', variant)) == True):
+                raise ValidationError(_('Property geoadmin:variant not correctly specified.'))
+
+
+class Keyword(models.Model):
+    collections = models.ManyToManyField(Collection)
```
boecklic, line 256: see above

project/stac_api/models.py
```diff
@@ -1,28 +1,307 @@
...
+    title = models.CharField(blank=True)  # string
+
+
+class Link_Item(models.Model):
+    items = models.ManyToManyField(Item)
+    href = models.URLField()  # string
+    rel = models.CharField()  # string
+    link_type = models.CharField(blank=True)  # string
+    # added link_ to the fieldname, as "type" is reserved
+    title = models.CharField(blank=True)  # string
```
boecklic, line 307: see comment above about M2M, 
I would just define one model `Link`, and have M2M fields for `links` in `Collection` and `Item`

boecklic (changes requested):
Nice, thanks! I made a few comments, and I would comment out model fields that are unclear yet, it's easier to add stuff (specially in context of DB schema) than to change/remove.

project/stac_api/models.py
hansmannj: BigAutoField is used for asset_id, I adapted the comment

project/stac_api/models.py
hansmannj: That's correct, I added null=True

project/stac_api/models.py
hansmannj: Good point, Thanks! I changed that

project/stac_api/models.py
hansmannj: Thanks for the info.
For the moment I will stick with the general clean() as I guess there will probably be checks for more than one field per class.
If that will not be the case in the end, we can still change the general clean() to a field specific one.
Or would it make sense to do that from the beginning due to better performance?

project/stac_api/models.py
hansmannj: That's correct. Looping over all items would be very cost intensive and does not make sense. It's enough to only check the currently added item. I will adapt it this way.

project/stac_api/models.py
hansmannj: Adapted that way

project/stac_api/models.py
hansmannj: done

hansmannj:
Thanks for the comments, I tried to implement them and also updated the save() functions.
There is some `TODO`s left. E.g. in one save() function a check needs to be done, if the collections' bbox needs to be updated. This is probably best done with GeoDjango, so I left it open for the moment and probably @rebert can have a look at that.

I'd be glad if we can merge this "monster" now into develop, once the checks pass (hopefully)

hansmannj:
I noticed that the logic behind the update of the start_ and end_date of the collection inside the asset's save function could be improved. Some cases are missing. Furthermore I guess that the collection's save function needs to be called, if it was updated elsewhere (e.g. an asset's save function).

I will implement these changes soon, probably this evening.

boecklic:
A general remark: it might be an idea to split the models into separate apps. This has mainly advantages that it will be easier for development, in particular regarding migrations. Migrations are per-app, so if someone is changing sth in the Collection model, and someone else in item, this will require manual migration resolution. If the models are in separate apps, migrations are separated as well, so there's less chance for conflicts... t.b.d.

rebert (approved):
as discussed with @boecklic, we could devide the model into three apps later on

rebert:
@boecklic sobald du deinen Segen dazu gegeben hast, können iwr @hansmannj erlösen - und diesen Branch mergen :-)

hansmannj:
> I noticed that the logic behind the update of the start_ and end_date of the collection inside the asset's save function could be improved. Some cases are missing. Furthermore I guess that the collection's save function needs to be called, if it was updated elsewhere (e.g. an asset's save function).
> 
> I will implement these changes soon, probably this evening.

OK, I added the explicit calls of the (grand)parent's save-function.
After re-checking the logic of the start_ and end_date updating, I did not find a missing case, so I hope it is correct as it is.

hansmannj:
> A general remark: it might be an idea to split the models into separate apps. This has mainly advantages that it will be easier for development, in particular regarding migrations. Migrations are per-app, so if someone is changing sth in the Collection model, and someone else in item, this will require manual migration resolution. If the models are in separate apps, migrations are separated as well, so there's less chance for conflicts... t.b. discussed

Sounds like a good idea to me, we should consider this. Probably opening a new branch for that once this is merged?

boecklic (approved):
good to merge. we can discuss about splitting next week

PR #20, opened 2020-10-01
fixed issues caused by renaming

hansmannj:
fixed some issues that were caused by renaming the app `project` to `stac_api`


hansmannj:
Don't know why checks are failing. I needed to do these changes in order to be able to test with a local sqlite3 db. Without those changes, the admin pages cannot be created..

ltshb:
I would suggest to first try to update the branch

hansmannj:
Currently the tests are failing because a sqlite db is used for testing, which cannot work, as a postgres db is required in order to use ArrayFields..

boecklic:
We should be careful with merge strategy in git: merging `develop` into a feature branch is often causing troubles. I propose we always **rebase** feature branches on `develop`, if upstream changes in `develop` are required in the feature branch. This keeps commit history in feature branches clean.

boecklic:
and yeah, I'll try to fix remaining points in #22 asap so we can merge

ltshb:
> We should be careful with merge strategy in git: merging `develop` into a feature branch is often causing troubles. I propose we always **rebase** feature branches on `develop`, if upstream changes in `develop` are required in the feature branch. This keeps commit history in feature branches clean.

I agree with this, currently we don't have any guideline regarding this and I started on (https://github.com/geoadmin/doc-guidelines/blob/feature_pr_workflow/PR_WORKFLOW.md) that we should all together discuss and agree.  Guideline are only useful if everyone stick to them :-)

boecklic (approved):
perfect, thnx!

hansmannj:
Checks were still failing. I added the initial migrations. Perhaps that helps.
Test with a local db work now. Even admin pages are up and running with local server and local db :-)

hansmannj:
finally :-)

PR #21, opened 2020-10-07
[WIP] admin user for django

rebert:


scripts/sql/create_users.sql
```diff
@@ -0,0 +1,2 @@
+DROP ROLE if exists service_stac_admin;
+CREATE ROLE service_stac_admin WITH INHERIT CREATEROLE CREATEDB LOGIN NOREPLICATION NOBYPASSRLS PASSWORD 'md5fae2881528025424e07c958ec3cbf595';
```
boecklic, line 2: not sure we should even have md5 hashes of PW in a public repo in github.

boecklic (changes requested):

PR #22, opened 2020-10-07
simplify app setup and local development

boecklic:
I propose to make some changes to simplify the app- and local setup for development and keep consistency with other services. 

**Disclaimer:**
This PR is heavily influenced and biased by experiences and learning from many previous Django projects and is by no means an offense regarding the work done so far. 

For development in general I would try to stick as closely as possible to the django way of doing stuff (e.g. using ./manage.py directly, without a `Makefile` wrapper for everything). Historically, development at igeb-b has taken place on shared, remote machines, using shared resources. This has changed for the good: we have own *NIX machines, so we can develop locally with local resources (DB et al.). This allows in my opinion to simplify a lot of stuff we carry on from older projects.

Summary: 
- local development basically relies on having your machine properly set up (python and stuff)
- local development should basically be done using directly `./manage.py ...` commands
- docker is used for CI (docker-compose.yml) and production only, even docker push is imho not necessary

In particular, I have:
- renamed `project/` to `app/` to be consistent with other services
- radically simplified `Makefile`: this is up for debate. In my opinion with don't need the whole timestamp magic anymore. I like Makefiles for the slightly more complicated targets that you don't want to remember. But I would rely on Django's `manage.py` where possible and other conventions (e.g. project src is always named `app` (even if this collides with Django's concept of projects and apps), we could have chosen `src` instead of `app` everywhere, but this is probably too much work). I would even go as far as e.g. having a `make serve` target (since it exists in all projects) but just print a comment on what to use. For the moment I just commented stuff out. Depending on reviews I'll remove completely.
- added a file `.env.default` that can be copied and sourced to `.env.local` by every dev to quickly set the necessary env variables (this also means I propose to not use summon for local development, this seem over-complicated to me, I think local DB credentials should stay local)
- splitted the `settings.py` into a `settings_prod.py` and `settings_dev.py`. `settings_dev.py` inherits everything from prod and overwrites for local development, every dev has to add the `settings.py` manually, which in the simplest case contains just a `from .settings_dev import *`, but provides a simple way to overwrite settings for local development.
- `dev_requirements.txt` inherits the `requirements.txt`
- removing stuff about sqlite in settings, service requires postgres for development as well. Adding short desc about how to setup a postgres db

I suggest to run tests in codebuild with a local PG instance, otherwise we'll run into issues with migrations. Therefore I
- created a multistage Dockerfile, with a target `test` and a target `production`
- adapted the buildspec.yml and docker-compose.yml


hansmannj:
Makes sense, thanks!
There could be a potential problem in the `settings_prod.py` I guess.
The sqlite3 DB for testing will not support ArrayFields that are used in the current version of the `models.py` hopefully to be merged soon..

.pylintrc
```diff
@@ -10,15 +10,15 @@ fail-under=10
 
 # Add files or directories to the blacklist. They should be base names, not
 # paths.
-ignore=CVS
+ignore=CVS,settings.py
```
ltshb, line 13: Why adding `settings.py` to the pylint ignore ? I would keep it as well in the linter, if there is a good reason to not lint it, then I would suggest to comment it in the pylintrc

Makefile
```diff
@@ -95,79 +98,87 @@ help:
...
 
 
 .PHONY: setup
-setup: $(REQUIREMENTS_TIMESTAMP)
-
+setup:
+	test -d $(VENV) || ( $(SYSTEM_PYTHON) -m venv $(VENV) && $(PIP) install --upgrade pip setuptools; \
+	$(PIP) install -U pip wheel; \
+	$(PIP) install -r $(REQUIREMENTS_DEV); )
+	test -d app/config/settings.py || echo "from .settings_dev import *" > app/config/settings.py
```
ltshb, line 117: The issue with this is that if we change the `requirement.txt` file (also remotely changed by a git pull) the requirements are not updated unless we run `make clean setup` of manually call `pip install`. In my opinion using here the timestamp way takes advantage of the makefile syntax and is more appropriate. It is also more user friendly

README.md
```diff
@@ -33,8 +33,71 @@
 
 ### Dependencies
 
-The **Make** targets assume you have **bash**, **curl**, **tar**, **gopass**, **summon**, **gopass summon provider**, **docker** and **docker-compose** installed.
+<!--The **Make** targets assume you have **bash**, **curl**, **tar**, **gopass**, **summon**, **gopass summon provider**, **docker** and **docker-compose** installed.-->
+Prerequisites for development:
+- a recent python version (>= 3.7) (check with `python -V`)
```
ltshb, line 38: Don't match the version in makefile which is 3.6

buildspec.yml
```diff
@@ -41,21 +41,27 @@ phases:
       - export IMAGE_TAG="${GITHUB_BRANCH}.${COMMIT_HASH}"
       - echo "GITHUB_BRANCH=${GITHUB_BRANCH} COMMIT_HASH=${COMMIT_HASH} IMAGE_TAG=${IMAGE_TAG}"
       - echo "creating a clean environment"
-      - make clean dev
+      - make setup
+      - echo "linting code"
+      - make lint
```
ltshb, line 46: Why not also use the django `./manage.py check` as well in the CI ?

ltshb (changes requested):
I don't have any experience with Django, therefore if using directly ./manage.py is a better choice based on exprerience I'm fine with it.
I also agree on removing the setup of python, but we should do it also on all new services (service-qrcode, service-color, etc).
Concerning the makefile timestamps for the `make setup` I would keep it as it make the makefile more user friendly and takes care of correct dependencies as makefile is designed for.
Another questions is why not using django check in CI ?

.pylintrc
boecklic: `settings.py` is not checked in anymore, it's a local user file and therefore we don't care about linting

ltshb: :+1:

README.md
boecklic: fixed

buildspec.yml
boecklic: just forgot to add it, btw (as you said) it's part of the default django testrunner: https://docs.djangoproject.com/en/3.1/topics/testing/advanced/#using-different-testing-frameworks

Makefile
boecklic: I changed it such that the `pip install -r requirements_dev.txt` runs every time you execute `make setup` (so e.g after a `git pull`)

buildspec.yml
boecklic: The Django default test runner doesn't produce junit reports, I see three options:
- [ ] use xmlrunner instead https://github.com/xmlrunner/unittest-xml-reporting#django-support or
- [ ] use pytest https://docs.pytest.org/en/stable/usage.html#creating-junitxml-format-files
- [ ] live without junit reports

I could live without junit reports, the django testrunner gives a nice summary about failing tests in the end. I personally would avoid pytest and try to stick to a django-conform test-runner. So for me option 1 or 3

ltshb: I find the Codebuild Test report (provided by the junit report) not very helpful in case of error, therefore I suggest option 3.

hansmannj:
Thanks for the proposed changes in this PR. Keeping local dev simple and consistent with other services makes sense to me.
I just updated `models.py` to resolve the conflicts.
If everyone can live with the changes in this PR, I'd suggest to merge it soon. This will hopefully also solve issues in https://github.com/geoadmin/service-stac/pull/20

hansmannj (approved):
:+1:
As you suggested, we'll merge now and the Makefile cleanup and docker stuff will follow soon :smile:

ltshb (approved):

PR #23, opened 2020-10-12
added 1 missing max_length / added list() to ArrayField's defaults

hansmannj:
* added one missing `max_length` to a CharField (missed one field so far, sorry)
* added a list() to some ArrayField's default values, in order to hopefully fix this error:
> ArrayField default should be a callable instead of an instance so that it's not shared between all field instances.
> --
> 1004 | HINT: Use a callable instead, e.g., use `list` instead of `[]`.


ltshb (approved):
:+1:

PR #24, opened 2020-10-12
solved issues with default values for ArrayFields

hansmannj:
added function that returns values of a dictionary with default values for stac extensions for ArrayFields.
This way the following issue is hopefully solved:

> ArrayField default should be a callable instead of an instance so that it's not shared between all field instances.
> --
> 1004 | HINT: Use a callable instead, e.g., use `list` instead of `[]`.

### DIFF
--- a/project/stac_api/models.py
+++ b/project/stac_api/models.py
@@ -1,9 +1,11 @@
 import re
-from django.contrib.gis.db import models
-from django.utils.translation import gettext_lazy as _
-from django.core.exceptions import ValidationError
-from django.contrib.postgres.fields import ArrayField
+
 import numpy as np
+
+from django.contrib.gis.db import models
+from django.contrib.postgres.fields import ArrayField
+from django.core.exceptions import ValidationError
+from django.utils.translation import gettext_lazy as _
 
 # pylint: disable=fixme
 # TODO remove this pylint disable once this is done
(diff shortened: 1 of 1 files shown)


project/stac_api/models.py
```diff
@@ -15,6 +17,30 @@
...
+# non-editable for the moment. Could be changed, should the need arise.
+# The following - a bit complicated approach - hopefully serves to solve the
+# error:
+# <begin Quote>
+# "*ArrayField default should be a callable instead of an
+# instance so that it's not shared between all field instances.
+# HINT: Use a callable instead, e.g., use `list` instead of `[]`.""
+# <end quote>
+def get_default_stac_extensions():
+    DEFAULT_STAC_EXTENSIONS = { # pylint: disable=invalid-name
```
ltshb, line 31: I would stick to the convention here and don't disable the pylint invalid-name. By convention we should have `default_stac_extensions` variable in lower case (upper case is for global constant)

hansmannj: OK, I had in mind, that we name constants in capital letters.. I will change that

hansmannj: After checking our guideline and also other services, I guess, capitals is correct, as it is a dictionary that represents constant values, e.g. like `ALLOWED_DOMAINS` in `service-qrcode` in the `url.py`.
What do you think?

ltshb: Python code guideline consider upper case for global constant, here the variable is a function scope constant and should be lower, however you could move the variable outside the function scope to be global and keep the upper case notation without having pylint complaining.

hansmannj: OK, done. (thought you meant "global" in terms of the whole app)

boecklic (approved):

project/stac_api/models.py
```diff
@@ -27,17 +26,19 @@
 # instance so that it's not shared between all field instances.
 # HINT: Use a callable instead, e.g., use `list` instead of `[]`.""
 # <end quote>
+DEFAULT_STAC_EXTENSIONS = { # pylint: disable=invalid-name
```
ltshb, line 29: The pylint pragma is not anymore useful here :-)

hansmannj: removed :-)

PR #25, opened 2020-10-13
BGDIINF_SB-1291: Added django request info to log output

ltshb:
Add the Django HttpRequest info to the log ouptut of the landing page.

This requires the new upcoming logging-utilities 1.1.0, currently it uses the alpha version which will be updated as soon the PR https://github.com/geoadmin/lib-py-logging-utilities/pull/3 is accepted and merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@
 
 ## Summary of the project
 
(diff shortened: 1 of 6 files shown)


hansmannj (approved):
:+1:

PR #27, opened 2020-10-13
makefile cleanup

boecklic:
I still have issues with gunicorn serve, but I'm not sure it's related to my changes...


ltshb (approved):
:+1:  just a minor changes, the commen in makefile line n°47 could be removed.

boecklic:
> +1 just a minor changes, the commen in makefile line n°47 could be removed.

fixed

ltshb:
I've tested it and unfortunately the target dependencies are not working correctly. So I suggest in order to not loose more time on this and to keep it simple to simply remove the target dependencies (aka removing the timestamps suff in target dependencies)

boecklic:
> I've tested it and unfortunately the target dependencies are not working correctly. So I suggest in order to not loose more time on this and to keep it simple to simply remove the target dependencies (aka removing the timestamps suff in target dependencies)

should be fixed now hopefully (was caused by a 'catch-all' target in Makefile `%:`)

rebert (changes requested):
When starting from scratch i get the fatal error:
Makefile:36: *** "No matching python version found on system, minimum 3.7 required".  Stop.

This is not related to this pr, but the title indicates, that this should work. 

ifeq ($(SYSTEM_PYTHON),)                                                        
$(error "No matching python version found on system, minimum $(PYTHON_VERSION) required")
endif  

In the point of view of the user, it would be great to get at least the .help or that make will install the python local v. 3.7

In min. make setup should work

rebert (approved):

PR #31, opened 2020-10-15
idea about using the geometry to avoid redunancy

rebert:


hansmannj (approved):
Looks good to me, thank you :smile:

boecklic (commented):
Nice, makes the code simpler which is always good! 

Two comments
- You may want to just have one migration file in the PR. But this is a nice to have (to do this, you would migrate back the DB to the state before starting work (./manage.py migrate stac_api 0002_auto_20201016_1423), delete all new migration files and create one new one with all changes)
- the test_serializer.py script is also part of Jürgens branch (and will likely be transformed into a management command), not sure if it should be included here as well

PR #33, opened 2020-10-19
Fixed codebuild patch

ltshb:
The badge changed on codebuild probably because the project has been re-created.


hansmannj (approved):
thanks!

PR #36, opened 2020-10-20
Allow migration module name in pylint

ltshb:
pylint complained about the migrations module name that they were not
snake_case because they started with a number.

Now we allow module name with number.


hansmannj (approved):
:+1: 
Merci

PR #37, opened 2020-10-20
WIP: first draft of collection endpoints and unit tests

hansmannj:
(Sorry, typo in branchname)

This is WIP and I'd like to open up the discussion.
I'd like to have a less detailed list view with specific fields only, e.g. collection_name, description, title and a link to the detail view of every collection.
For me, the default list view does not make sense, as it shows much too much details (same level of detail as in the default detail view).

I've been experimenting locally with dynamic serializers, or having two separate serializers, one for the list view and one for the detail view. Not very successful so far, but now I am too tired to continue with this today ;-)

@boecklic: Do you have suggestions from your Django experience so far?


hansmannj:
P.S.:
I had a version that worked as I wanted to. But I am not sure, if that was Djangonian-like, or if it was too much customization/forcing in the right direction..

If you want to try yourself, replacing the `CollectionList` class with the following snippet will generate a shorter list view (with working links to the detail views) of all collections. You might want to `.manage.py flush` and populate the local db with my testscript (will commit that in a few minutes) first. Also youi'll need to do `from rest_framework.response import Response
`

''' python
class CollectionList(generics.ListAPIView):
    '''
    collection endpoint
    '''
    queryset = Collection.objects.all()
    serializer_class = CollectionSerializer

    def get(self, request, format=None):
        """
        Returns list of all collections by name together with their descriptions and title.
        Also a link to the detail view of the collection is provided for every collection in
        the returned list.
        Can be extended by other properties if needed.
        """
        url_base = request.build_absolute_uri().replace("?format=api", "")
        collection_name = [collection.collection_name for collection in Collection.objects.all()]
        collection_link = [
            f"{url_base}{collection.collection_name}" for collection in Collection.objects.all()
        ]
        collection_description = [collection.description for collection in Collection.objects.all()]
        collection_title = [collection.title for collection in Collection.objects.all()]
        content = zip(collection_name, collection_description, collection_title, collection_link)
        return Response(content)
'''

app/stac_api/views.py
```diff
@@ -54,3 +60,30 @@ def checker(request):
...
+    '''
+    Returns a detail view of the collection instance with all its properties
+    '''
+    serializer_class = CollectionSerializer
+    lookup_url_kwarg = "collection_name"
+    queryset = Collection.objects.all()
+
+    def get_object(self):
+        collection_name = self.kwargs.get(self.lookup_url_kwarg)
+        queryset = self.get_queryset().filter(collection_name=collection_name)
```
ltshb, line 87: Just a small coding style comment here. I would put these class into a collection.py module and put two functions in views.py as follow:

```python
def collection_list():
    return CollectionList.as_view()
```

This would keep views.py concise as we add endpoints, otherwise this file will become too big. Thg views.py import gets also simplified as the `from . import views` is sufficient.

ltshb (commented):
Seems good to me except two minor comment on code styling

hansmannj (commented):
With this commit, dynamic serialization is possible now.
Now the next issue to resolve would be, to add dynamic context (i.e. in the list view for every collection a link to its detail view) to the list view.
Will continue to work on that tomorrow..
Any experiences with that @boecklic?

app/stac_api/urls.py
```diff
@@ -1,10 +1,16 @@
...
 
 from . import views
 
 urlpatterns = [
     path(API_BASE_PATH, views.landing_page, name='landing-page'),
     path('checker/', views.checker, name='checker'),
+    path('collections/', CollectionList.as_view(), name='CollectionList'),
+    path(
+        'collections/<slug:collection_name>/', CollectionDetail.as_view(), name='CollectionDetail'
+    ),
```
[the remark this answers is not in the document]
hansmannj: I think the Django examples also use `kebab-case`, so I'll change it

tests/unit/test_collections_endpoint.py
```diff
@@ -0,0 +1,183 @@
+from datetime import datetime
```
ltshb, line 1: This test must move into /app/tests ! otherwise it won't be started by `make test` or `./manage.py test`

hansmannj: strange, it worked before and I am sure, in older versions the path was correct. Perhaps directory structure changes due to new testrunner?
How shell I proceed? Move the file?

hansmannj: I moved the file to the new directory

app/stac_api/views.py
hansmannj: Good idea, did that

hansmannj:
Thanks for your feedback.
Now:
* no dynamic serializers (not needed at the moment)
* collection views are "externalized" into separate file
* path to collections endpoint is corrected
* unit tests moved to new directory

hansmannj:
Could be merged now (finally) ;-)

ltshb (approved):
:+1:

PR #39, opened 2020-10-21
Fixed gunicorn dependency issue and moved to pipenv

ltshb:
gunicorn did not work due to greenlet release 0.17.0 which was not compatible with gevent 20.6.2.
Moved to pipenv in order to fixes such dependencies.


pakb:
I didn't find in `pipenv` doc if, when a `Pipfile.lock` is present, it will be used ignoring the `Pipfile`. Have you found anything about that?

ltshb:
> I didn't find in `pipenv` doc if, when a `Pipfile.lock` is present, it will be used ignoring the `Pipfile`. Have you found anything about that?

See https://pipenv.pypa.io/en/latest/advanced/#using-pipenv-for-deployments. From my understanding your can do this by `pypenv sync` or have a similar behavior with `pipenv install --deploy` which failed if the lock file is outdated. You could also uses `pipenv install --ignore-pipfile` which explicitely use the lock file instead pipfile. I currently use  `pipenv install --deploy --system`  in dockerfile, the system flag avoid installing dependencies in virtualenv.

pakb:
Ok, then I think we should make our CI uses `pipenv --ignore-pipfile` somehow. That's the best way of checking that our dependencies are in sync with what was on the developer's machine, don't you think?

ltshb:
> Ok, then I think we should make our CI uses `pipenv --ignore-pipfile` somehow. That's the best way of checking that our dependencies are in sync with what was on the developer's machine, don't you think?
> 
> Edit: With `npm` that's what I do using `npm ci` which doesn't read the `package.json` file but install all the things straight from the `package-lock.json` file

I did some test and based on https://pipenv.pypa.io/en/latest/advanced/#using-pipenv-for-deployments  I'll do as follow:
- use `pipenv sync` on the CI which makes sure to use the lock file
- use `pipenv install --ignore-pipfile --deploy --system` to build the container

`pipenv sync` is similar to `pipenv install --ignore-pipfile --deploy` however the `sync` command don't supports the `--system` flag which avoid building the virtualenv on the container image.

Thanks @pakb for your precious inputs ;-)

pakb (approved):
no problem, this way we are sure that we have the outcome we wanted

Makefile
```diff
@@ -17,7 +16,7 @@ TEST_DIR := $(CURRENT_DIR)/$(APP_SRC_DIR)/tests
 
 # general targets timestamps
 TIMESTAMPS = .timestamps
-SETUP_TIMESTAMP = $(TIMESTAMPS)/.setup.timestamp
+SETTINGS_TIMESTAMP = $(TIMESTAMPS)/.settins.timestamp
```
boecklic, line 19: typo `settings`

boecklic (approved):
since pipenv is a pip package itself: could we make the only entry in a `requirements.txt` and include it in the setup (make env et al.) process?

ltshb:
> e pipenv is a pip package itself: could we make the only entry in a

I though about this, but pipenv needs to be installed globally on the system and I don't think it make sense to always try to install it with pip. On my machine I also had to update globally keyrings.alt to avoid errors which might not be the case on all machine (keyrings is architecture dependend, linux, mac, windows). So I would prefer letting pipenv as an outside requirement like python and the postgres.

boecklic:
> > e pipenv is a pip package itself: could we make the only entry in a
> 
> I though about this, but pipenv needs to be installed globally on the system and I don't think it make sense to always try to install it with pip. On my machine I also had to update globally keyrings.alt to avoid errors which might not be the case on all machine (keyrings is architecture dependend, linux, mac, windows). So I would prefer letting pipenv as an outside requirement like python and the postgres.

:+1:

PR #41, opened 2020-10-22
adding django_debug toolbar

boecklic:
debug_toolbar is a handy add-on that's useful for debugging. I propose to install it since login fails on `bgdi-dev` due to a missing csrf token so it would be useful to debug http headers and stuff

### DIFF
--- a/app/middleware/debug_toolbar.py
+++ b/app/middleware/debug_toolbar.py
@@ -7,9 +7,8 @@ def check_toolbar_env(request):
     for details see
     https://django-debug-toolbar.readthedocs.io/en/latest/configuration.html#debug-toolbar-config  # pylint: disable=line-too-long
     """
-    print(os.environ.get('APP_ENV'))
+
     if os.environ.get('APP_ENV', 'prod') in ['local', 'dev']:
-        print('blublu')
         return True
 
     return False
--- a/app/middleware/debug_toolbar.py
+++ b/app/middleware/debug_toolbar.py
@@ -1,0 +1,15 @@
+import os
+
+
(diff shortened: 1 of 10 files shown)


app/middleware/debug_toolbar.py
```diff
@@ -0,0 +1,15 @@
...
+
+def check_toolbar_env(request):
+    """ callback to check whether debug toolbar should be shown or not
+
+    for details see
+    https://django-debug-toolbar.readthedocs.io/en/latest/configuration.html#debug-toolbar-config  # pylint: disable=line-too-long
+    """
+    print(os.environ.get('APP_ENV'))
+    if os.environ.get('APP_ENV', 'prod') in ['local', 'dev']:
+        print('blublu')
```
ltshb, line 12: We should use the logging instead of print statement here with a meaningful message :-)

ltshb (commented):
See inline comment

app/middleware/debug_toolbar.py
boecklic: :+1: leftover from development, removed it

ltshb (approved):
:+1:

boecklic:
General question: currently we have none of the dev-tools in the Docker image that's deployed on dev. Does it make sense to deploy a `dev` image on dev env? Cause with the current image the debug_toolbar won't be usable on bgdi-dev

rebert (approved):

boecklic:
Note: this will go together with this PR https://github.com/geoadmin/infra-kubernetes/pull/23

ltshb (approved):
:+1:

PR #44, opened 2020-10-23
BGDIINF_SB-1374: Added pagination for collections endpoint in list view

hansmannj:
[BGDIINF_SB-1374] Added pagination for the collections endpoint in list view


app/stac_api/collection_views.py
```diff
@@ -1,17 +1,27 @@
...
 
 
+# using this extra pagination class for collection list view allows for having
+# global pagination settings in `settings_prod.py` that apply generally and to
+# override those settings for specific views if needed.
+# TODO The very small value of 2 is chosen here for testing, as we only have a very #pylint: disable=fixme
+# small number of collections in the test db so far. In the end, this value has
+# to be set to something more meaningful.
+class CollectionPagination(PageNumberPagination):
+    page_size = 2
```
ltshb, line 15: I would remove this and set the pagination in the `settings_dev.py` as follow:
```python
if DEBUG:
    REST_FRAMEWORK['PAGE_SIZE'] = os.getenv(PAGE_SIZE, 2)
```
This avoid the TODO im code and being able to reduce the page size even once we are on prod can always be useful for development where we won't develop using the actual data. It also gives the possibility to change the page size by environment variable..

ltshb (changes requested):
:+1:  except for the PAGE_SIZE TODO that I would remove and make it configurable in the settings_dev.py.

app/stac_api/collection_views.py
boecklic: :+1:

boecklic (approved):

app/stac_api/collection_views.py
hansmannj: done

boecklic (approved):

ltshb (commented):
Does the Rest Framework adds the pagination as in our spec ? Meaning in the `links: {'next': ...}` ?

ltshb:
Should we not have a jira ticket for this work as well to track it in the sprint review ? and add the ticket in the pr title

boecklic:
> Should we not have a jira ticket for this work as well to track it in the sprint review ? and add the ticket in the pr title

Yeah, in an ideal world we should do / have this definitely, our world is unfortunately not ideal yet ;-).

hansmannj:
> Does the Rest Framework adds the pagination as in our spec ? Meaning in the `links: {'next': ...}` ?

Good point, not yet. After discussion with Brice: I will continue working on the other open PR on collection serializer/endpoint and Brice will continue working on pagination

ltshb (approved):
The correct pagination output will be done in another PR.

PR #46, opened 2020-10-27
Used JSONFields for extent and summaries 

hansmannj:
JSON fields provide an easy way to deal with the extend and summaries properties of collections.
They can be (de-)serialise into the desired formats straight-forward.
The validation part is yet to be implemented.

I allowed myself to also adapt things in the models and serializers for items and assets already.
This is just a suggestion.


ltshb (commented):
For some uses cases I think `JSONField` has advantages. The drawback is that we have to do validation of values manually. Depending on the values types we might need to  use the `DjangoJSONEncoder` to handle `datetime` objects. Another drawback is if a json field contains a relational value like the `Item.properties.providers` field, I don't if this can easily handled. Therefore I think for `Item.properties` `JSONField` are not well suited, but it can be a good choices for other field though.

hansmannj:
This PR is included in PR https://github.com/geoadmin/service-stac/pull/52.
Once https://github.com/geoadmin/service-stac/pull/52 is merged, I will delete this one

hansmannj:
Already integrated in https://github.com/geoadmin/service-stac/pull/52
So I'll close this PR now, as it has been overtaken ;-)

hansmannj:
Already integrated in #52
So I'll close this PR now, as it has been overtaken ;-)

PR #47, opened 2020-10-28
Improved logging configuration for development

ltshb:
Now you can completely disable logging via an simple boolean environment
variable `DISABLE_LOGGING`. This is usefull espacialy when working with
the django shell.

Also fixed the `TEST_ENABLE_LOGGING` which true when defined but didnot
checked the value correctly.

Updated the documentation accordingly and also removed old mention to
`pip` in docu while we are now using `pipenv`

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@
   - [Setup app](#setup-app)
   - [Starting dev server](#starting-dev-server)
   - [Running test](#running-test)
+  - [Using Django shell](#using-django-shell)
   - [Linting and formatting your work](#linting-and-formatting-your-work)
 - [Deploying the project and continuous integration](#deploying-the-project-and-continuous-integration)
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -131,6 +131,24 @@ you can choose to create a new test-db on every run or to keep the db, which spe
...
+Django shell can be use for development purpose (see [Django: Playing with the API](https://docs.djangoproject.com/en/3.1/intro/tutorial02/#playing-with-the-api))
+
+```bash
+./manage.py shell
+```
+
+You can disable totally logging while playing with the shell as follow:
+
+```bash
+DISABLE_LOGGING=1 ./manage.py shell
```
boecklic, line 147: :+1:

boecklic (approved):

PR #48, opened 2020-10-28
modifications to state from doc-api-spec

boecklic:
I started to make some cleanup. I suspect further little things will pop up, specially related to examples.

Note: if commenting pls comment on the source files, NOT `openapi.yaml` or `openapitransactional.yaml`, those are automatically compiled from the different sources. 

What I changed in particular: 
- removed the big example on item, and have the example for item be automatically generated from examples of the individual parts. 
- use stac's `item` and `itemCollection` instead of `featureGeoJson` which stems from the ogc feature api (this is what stac does as well, it was not correctly compiled in the previous spec so far. the spec for body of `item` is more complete now.

Unfortunately our docker0 don't work at the moment (issue pending @ c2c), otherwise this version would be visible at https://ltboc.infra.bgdi.ch/static/products/data.geo.admin.ch/api.html


hansmannj:
:+1:
Thanks, I had a quick look at it and it seems good to me. 
Anyways in case further little things will pop up while we are working on it and diving further and further into details, we'll have to adapt this in future.

hansmannj (approved):

PR #49, opened 2020-10-28
adding some debug packages to the image

boecklic:
The minimal image we're currently building doesn't offer tools for debugging networking issues in the container running on k8s. 

A slightly bigger image is built with this PR including some of the popular tools for debugging networking issues. Can be removed at a later stage once issue is found.


ltshb (approved):
Might create a Jira Ticket to not forgot to remove these tools. Although they might be usefull on PROD if once we got some trouble.

PR #50, opened 2020-10-28
BGDIINF_SB-1266: Added Item serializer (only serialization)

ltshb:
For the moment only the serialization is implemented, the de-serialization can be done later on when needed.


hansmannj (approved):
Looks good to me, thanks :+1:

app/stac_api/serializers.py
```diff
@@ -183,3 +192,106 @@ def update(self, instance, validated_data):
...
+        python_native = super().to_representation(instance)
+        return python_native['bbox']
+
+
+class AssetsDictSerializer(DictSerializer):
+    # pylint: disable=abstract-method
+    key_identifier = 'asset_name'
+
+
+class AssetsItemSerializer(serializers.ModelSerializer):
```
hansmannj, line 237: Not sure, if we need to explicitly define the `id` field, since we override the auto-generated primary key. But probably the declarations in `models.py` are enough and no further explicitness needed here.

rebert (approved):

PR #52, opened 2020-10-29
BGDIINF_SB-1264: combined new item and collection serializers

hansmannj:
Serializers for collection and items should now work.
Collection serializer (with JSONFields for summaries and extent) in both ways, items only serialization so far.
Asset model is adapted to be compatible with new collections model.


app/stac_api/models.py
```diff
@@ -120,46 +118,44 @@ class Collection(models.Model):
...
     )
     stac_version = models.CharField(max_length=10)  # string
 
     # "summaries" values will be updated on every update of an asset inside the
     # collection
-    summaries_eo_gsd = ArrayField(models.FloatField(), blank=True, null=True)
-    summaries_proj = ArrayField(models.IntegerField(), blank=True, null=True)
+    summaries = models.JSONField(blank=True, null=True)
+    #summaries_eo_gsd = ArrayField(models.FloatField(), blank=True, null=True)
+    #summaries_proj = ArrayField(models.IntegerField(), blank=True, null=True)
```
ltshb, line 138: These should be removed

app/stac_api/models.py
```diff
@@ -120,46 +118,44 @@ class Collection(models.Model):
...
     # after discussion with Chris and Tobias: geoadmin_variant will be an
     # array field of CharFields. Simple validation is done (e.g. no "Sonderzeichen"
     # in array)
     # geoadmin_variant will be also auto-populated on every
     # update of an asset
-    geoadmin_variant = ArrayField(models.CharField(max_length=15), blank=True, null=True)
-
+    #geoadmin_variant = ArrayField(models.CharField(max_length=15), blank=True, null=True)
```
ltshb, line 144: Clean up

app/stac_api/models.py
```diff
@@ -120,46 +118,44 @@ class Collection(models.Model):
...
     title = models.CharField(blank=True, max_length=255)  # string
 
     def __str__(self):
         return self.collection_name
 
     def clean(self):
         # TODO: move this check to the items save()
-        if self.start_date is None and self.end_date is None:
+        if self.extent["temporal"]["interval"][0][0] is None and self.extent["temporal"][
```
ltshb, line 152: Exception handling is missing; KeyError and IndexError must be handle here

app/stac_api/models.py
```diff
@@ -120,46 +118,44 @@ class Collection(models.Model):
...
+            "interval"][0][1] is None:
             raise ValidationError(_('At least a start date or an end date has to be defined.'))
 
         # very simple validation, raises error when geoadmin_variant strings contain special
         # characters or umlaut.
-        for variant in self.geoadmin_variant:
+        for variant in self.summaries["geoadmin:variant"]:
```
ltshb, line 158: KeyError must be handle here otherwise it will crash if the `geoadmin:variant` is missing

app/stac_api/models.py
```diff
@@ -210,17 +206,20 @@ def save(self, force_insert=False, force_update=False, using=None, update_fields
         # I leave this open for the moment.
 
         # check if collection's start_ and end_dates need to be updated
-        if self.collection.start_date is None:
-            self.collection.start_date = self.properties_datetime
+        if self.collection.extent["temporal"]["interval"][0][0] is None:
```
ltshb, line 209: Exception of `KeyError` and `IndexError`

app/stac_api/models.py
```diff
@@ -210,17 +206,20 @@ def save(self, force_insert=False, force_update=False, using=None, update_fields
...
+
+            self.collection.extent["temporal"]["interval"][0][0] = self.properties_datetime
```
ltshb, line 211: Exception of `KeyError` and `IndexError`

app/stac_api/models.py
```diff
@@ -210,17 +206,20 @@ def save(self, force_insert=False, force_update=False, using=None, update_fields
...
             self.collection.save()
 
-        elif self.properties_datetime < self.collection.start_date:
-            self.collection.start_date = self.properties_datetime
+        elif self.properties_datetime < self.collection.extent["temporal"]["interval"][0][0]:
```
ltshb, line 214: Exception of `KeyError` and `IndexError`

app/stac_api/models.py
```diff
@@ -210,17 +206,20 @@ def save(self, force_insert=False, force_update=False, using=None, update_fields
...
+
+            self.collection.extent["temporal"]["interval"][0][0] = self.properties_datetime
```
ltshb, line 216: Exception of `KeyError` and `IndexError`

app/stac_api/models.py
```diff
@@ -210,17 +206,20 @@ def save(self, force_insert=False, force_update=False, using=None, update_fields
...
             self.collection.save()
 
-        if self.collection.end_date is None \
-            or self.properties_datetime > self.collection.end_date:
-            self.collection.end_date = self.properties_datetime
+        elif self.properties_datetime > self.collection.extent["temporal"]["interval"][0][1] or \
```
ltshb, line 219: Exception of `KeyError` and `IndexError`

app/stac_api/models.py
```diff
@@ -281,24 +280,24 @@ def save(self, force_insert=False, force_update=False, using=None, update_fields
         self.collection = self.feature.collection
 
         # check if the collection's geoadmin_variant needs to be updated
-        if self.geoadmin_variant not in self.feature.collection.geoadmin_variant:
-            self.feature.collection.geoadmin_variant.append(self.geoadmin_variant)
+        if self.geoadmin_variant not in self.feature.collection.summaries["geoadmin:vairant"]:
```
ltshb, line 283: Exception of `KeyError`

app/stac_api/models.py
```diff
@@ -281,24 +280,24 @@ def save(self, force_insert=False, force_update=False, using=None, update_fields
...
+            self.feature.collection.summaries["geoadmin:vairant"].append(self.geoadmin_variant)
             self.feature.collection.save()
 
         # proj_epsq (integer) is defined on collection level as well
         # and eo_gsd (float) on item AND collection level as well.
         # So we need to check if these properties need an update on parent
         # and grandparent level.
-        if self.proj_epsq not in self.feature.collection.summaries_proj:
-            self.feature.collection.summaries_proj.append(self.proj_epsq)
+        if self.proj_epsq not in self.feature.collection.summaries["proj:epsg"]:
```
ltshb, line 291: Exception of `KeyError`

app/stac_api/models.py
```diff
@@ -281,24 +280,24 @@ def save(self, force_insert=False, force_update=False, using=None, update_fields
...
+            self.feature.collection.summaries["proj:epsg"].append(self.proj_epsq)
             self.feature.collection.save()
 
         # for float-comparison:
         def float_in(flt, floats, **kwargs):
             return np.any(np.isclose(flt, floats, **kwargs))
 
-        if not float_in(self.eo_gsd, self.feature.collection.summaries_eo_gsd):
-            self.feature.collection.summaries_eo_gsd.append(self.eo_gsd)
+        if not float_in(self.eo_gsd, self.feature.collection.summaries["es:gsd"]):
```
ltshb, line 299: Exception of `KeyError`

ltshb (changes requested):
Error handling for the Json field is missing, it needs to handle some cases otherwise the application will crash.

app/stac_api/models.py
hansmannj: as for all of your "KeyError" comments: I will add handling, but as these fields are only for internal write access, the requested keys and indices should always work. But yes, if something is wrong with the internal writing, this would cause errors. I'll add that soon

app/stac_api/models.py

app/stac_api/models.py
hansmannj: added exception handling here and also (hopefully) everywhere, where Key- or IndexErrors might occur

app/stac_api/models.py
```diff
@@ -125,48 +137,52 @@ class Collection(models.Model):
...
         # very simple validation, raises error when geoadmin_variant strings contain special
         # characters or umlaut.
-        for variant in self.geoadmin_variant:
-            if not bool(re.search('^[a-zA-Z0-9]*$', variant)):
-                raise ValidationError(_('Property geoadmin:variant not correctly specified.'))
+
+        try:
+            for variant in self.summaries["geoadmin:variant"]:
+                if not bool(re.search('^[a-zA-Z0-9]*$', variant)):
+                    raise ValidationError(_('Property geoadmin:variant not correctly specified.'))
```
boecklic, line 179: The error message could also give hints on what was wrong (we don't have security implications here, so we can very well tell the user what exactly was wrong or what we expect)

app/stac_api/models.py
```diff
@@ -214,19 +230,35 @@ def save(self, force_insert=False, force_update=False, using=None, update_fields
...
-            self.collection.start_date = self.properties_datetime
-            self.collection.save()
+        try:
+
+            if self.collection.extent["temporal"]["interval"][0][0] is None:
 
-        elif self.properties_datetime < self.collection.start_date:
-            self.collection.start_date = self.properties_datetime
-            self.collection.save()
+                self.collection.extent["temporal"]["interval"][0][0] = self.properties_datetime
```
boecklic, line 237: `extent["temporal"]["interval"][0][0]` <- this smells a bit like it could cause troubles if for some reason the extent field doesn't have this exact structure. This should never happen, but still...

app/stac_api/models.py
```diff
@@ -277,37 +309,66 @@ def __str__(self):
...
-
-        if not float_in(self.eo_gsd, self.feature.properties_eo_gsd):
-            self.feature.properties_eo_gsd.append(self.eo_gsd)
-            self.feature.collection.save()
+        try:
+            # check if the collection's geoadmin_variant needs to be updated
+            if self.feature.collection.summaries["geoadmin:variant"] is None:
+
+                self.feature.collection.summaries["geoadmin:variant"] = [self.geoadmin_variant]
+                self.feature.collection.save()
```
boecklic, line 330: For these checks it would maybe be cleaner to create little helpers/wrappers in the collection model, like `update_variants` or the like. This would keep the internal logic of these fields encapsulated to the collection, item in the end doesn't need to know about how these fields are organized internally. If we change it, it's 'just' a change in the internal representation in the collection and doesn't need to interest others (in the spirit of OOP)

boecklic (changes requested):

app/stac_api/models.py
hansmannj: Every access to the JSON-fields reps. dictionaries should now be inside a try ... except statement, so that at least an error message will be raised and logged, in case something unexpected happens

app/stac_api/models.py
hansmannj: error messages are now more verbose

app/stac_api/models.py
hansmannj: Good point, done!

app/stac_api/models.py
```diff
@@ -125,48 +145,137 @@ class Collection(models.Model):
...
+                self.summaries["proj:epsg"].append(asset_proj_epsg)
+                self.save()
+
+            if self.summaries["eo:gsd"] is None:
+                self.summaries["eo:gsd"] = [asset_eo_gsd]
+                self.save()
+
+            elif not float_in(asset_eo_gsd, self.summaries["eo:gsd"]):
+                self.summaries["eo:gsd"].append(asset_eo_gsd)
+                self.save()
```
ltshb, line 215: Just a minor code style, I would group the `if ... elif` without blank lines and have only blank line between `if ... elif` block this make it easier to read.

app/stac_api/models.py
```diff
@@ -125,48 +145,137 @@ class Collection(models.Model):
...
+
+                self.extent["temporal"]["interval"][0][0] = item_properties_datetime
+                self.save()
+
+            elif item_properties_datetime < self.extent["temporal"]["interval"][0][0]:
+
+                self.extent["temporal"]["interval"][0][0] = item_properties_datetime
+                self.save()
+
+            elif self.extent["temporal"]["interval"][0][1] is None:
```
ltshb, line 245: Should not it be `if` here ? and same comment about blank lines as above.

app/stac_api/models.py
```diff
@@ -125,48 +145,137 @@ class Collection(models.Model):
...
+            if self.summaries["eo:gsd"] is None:
+                self.summaries["eo:gsd"] = [asset_eo_gsd]
+                self.save()
+
+            elif not float_in(asset_eo_gsd, self.summaries["eo:gsd"]):
+                self.summaries["eo:gsd"].append(asset_eo_gsd)
+                self.save()
+
+        except (KeyError, IndexError) as err:
+            logger.error(_(
```
ltshb, line 218: I would not use `_()` getext in logging. Getext is for localization and we don't want to localize logging. All logging must be written in English. Localization should be kept for user interface message (e.g. like in the `ValidationError` where the message is for the user)

app/stac_api/models.py
```diff
@@ -125,48 +145,137 @@ class Collection(models.Model):
...
+                self.save()
+
+            elif item_properties_datetime > self.extent["temporal"]["interval"][0][1]:
+
+                self.extent["temporal"]["interval"][0][1] = item_properties_datetime
+                self.save()
+
+        except (KeyError, IndexError) as err:
+
+            logger.error(_('Updating the collection extent due to item update failed: %s'), err)
```
ltshb, line 257: I would not use `_()` getext in logging. Getext is for localization and we don't want to localize logging. All logging must be written in English. Localization should be kept for user interface message (e.g. like in the `ValidationError` where the message is for the user)

app/stac_api/models.py
```diff
@@ -125,48 +145,137 @@ class Collection(models.Model):
...
         # very simple validation, raises error when geoadmin_variant strings contain special
         # characters or umlaut.
-        for variant in self.geoadmin_variant:
-            if not bool(re.search('^[a-zA-Z0-9]*$', variant)):
-                raise ValidationError(_('Property geoadmin:variant not correctly specified.'))
+
+        try:
+            for variant in self.summaries["geoadmin:variant"]:
+                if not bool(re.search('^[a-zA-Z0-9]*$', variant)):
+                    raise ValidationError(_("Property geoadmin:variant not compatible with the" +
```
ltshb, line 267: A log message would be appropriate here as well.

app/stac_api/models.py
```diff
@@ -125,48 +145,137 @@ class Collection(models.Model):
...
+                                            "naming conventions."))
+
+        except (KeyError, IndexError) as err:
+            # we should only land here, if the default values for the extent are corrupted.
+            # they can only be overriden by internal write processes.
+            logger.error(_("Error when trying to access property geoadmin:variant from inside " +
```
ltshb, line 273: I would not use `_()` getext in logging. Getext is for localization and we don't want to localize logging. All logging must be written in English. Localization should be kept for user interface message (e.g. like in the `ValidationError` where the message is for the user)

app/stac_api/models.py
```diff
@@ -209,24 +318,40 @@ class Item(models.Model):
...
+        needs to be updated. If so, it will be either
+        updated or an error will be raised, if updating fails.
+        '''
+
+        try:
+            # check if eo:gsd on feature/item level needs updates
+            if self.properties_eo_gsd is None:
+                self.properties_eo_gsd = [asset_eo_gsd]
+                self.save()
+
```
ltshb, line 337: Remove blank line

app/stac_api/models.py
```diff
@@ -209,24 +318,40 @@ class Item(models.Model):
...
+            elif not float_in(asset_eo_gsd, self.properties_eo_gsd):
+                self.properties_eo_gsd.append(asset_eo_gsd)
+                self.save()
+
+        except (KeyError, IndexError) as err:
+            logger.error(_(
```
ltshb, line 343: Remove getext

app/stac_api/models.py
```diff
@@ -277,37 +402,29 @@ def __str__(self):
...
+                "Error when trying to access property geoadmin:variant from within asset's " +
+                "clean function: %s"),
+                         err
+                        )
+            raise ValidationError(_(
+                "Error when trying to access property geoamin:variant from within asset's " +
+                "clean function"
+            ))
```
ltshb, line 419: Remove the `try ... except` which is not needed here. And add a log in the case of validation error

ltshb (changes requested):
Good job, only issues with the `if ... elif` block and the removing the `_()` getext from logging. We need to synchronize to see who is merging first with PR #56 and #57.

What is the purpose of the `scripst/test_serializer.py` ? should this one not be removed and only use the unit test ?

app/stac_api/models.py
hansmannj: From my understanding, `elif` is correct, but the whole block is quite nasty to read.

When reaching the line (formerly known as line 245), so far it has been checked, that:
* start_date in the extent's temporal inteval is **not** `None` **and**
* current item instance's datetime value is **not** < extent's temporal interval's start_date, so **only** then:
* check if end_date in extent's temporal interval is `None` and if so,
* set that to the current item's instances datetime value

Hence, it should be an `eflif` and not only an `if` here, from my understanding.

app/stac_api/models.py
hansmannj: done, I'll check, if that might be a bug in yapf or was a bug in my brain ;-)

app/stac_api/models.py
hansmannj: make sense, I've changed that

app/stac_api/models.py
hansmannj: yeah, why not :-)

app/stac_api/models.py
hansmannj: Done

boecklic (dismissed):
nice!
One little thing: `scripts/test_serializer.py` should be transformed to a regular unit test

hansmannj:
> nice!
> One little thing: `scripts/test_serializer.py` should be transformed to a regular unit test

I added extra unit tests. The "test_serializer.py" had a wrong name. It was a script to fill the local test db with some values (until the manage.py fill_db_command works). I renamed it, to avoid confusions.

app/stac_api/models.py
```diff
@@ -209,24 +314,39 @@ class Item(models.Model):
...
+        try:
+            # check if eo:gsd on feature/item level needs updates
+            if self.properties_eo_gsd is None:
+                self.properties_eo_gsd = [asset_eo_gsd]
+                self.save()
+            elif not float_in(asset_eo_gsd, self.properties_eo_gsd):
+                self.properties_eo_gsd.append(asset_eo_gsd)
+                self.save()
+
+        except (KeyError, IndexError) as err:
```
ltshb, line 337: Remove unnecessary `try ... except`

app/stac_api/models.py
```diff
@@ -125,48 +145,131 @@ class Collection(models.Model):
...
+        try:
+            for variant in self.summaries["geoadmin:variant"]:
+                if not bool(re.search('^[a-zA-Z0-9]*$', variant)):
+                    logger.error(
+                        "Property geoadmin:variant not compatible with the naming conventions."
+                    )
+                    raise ValidationError(_("Property geoadmin:variant not compatible with the"
+                                            "naming conventions."))
+
+        except (KeyError, IndexError) as err:
```
ltshb, line 261: Remove unnecessary `IndexError` from `try .... except`

app/stac_api/models.py
```diff
@@ -125,48 +145,131 @@ class Collection(models.Model):
...
+                self.save()
+
+            if self.summaries["eo:gsd"] is None:
+                self.summaries["eo:gsd"] = [asset_eo_gsd]
+                self.save()
+            elif not float_in(asset_eo_gsd, self.summaries["eo:gsd"]):
+                self.summaries["eo:gsd"].append(asset_eo_gsd)
+                self.save()
+
+        except (KeyError, IndexError) as err:
```
ltshb, line 212: Remove unnecessary `IndexError` from `try .... except`

ltshb (approved):
Looks great :+1: Just minor clean up in the `try ... except` error handling, see inline comments

app/stac_api/models.py

app/stac_api/models.py
hansmannj: done

app/stac_api/models.py

app/stac_api/models.py
hansmannj: this one should stay I guess

ltshb: It should be `except KeyError as err:` as there should not have any `IndexError` in this code, only `KeyError`

PR #54, opened 2020-10-29
BGDIINF_SB-1362: Added Items and Item endpoint and corrected the Collections endpoint

ltshb:
Moved the collection view back to views.py as it was internal discussed.
    
In the collections endpoint encapsulate the list into an object in the
'collections' key as defined in the spec.
    
Fixed a bug in the collection endpoint where the collection name is a
string and not a slug. Slug doesn't allow special characters like `.`
which might be part of the collection name.
    
Added Items and Item endpoint.

NOTE: I'll add unit test for the items endpoints in a separate PR.


app/stac_api/urls.py
```diff
@@ -1,18 +1,32 @@
...
 
-from config.settings import API_BASE_PATH
 from stac_api.views import CollectionDetail
 from stac_api.views import CollectionList
+from stac_api.views import ItemDetail
+from stac_api.views import ItemsList
 
 from . import views
 
+API_BASE = settings.API_BASE_PATH
```
boecklic, line 11: `API_BASE` is called `API_BASE_PATH` below, maybe stick to one of the two and use the same in settings as well

boecklic (approved):

app/stac_api/urls.py
ltshb: It was on purpose to make the code lighter, I'll rename it everywhere to `API_BASE` which is shorter ;-)

boecklic: but less descriptive :stuck_out_tongue:

boecklic (approved):
Nice! Think about adding logging where appropriate (e.g. view functions), will help in debugging

ltshb:
> Nice! Think about adding logging where appropriate (e.g. view functions), will help in debugging

Good point I added some logging. We still need to add logging in the models though  but I leave this to the Collections PR #52 which did some modification in models

PR #55, opened 2020-10-29
BGDIINF_SB_1364 api spec

boecklic:
This PR adds the openapi yaml files for the spec to the repository.

It furthermore contains entries in `Makefile` to build the spec and serve locally. The build results of the spec are included in django's static directory, so they will be accessible under `/static` as well. The build of the spec is not part of the build of the service. Changes in the spec require proper review and communication and hence usually have a different workflow than the code. To avoid publishing uncommunicated changes, the builds are separated.

@ltshb pls keep an eye on the change I made in `settings_prod.py` regarding the `STATIC_URL` (note that the root of the service will be at `<host>/api/stac/v0.9/`), it would be good to have the correct url pattern on develop already.


ltshb (approved):
:+1:

ltshb:
I've just seen that the default port for serving the spec is the same as for serving django, for locally testing is not really useful as we cannot serve both together at the same time, so having different default ports would be good :-)

hansmannj (approved):
:+1: 
Looks good to me

rebert (approved):

PR #56, opened 2020-10-29
BGDIINF_SB-1362: Improved test code and added items endpoint test

ltshb:
Instead of redoing the db initialization in each test, moved the code in a package, database.py and reused it in tests.
    
Added items endpoints test and also the collections list endpoint test.

PR #57, opened 2020-10-29
BGDIINF_SB-1262: Added Asset Serializer and Endpoint

ltshb:
Added Asset Serializer and Endpoint

NOTE: Once PR #56 is merged I'll rebase this one on top of develop


app/stac_api/urls.py
```diff
@@ -13,20 +15,26 @@
...
     ),
+    path(f"{API_BASE}collections/<collection_name>/items", ItemsList.as_view(), name='items-list'),
     path(
-        f"{API_BASE}collections/<str:collection_name>/items",
-        ItemsList.as_view(),
-        name='items-list'
-    ),
-    path(
-        f"{API_BASE}collections/<str:collection_name>/items/<str:item_name>/",
+        f"{API_BASE}collections/<collection_name>/items/<item_name>",
```
hansmannj, line 26: Is there some kind of validation of the URL parameters? Before there was something like `<str: >`.

app/tests/database.py
```diff
@@ -0,0 +1,140 @@
...
+    '''Create a dummy collection db object for testing
+    '''
+    collection = Collection.objects.create(
+        crs=["http://www.opengis.net/def/crs/OGC/1.3/CRS84"],
+        created='2020-10-28T13:05:10.473602Z',
+        updated='2020-10-28T13:05:10.473602Z',
+        description='This is a description',
+        start_date=None,
+        end_date=None,
+        extent=[200000, 100000, 200001, 100005],
```
hansmannj, line 21: Not compatible with the latest collection model any more.

app/tests/database.py
```diff
@@ -0,0 +1,140 @@
...
+        collection_name='ch.swisstopo.pixelkarte-farbe-pk200.noscale',
+        item_type='Feature',
+        license='test',
+        stac_extension=get_default_stac_extensions(),
+        stac_version="0.9.0",
+        summaries_eo_gsd=[10.1, 20.3, 30.44],
+        summaries_proj=[1, 4, 22],
+        geoadmin_variant=['blubb', 'blabb', 'blibb'],
```
hansmannj, line 29: Not compatible with the latest collection model. Summaries are now stored in JSONField.

hansmannj (changes requested):
:+1: 
Just some minor changes, i.e. adaptions to the latest `models.py` preferably after the PRs regarding collection endpoints are merged

app/stac_api/urls.py
ltshb: `<str:>` is the default, so having it defined or not is the same, wihtout it, it makes it a little bit shorter and we don't need to split it over several lines. (See commit message https://github.com/geoadmin/service-stac/pull/57/commits/bb3a2bce74de99eb0e25e5db03b6d02375eb36cd ;-) )

app/tests/database.py
ltshb: Well it depends on who is merging first :-). Each PR must be compatible with the develop branch and not on other PR. So Either I merged it now (this changes needs to be part of your PR then) or I wait that you merge the Collection serializer and do the changes within this PR :-)

hansmannj (approved):

PR #58, opened 2020-11-03
BGDIINF_SB-1374: implemented cursor pagination

ltshb:
Implemented a cursor pagination that matchs the STAC API spec.

Also fixed a bug in the collection model when creating a collection from admin page.

### DIFF
--- a/app/stac_api/apps.py
+++ b/app/stac_api/apps.py
@@ -1,7 +1,14 @@
+import logging
+from gettext import gettext as _
+
 from django.apps import AppConfig
+from django.conf import settings
 
 from rest_framework import pagination
 from rest_framework.response import Response
+from rest_framework.exceptions import ValidationError
+
+logger = logging.getLogger(__name__)
 
 
 class StacApiConfig(AppConfig):
@@ -10,6 +17,8 @@ class StacApiConfig(AppConfig):
 
 class CursorPagination(pagination.CursorPagination):
     ordering = 'id'
+    page_size_query_param = 'limit'
+    max_page_size = settings.REST_FRAMEWORK['PAGE_SIZE_LIMIT']
 
     def get_paginated_response(self, data):
         links = {}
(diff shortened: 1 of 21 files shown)


app/stac_api/models.py
```diff
@@ -182,30 +182,21 @@ def update_geoadmin_variants(self, asset_geoadmin_variant, asset_proj_epsg, asse
...
                 self.summaries["proj:epsg"].append(asset_proj_epsg)
                 self.save()
 
-            if self.summaries["eo:gsd"] is None:
-                self.summaries["eo:gsd"] = [asset_eo_gsd]
-                self.save()
-            elif not float_in(asset_eo_gsd, self.summaries["eo:gsd"]):
+            if not float_in(asset_eo_gsd, self.summaries["eo:gsd"]):
                 self.summaries["eo:gsd"].append(asset_eo_gsd)
                 self.save()
```
hansmannj, line 201: Nice idea setting the defaults for the summaries to `[]` instead of `None`. Makes this if-block much more elegant :smile:

rebert (approved):

app/stac_api/apps.py
```diff
@@ -1,5 +1,69 @@
...
+        elif links:
+            data['links'].append(links)
+        return Response(data)
+
+    def get_page_size(self, request):
+        # Overwrite the default implementation about the page size as this one
+        # don't validate the query parameter, its simply correct it if it is not valid
+        # here we want to return a 400 BAD REQUEST when the provided page size is invalid.
+        try:
+            integer_string = request.query_params[self.page_size_query_param]
```
boecklic, line 45: can you use `request.query_params.get(self.page_size_query_param, self.page_size)`? (I'm not sure if this is supported for the `query_params` though), this would be a bit less verbose

boecklic (approved):

app/stac_api/apps.py
ltshb: Good point, I've made the changes and it is supported :-)

PR #59, opened 2020-11-03
Fixes to be able to load colletions and items

rebert:
As a quick and dirty way to be able to load collections and items

Are we going to safe (deal with) time with offset (UTC) or with local time?

### DIFF
--- a/app/stac_api/models.py
+++ b/app/stac_api/models.py
@@ -8,6 +8,8 @@ from django.contrib.postgres.fields import ArrayField
 from django.core.exceptions import ValidationError
 from django.core.serializers.json import DjangoJSONEncoder
 from django.utils.translation import gettext_lazy as _
+from datetime import datetime
+
 
 # pylint: disable=fixme
 # TODO remove this pylint disable once this is done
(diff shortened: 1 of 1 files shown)


app/stac_api/models.py
```diff
@@ -248,14 +252,14 @@ def update_extent(self, item_properties_datetime):
...
-                raise ValidationError(_("Property geoadmin:variant not compatible with the"
-                                        "naming conventions."))
+        if self.summaries["geoadmin:variant"]: # TODO: quick and dirty fix
+            for variant in self.summaries["geoadmin:variant"]:
+                if not bool(re.search('^[a-zA-Z0-9]*$', variant)):
+                    logger.error(
+                        "Property geoadmin:variant not compatible with the naming conventions."
+                    )
+                    raise ValidationError(_("Property geoadmin:variant not compatible with the"
+                                            "naming conventions."))
```
ltshb, line 262: I've fixed this in PR #58 where the `self.summaries["geoadmin:variant"]` is then an empty list per default, so I would remove this.

ltshb (changes requested):

app/stac_api/models.py
```diff
@@ -230,13 +232,15 @@ def update_extent(self, item_properties_datetime):
             if self.extent["temporal"]["interval"][0][0] is None:
                 self.extent["temporal"]["interval"][0][0] = item_properties_datetime
                 self.save()
-            elif item_properties_datetime < self.extent["temporal"]["interval"][0][0]:
+            elif item_properties_datetime.replace(tzinfo=None) < datetime.strptime(
```
ltshb, line 235: I would not remove the tzinfo, but add it to the extent datetime. I would uses timezone aware times it makes usually life easier. From the spec the datetime should be displayed as UTC using iso format with timezone info from my understanding,

ltshb (changes requested):

PR #60, opened 2020-11-03
BGDIINF_SB-1333: Propagate the bbox to the collection

rebert:
added a function in the modell which updates the bbox of the collection when an item is being saved


app/stac_api/models.py
```diff
@@ -236,6 +237,26 @@ def update_extent(self, item_properties_datetime):
...
+        '''
+        if self.extent["spatial"]["bbox"]:
+            if self.extent["spatial"]["bbox"] is None:
+                self.extent["spatial"]["bbox"] = GEOSGeometry(geometry).extent
+            else:
+                qs = Item.objects.filter(collection_id=self.pk)
+                union_geometry = GEOSGeometry('Multipolygon EMPTY')
+                for item in qs:
+                    union_geometry = union_geometry.union(item.geometry)
+                self.extent["spatial"]["bbox"] = union_geometry.extent
```
boecklic, line 257: looping over all items in the collection whenever an item is saved will work, but will not be performant for more than a few houndred items (remember that we'll have collections with a few hundred-thousand items). I suggest to compare here just the extent new `geometry` with the existing extent on the collection and adapt if necessary.
I would keep this functionality in a dedicated mgmt command (`recalculate_collection_cache` or the like), where we can explicitly trigger the recalculation of these values from all items if necessary.

boecklic (changes requested):

app/stac_api/models.py
```diff
@@ -236,6 +237,26 @@ def update_extent(self, item_properties_datetime):
...
+        '''
+        updates the collection's spatial extent when an item is updated.
+        :param geometry: the geometry of the item
+        This function generates a new extent regarding all the items with the same
+        collection foreign key. If there is no spatial bbox yet, the one of the geometry of the
+        item is being used.
+        Warning: This function may be slow, when there are millions of items (because of the query)
+        '''
+        if self.extent["spatial"]["bbox"]:
+            if self.extent["spatial"]["bbox"] is None:
```
ltshb, line 250: Redundant if

app/stac_api/models.py
```diff
@@ -310,12 +331,10 @@ def update_properties_eo_gsd(self, asset_eo_gsd):
...
 
     def save(self, force_insert=False, force_update=False, using=None, update_fields=None):
-        # TODO: check if collection's bbox needs to be updated
-        # --> this could probably best be done with GeoDjango? (@Tobias)
-        # I leave this open for the moment.
-        self.collection.update_extent(self.properties_datetime)
+        self.collection.update_temporal_extent(self.properties_datetime)
 
         super().save(force_insert, force_update, using, update_fields)
+        self.collection.update_spatial_bbox(self.geometry)
```
ltshb, line 337: Should not this be done before the save as for the temporal extent ?

ltshb, line 337: I would call the update extent functions only if needed, meaning only if the `properties_datetime` and/or `geometry` changed.

app/stac_api/models.py
ltshb: From my understanding of the bbox the update is only needed in these cases:

1. New item
    a. new item bbox is outside of collection bbox => no loop needed for update
2. Item update
    a. old item inside collection bbox without neighbor border
        i. new item outside collection bbox => no loop needed for update
    b. old item inside collection bbox with neighbor border
        i. new item outside collection bbox => no loop needed for update
        ii. new item inside collection bbox and neighbor border changed => loop needed for update

To resume only the 4 use case needs an update of collection and only one needs a loop:
* 1.a => no loop needed
* 2.a.i => no loop needed
* 2.b.i => no loop needed
* 2.b.ii => loop needed

The question is how often an item geometry update would happen ? would it happen at all ? And then is the item.save() method blocking API request or admin page ?
When the item update geometry doesn't happen often and the save() method is not blocking, then I think the loop only when needed is fine. When it is blocking and we can make it asynchrone then it is also fine to have the loop. When it might be often updated we might want to search for a better loop algorithm. When the checks above are too complex and the item geometry update is a rare case then the loop asynchrone might be an acceptable solution in my opinion.

ltshb (dismissed):
See comment, maybe we should discuss the geometry update in a meeting.

rebert:
@ltshb @boecklic what about the idea to store a geometry field in the collection? with the current state of all geometries of the items involved? and using this geometry to compare and to adapt when an item geometry changes?

app/stac_api/models.py
rebert: @ltshb thanks for that, I lost a day (because I did not listen to your advice and tried an other soloution)

app/stac_api/models.py
```diff
@@ -229,6 +231,52 @@ def update_extent(self, item_properties_datetime):
...
+            if Polygon.from_bbox(GEOSGeometry(item_geom).extent).covers(self.extent_geometry):
+                self.extent_geometry = Polygon.from_bbox(GEOSGeometry(item_geom).extent)
+            # we need to iterat trough the items
+            else:
+                qs = Item.objects.filter(collection_id=self.pk).exclude(id=item_id)
+                union_geometry = GEOSGeometry(item_geom)
+                for item in qs:
+                    union_geometry = union_geometry.union(item.geometry)
+                self.extent_geometry = Polygon.from_bbox(union_geometry.extent)
+
```
boecklic, line 267: I think this case is not necessary (or do I miss sth?): if the current 'cached' extent is bigger than the extent of the new item, then we don't need to do anything I think.

app/stac_api/models.py
```diff
@@ -229,6 +231,52 @@ def update_extent(self, item_properties_datetime):
...
+        # delete, we need to iterate trough the items
+        if action == 'rm':
+            qs = Item.objects.filter(collection_id=self.pk).exclude(id=item_id)
+            union_geometry = GEOSGeometry('Multipolygon EMPTY')
+            for item in qs:
+                union_geometry = union_geometry.union(item.geometry)
+            self.extent_geometry = Polygon.from_bbox(union_geometry.extent)
```
boecklic, line 274: We can leave this case like this for the moment. We'll have to change this to an asynchronous, e.g. daily cleanup task. Imho this extent information on the collection is kind of a 'nice to have' info for a client system, (we should in any case aks inser if they will rely on it to be 'realtime' accurat). If not, we can add this to the spec saying that we update this info e.g. once a day.

app/stac_api/models.py
```diff
@@ -229,6 +231,52 @@ def update_extent(self, item_properties_datetime):
...
+            else:
+                qs = Item.objects.filter(collection_id=self.pk).exclude(id=item_id)
+                union_geometry = GEOSGeometry(item_geom)
+                for item in qs:
+                    union_geometry = union_geometry.union(item.geometry)
+                self.extent_geometry = Polygon.from_bbox(union_geometry.extent)
+
+        # delete, we need to iterate trough the items
+        if action == 'rm':
+            qs = Item.objects.filter(collection_id=self.pk).exclude(id=item_id)
```
boecklic, line 270: just as info, could also be done like this: `self.item_set.exclude(id=item_id)` (see https://docs.djangoproject.com/en/3.1/ref/models/relations/#related-objects-reference).

boecklic (changes requested):
nice with the geometry field, I have one remark in terms of functionality

app/stac_api/models.py
rebert: when the updated extent is smaller than the current collection extent, it could be, that the collection extent has to become smaller - we don't know -> so we have to iterate

app/stac_api/models.py
rebert: thanks

app/stac_api/models.py
boecklic: Yeah true, didnt think about that one

boecklic (approved):

PR #62, opened 2020-11-04
BGDIINF_SB-1395: Changed HTTP error message to json format

ltshb:
Now all Rest API HTTP error message follow the API spec and are json  format:
    
`{"code": int, "description": str}`
    
Exception are 404 NOT FOUND when given an invalid URL, e.g. http://data.geoadmin.ch/unknown-url => gives an HTML 404 response.


boecklic (approved):
Nice solution!

PR #63, opened 2020-11-04
BGDIINF_SB-1397: enable asset and item filters based on url

ltshb:
The assets and items endpoints did returned all assets and items without
filtering based on url collection and item.


app/stac_api/views.py
```diff
@@ -146,11 +150,12 @@ class AssetsList(generics.GenericAPIView):
...
-    # def get_queryset(self):
-    #     # filter based on the url
-    #     return Asset.objects.filter(
-    #         collection=self.kwargs['collection_name'], feature=self.kwargs['item_name']
-    #     )
+    def get_queryset(self):
+        # filter based on the url
+        return Asset.objects.filter(
+            collection__collection_name=self.kwargs['collection_name'],
+            feature__item_name=self.kwargs['item_name']
```
boecklic, line 157: filtering on the item_name should be sufficient, otherwise we might need a combined index for perf reasons.

boecklic (approved):

app/stac_api/views.py
ltshb: True, I'll remove the filter on collection name.

PR #65, opened 2020-11-04
BGDIINF_SB-1403: changing item geometry type to polygon

boecklic:
See ticket https://jira.swisstopo.ch/browse/BGDIINF_SB-1403 for reference why this was done.

Special note on the migrations: 
- they seem complicated but I wanted to have a smooth migration experience including data migration. Therefore it has to be done in several steps and cannot be further squashed.
- apparently this is a rare case of django failing to create correct migrations when renaming the `geometry` field to `geometry_old`: the index which is automatically created is not renamed as well, therefore the (mostly unnecessary) RAW SQL statement. When removing the field the index gets properly deleted.


.pylintrc
```diff
@@ -14,7 +14,8 @@ ignore=CVS,settings.py
 
 # Add files or directories matching the regex patterns to the blacklist. The
 # regex matches against base names, not paths.
-ignore-patterns=
+# ^[0-9]{4}_.*\.py$ : ignore migration files
+ignore-patterns=^[0-9]{4}_.*\.py$
```
ltshb, line 18: Just curious what were the linting issue in the migration file that you blacklisted them ?

app/tests/test_collections_endpoint.py
```diff
@@ -51,6 +52,7 @@ def test_single_collection_endpoint(self):
             msg="Returned data does not match expected data"
         )
 
+    @skip("skipping due to bug https://jira.swisstopo.ch/browse/BGDIINF_SB-1404")
```
ltshb, line 55: I would add a comment to the ticket to not forgot about this `skip`

ltshb (approved):
:+1:  see inline comment

app/tests/test_collections_endpoint.py
boecklic: done

.pylintrc
boecklic: When you do data migrations you have to 'generate' a representation of the model at the time of the migration (might be different to the current version in the code). This Model variable is typically uppercase (since the model itself is uppercase), and pylint doesn't like this.

app/stac_api/migrations/0012_auto_20201104_1647.py
```diff
@@ -0,0 +1,32 @@
...
+        migrations.RunSQL(
+            'ALTER INDEX IF EXISTS stac_api_item_geometry_id RENAME TO stac_api_item_geometry_old_id',
+            'ALTER INDEX IF EXISTS stac_api_item_geometry_old_id RENAME TO stac_api_item_geometry_id'
+        ),
+        migrations.AddField(
+            model_name='item',
+            name='geometry',
+            field=django.contrib.gis.db.models.fields.PolygonField(
+                default=
+                'SRID=2056;POLYGON((2317000 913000,3057000 913000,3057000 1413000,2317000 1413000,2317000 913000))',
```
rebert, line 28: WGS84

app/stac_api/migrations/0013_auto_20201104_1649.py
```diff
@@ -0,0 +1,39 @@
...
+
+# Forwares
+def convert_multipolygons_to_polygons(apps, schema_editor):
+    # We can't import the Person model directly as it may be a newer
+    # version than this migration expects. We use the historical version.
+    Item = apps.get_model('stac_api', 'Item')
+    for item in Item.objects.all():
+        wkt = wkt_w(dim=2).write(item.geometry_old.convex_hull).decode()
+        item.geometry = GEOSGeometry(wkt, srid=2056)
+        item.save()
```
rebert, line 18: WGS84

app/stac_api/models.py
```diff
@@ -24,6 +24,15 @@
     '2317000 1413000 0,'
     '2317000 913000 0)))'
 )
+BBOX_CH = (
+    'SRID=2056;'
```
rebert, line 28: WGS84

rebert (changes requested):
LGTM beside the decision to use wgs84 alover

hansmannj (approved):
:+1:

rebert (approved):

PR #69, opened 2020-11-05
Fix bgdiinf sb 1404 temporal extent

hansmannj:
* Temporal extent is no longer a JSON field.
* nested serializer for read-only implemented
* Tests seem to fail due to some datetime issue, still on it..


hansmannj:
Didn't even notice, that now I have the same error as described in the jira ticket. Because I focused on the output of the local tests. Now I've tried a populate_testdb again and noticed it. I am on it..

hansmannj:
The error described in the ticket seems to result from item's datetime properties to have a wrong format (they will be recognized as string and not a datetime). For testing I added a hackish way to convert them, e.g. 2019-01-01 will be 2019-01-01 at 00:00:00 o'clock now. This seems to work

app/stac_api/models.py
```diff
@@ -137,8 +139,18 @@ class Collection(models.Model):
...
+    # )
+    start_date = models.DateTimeField(editable=False, null=True, blank=True)
+    end_date = models.DateTimeField(editable=False, null=True, blank=True)
+    # the bbox field below has no meaning yet and was only introduced for testing
+    # while working on the temporal extent.
+    bbox = ArrayField(
+        ArrayField(models.FloatField(null=True, blank=True), null=True, blank=True),
+        null=True,
+        blank=True,
+        editable=False
```
boecklic, line 153: I think this is not necessary anymore since @rebert is using a geometry field, bbox can be deduced on serialization time from the geometry (or did you intend to 'cache' the calculation of the bbox from the geomtry in this field?)

app/stac_api/models.py
```diff
@@ -137,8 +139,18 @@ class Collection(models.Model):
     # furthermore GeoDjango and its functionality will be used for that.
     # TODO: overwrite items save() function accordingly
     # suggestions of fields to be auto-populated:
-    extent = models.JSONField(
-        default=get_default_extent_value, encoder=DjangoJSONEncoder, editable=False
+    # extent = models.JSONField(
+    #    default=get_default_extent_value, encoder=DjangoJSONEncoder, editable=False
+    # )
+    start_date = models.DateTimeField(editable=False, null=True, blank=True)
+    end_date = models.DateTimeField(editable=False, null=True, blank=True)
```
boecklic, line 146: maybe name them `start_datetime` and `end_datetime` (same naming as in the item). 

I'm just wondering if it would make sense to mark all these 'cache' fields with e.g. a `cache_*` prefix... just to clearly indicate which values are actually properties of the Model itself and which values are just precalculated (i.e. cached) values from other models.

app/stac_api/serializers.py
```diff
@@ -131,6 +131,33 @@ def update(self, instance, validated_data):
...
+
+    def to_representation(self, value):
+        bbox = {"bbox": [[value.bbox]]}  # probably one pair of brackets too much here?
+        return bbox
+
+
+class ExtentSerializer(serializers.Serializer):
+    spatial = ExtentSpatialSerializer(source="*")
+    temporal = ExtentTemporalSerializer(source="*")
+
```
boecklic, line 159: nice! I didn't know you can build serializers from other serializers!

boecklic (commented):
Nice! made a few comments (up for discussion)

app/stac_api/models.py
hansmannj: So far I did not really had anything in mind. The bbox field was just for test reasons (not filled with any functionality so far), as the "mother of all extents" has both, spatial and temporal. Just wanted to be able to have a correct extent in the end.

app/stac_api/serializers.py
hansmannj: It didn't work quite a while ago, when in the first trials start_date and end_date have been implemented in a flat way, too. I came along a description of nested serializers, but must have done something wring. So I switched to the JSONField then. But on the second try, after a short chat with @ltshb (thanks!) retried and it works :smile:

app/stac_api/models.py
hansmannj: makes sense, both. So it will be `cache_start_datetime` and `cache_end_datetime`

app/stac_api/models.py
```diff
@@ -304,7 +320,7 @@ def save(self, force_insert=False, force_update=False, using=None, update_fields
         # TODO: check if collection's bbox needs to be updated
         # --> this could probably best be done with GeoDjango? (@Tobias)
         # I leave this open for the moment.
-        self.collection.update_extent(self.properties_datetime)
+        self.collection.update_temporal_extent(self.properties_datetime)
 
         super().save(force_insert, force_update, using, update_fields)
```
rebert, line 325: do you need to hanlde super().delete() as well?

hansmannj: Good point, yes, the delete case is still to be implemented.

app/stac_api/models.py
hansmannj: Merging @rebert PR on this branch should solve it

app/stac_api/models.py

app/stac_api/models.py
hansmannj: Will do that in a new PR

rebert (approved):
as far as I can judge - LGTM

PR #75, opened 2020-11-09
BGDIINF_SB-1363: Fixed samples unittest

ltshb:
Fixing samples unittest

Currently one item sample has an issue and has been renamed to *.bck in order to be skipped, this needs to be done within a separate PR (see https://jira.swisstopo.ch/browse/BGDIINF_SB-1427)


app/stac_api/models.py
```diff
@@ -236,37 +238,50 @@ def update_bbox_extent(self, action, item_geom, item_id):
...
+                'Failed to update spatial extend in collection %s with item %s action=%s: %s',
+                self.collection_name,
+                item_name,
+                action,
+                error
+            )
+            raise GEOSException(
+                f'Failed to update spatial extend in colletion {self.collection_name} with item '
+                f'{item_name}: {error}'
+            )
```
boecklic, line 284: I usually prefer to just re-raise the original exception to keep the stacktrace with the error:
```
except GEOSException as error:
    logger.error(...)
    raise
```

app/stac_api/sample_data/swissMapRaster200/collection.json
```diff
@@ -1,38 +1,8 @@
...
-    },
-    "temporal": {
-      "interval": [
-        [
-          "2019-01-01T00:00:00Z",
-          null
-        ]
-      ]
-    }
-  },
```
boecklic, line 35: Pls don't just remove this part, it's needed by inser until we're on int with some examples. And also besides that, I think it's useful to have a 'should' version of the automatically generated stuff. If it's wrong, it should be corrected.

boecklic (changes requested):
see comments, we shouldn't change the example data in that way

app/stac_api/models.py
ltshb: Well the stack trace about is kept as well as the error, the error message is just enhanced to give a hint to the user what is failing otherwise the user doesn't have a clue about the error, the orignal GEOSException message doesn't help at all that's why I completed it with useful infos.

app/stac_api/sample_data/swissMapRaster200/collection.json
ltshb: I've reverted this commit and did fixed small issues in the samples. I also created a new Jira issue concerning a mismatch from the implementation and spec, see https://jira.swisstopo.ch/browse/BGDIINF_SB-1429

boecklic: :+1: (i've changed the ticket type to bug)

boecklic (approved):

PR #76, opened 2020-11-09
BGDIINF_SB-1371: Added start/end datetime item propertes serializer

ltshb:


app/stac_api/serializers.py
```diff
@@ -27,9 +28,22 @@ class NonNullModelSerializer(serializers.ModelSerializer):
...
 
     def to_representation(self, instance):
-        ret = super().to_representation(instance)
-        ret = OrderedDict(filter(lambda x: x[1] is not None, ret.items()))
-        return ret
+        obj = super().to_representation(instance)
+        return self._filter_null(obj)
+
+    @classmethod
+    def _filter_null(cls, obj):
```
boecklic, line 35: calling a classmethod with `self._cls_method` works but doesn't really make sense in my opinion. I would either
- make it a normal method
- call it with `NonNullModelSerializer._filter_null` (though this might cause troubles with inheritance)
- just include the functionality in the `to_representation` method
- make it a regular function (might be easier to write unit tests for)

boecklic (changes requested):

app/stac_api/serializers.py
ltshb: Well `NonNullModelSerializer._filter_null` would definitely not work with inheritance and is not a pythonic way in my opinion. I used a class method because I wanted to keep this functionality within the class. But I can do either one of the two solutions:

1. Made `_filter_null` a regular functions within `stac_api.utils` and call it `filter_null_from_dict`
2. Or made the `_filter_null` a regular functions inside `to_representation`

What do you prefer ?

boecklic: But it's how you'd have to call a `@classmethod`... both options work for me, but since the functionality is imho specifically linked to the serializer I'd keep it there

ltshb: > But it's how you'd have to call a `@classmethod`... both options work for me, but since the functionality is imho specifically linked to the serializer I'd keep it there (so 2.)

`@classmethod` can be called either by `self` or by `Class.` (see https://stackoverflow.com/questions/44228916/calling-classmethod-using-self and doc https://docs.python.org/3.7/library/functions.html#classmethod) is just a question of design both are correct but have different behavior with inheritance. I'll do option 2, I prefer it as well over 1.

boecklic: :+1: thanks for the hint. I knew it's possible but thought it's bad practice (or bad design), but apparently not ;-).

boecklic (approved):

PR #77, opened 2020-11-11
BGDIINF_SB-1412: filter items by bbox

rebert:
Filtering items by bbox

### DIFF
--- a/app/stac_api/views.py
+++ b/app/stac_api/views.py
@@ -4,6 +4,7 @@ from datetime import datetime
 from django.http import JsonResponse
 from django.shortcuts import get_object_or_404
 from django.utils.translation import gettext_lazy as _
+from django.contrib.gis.geos import Polygon
 
 from rest_framework import generics
 from rest_framework.exceptions import ValidationError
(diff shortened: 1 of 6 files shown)


app/stac_api/views.py
```diff
@@ -152,13 +153,24 @@ def get_queryset(self):
...
+            queryset = self.filter_by_bbox(queryset, bbox)
 
         if date_time:
             queryset = self.filter_by_datetime(queryset, date_time)
 
         return queryset
 
+    def filter_by_bbox(self, queryset, bbox):
+        try:
+            logger.debug(bbox)
```
ltshb, line 165: I would be a bit more verbose here in the log or remove it, e.g. 'filter with bbox %s', bbox.

app/stac_api/views.py
```diff
@@ -152,13 +153,24 @@ def get_queryset(self):
...
+            query_bbox_polygon = Polygon.from_bbox(bbox.split(','))
+        except:
```
ltshb, line 167: Never ever catch all exceptions ! Only catch known exceptions e.g. 'ValueError', see https://github.com/geoadmin/doc-guidelines/blob/master/PYTHON.md#7-error-handling---rules-of-thumb

ltshb (changes requested):
Exceptions handling needs to be corrected based on https://github.com/geoadmin/doc-guidelines/blob/master/PYTHON.md#7-error-handling---rules-of-thumb

app/stac_api/views.py
```diff
@@ -152,13 +153,24 @@ def get_queryset(self):
...
+            raise ValidationError(_('Invalid bbox query parameter, '
```
ltshb, line 168: also all error should be logged

app/stac_api/views.py
rebert: Thanks - I will aprove

app/stac_api/views.py
rebert: this was a develop logger - I removed it

ltshb (changes requested):
:+1:  however the openapi.yaml should be directly edited ! it is generated with `make build-spec`

ltshb (commented):
Very nice :+1: , I've noticed that unittest is missing. Because the query parameter is a user input, the unittest is quite important. We need to test all use cases in unittest, not only the good one but also the border case (e.g. bbox=foo or bbox= or bbox=2).
Another point is the ValidationError, which could be improved by giving a real example: `bbox=5.96,45.82,10.49,47.81`, this it gives a hint on the syntax.

ltshb (commented):
Maybe also change the example in the spec `components/schemas.yml` line 103 to reflect switzerland instead of NZ :-)

boecklic (changes requested):
Nice! Same comments about unit tests and openapi*.yaml as Brice

app/tests/test_items_endpoint.py
```diff
@@ -215,3 +216,32 @@ def test_items_endpoint_datetime_invalid_range_query(self):
...
+
+    def test_items_endpoint_bbox_valid_query(self):
+        # test bbox
+        response = self.client.get(
+            f"/{API_BASE}collections/{self.collections[0].collection_name}/items"
+            f"?bbox=5.96,45.82,10.49,47.81&limit=100"
+        )
+        json_data = response.json()
+        self.assertEqual(200, response.status_code, msg=get_description(json_data))
+        self.assertEqual(3, len(json_data['features']), msg="More than one item found")
```
ltshb, line 229: Message is wrong, it should be `More or less than 3 item found`

ltshb (approved):
:+1:  just a minor issue i unittest message

PR #78, opened 2020-11-11
BGDIINF_SB-1409: logging middleware

boecklic:
log every incoming request and response with all headers etc.


app/config/logging-cfg-local.yml
```diff
@@ -42,6 +42,7 @@ filters:
     exclude_keys:
       - request.headers.Authorization
       - request.headers.Proxy-Authorization
+      - request.headers.Cookie
```
ltshb, line 45: This needs also to be  added to other logging configuration

app/config/logging-cfg-local.yml
```diff
@@ -68,7 +69,7 @@ formatters:
 handlers:
   console:
     class: logging.StreamHandler
-    formatter: standard
+    formatter: json
```
ltshb, line 72: Please keep the local to use standard, the other two configurations uses json format.

ltshb (changes requested):

ltshb:
Nice middleware, only issues with the logging configuration. For local development we should keep as default the standard formater as it is easier to read from console. The drawback of standard formater is it can't print the extra but we should add this functionality in the logging-utilities then.

app/config/logging-cfg-local.yml
boecklic: reverted

app/config/logging-cfg-local.yml
boecklic: done

ltshb (approved):
:+1:

PR #80, opened 2020-11-11
fix_BGDIINF_SB-1333: improve performance when updating an item

rebert:
When updating an item, the bbox of the collection may change.

But only when the geometry of the item will change, this may be.

So this improvement checks if the geometry has changed.

### DIFF
--- a/app/stac_api/models.py
+++ b/app/stac_api/models.py
@@ -232,7 +232,10 @@ class Collection(models.Model):
 
         try:
             # insert (as item_id is None)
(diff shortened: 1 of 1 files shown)


ltshb (approved):
:+1:  
I would add a `logger.warning('Looping over all items to update spatial extent, this may take a while')` before each loop when updating the spatial extent. Maybe also add a `logger.debug()` for each actions (`insert`, `update` and `remove`).

PR #81, opened 2020-11-11
BGDIINF_SB-1430: bugfix for logging middleware

boecklic:
fixes the error in `make serve`


rebert (approved):
works 4 me

ltshb (approved):
:+1: 

PS: commit message could be better, explanation of the problem and ticket number ;-)

PR #82, opened 2020-11-11
BGDIINF_SB-1363: corrected samples and Feature eo:gsd type

ltshb:
Correction made by @p1d1d1 and @ltshb. 

Note I will update the SPEC in a separate PR.


boecklic (changes requested):
hard to tell the difference if it's mostly just intendation that changed. @ltshb yq can format json as well (with e.g. a standardized intendation). I think we should have a look at this

ltshb:
> hard to tell the difference if it's mostly just intendation that changed. @ltshb yq can format json as well (with e.g. a standardized intendation). I think we should have a look at this

I think one of the issue is that @p1d1d1 did changed the examples directly within github, so the yq won't help in this case. Reviewing can also be done per commit basis, this helps

boecklic (approved):
Thanks! I would wait with merging until we decided how to proceed with eo:gsd. I would include that in this PR as well, this will give a nice update info for inser.

boecklic (approved):
Nice. Are you sure that the migrations (in particular data migration) works this way? Cause you're removing the field and adding it again in the first migration, so I suspect the data migration won't have any effect (since the field was newly added and will always be empty). If you want do migrate data, you have to add the new field in the first migration (and maybe rename the old one), then do data migration, then in a third migration remove the old field.

ltshb:
> Nice. Are you sure that the migrations (in particular data migration) works this way? Cause you're removing the field and adding it again in the first migration, so I suspect the data migration won't have any effect (since the field was newly added and will always be empty). If you want do migrate data, you have to add the new field in the first migration (and maybe rename the old one), then do data migration, then in a third migration remove the old field.

It works I tested it, not sure that I perfectly understood the full functionality of them though. But from my understanding the first migration does changing the field type and with the second we get the previous Item model (before migation) in order to do the data migration. For the revert data migration I did it very simple by setting a `None` value which is accepted in both models, but I don't know how I could do it if we wanted to really revert back to an array. It could be actually a good learning case, but I would try to do it in another PR, what do you think @boecklic  ?

boecklic:
> > Nice. Are you sure that the migrations (in particular data migration) works this way? Cause you're removing the field and adding it again in the first migration, so I suspect the data migration won't have any effect (since the field was newly added and will always be empty). If you want do migrate data, you have to add the new field in the first migration (and maybe rename the old one), then do data migration, then in a third migration remove the old field.
> 
> It works I tested it, not sure that I perfectly understood the full functionality of them though. But from my understanding the first migration does changing the field type and with the second we get the previous Item model (before migation) in order to do the data migration. For the revert data migration I did it very simple by setting a `None` value which is accepted in both models, but I don't know how I could do it if we wanted to really revert back to an array. It could be actually a good learning case, but I would try to do it in another PR, what do you think @boecklic ?

I'm sure the migrations run without errors, but they won't perform the functionality you wanna have. With the first migration you're in fact not changing the data type, you're removing the field and adding it again with a different type. This means all existing values are lost. Hence the data migration doesn't have any effect. As described above, if you want to keep the data (and maybe convert it somehow), you have to do it in three steps. In the data migration you need to have access to both fields, the one with the old datatype and the one with the new datatype, and you just do the kind of type conversion that PSQL cant do automatically (casting list to single value and vice versa). Backward migration becomes obvious then as well, you just create a list with a single entry.

ltshb:
> > > Nice. Are you sure that the migrations (in particular data migration) works this way? Cause you're removing the field and adding it again in the first migration, so I suspect the data migration won't have any effect (since the field was newly added and will always be empty). If you want do migrate data, you have to add the new field in the first migration (and maybe rename the old one), then do data migration, then in a third migration remove the old field.
> > 
> > 
> > It works I tested it, not sure that I perfectly understood the full functionality of them though. But from my understanding the first migration does changing the field type and with the second we get the previous Item model (before migation) in order to do the data migration. For the revert data migration I did it very simple by setting a `None` value which is accepted in both models, but I don't know how I could do it if we wanted to really revert back to an array. It could be actually a good learning case, but I would try to do it in another PR, what do you think @boecklic ?
> 
> I'm sure the migrations run without errors, but they won't perform the functionality you wanna have. With the first migration you're in fact not changing the data type, you're removing the field and adding it again with a different type. This means all existing values are lost. Hence the data migration doesn't have any effect. As described above, if you want to keep the data (and maybe convert it somehow), you have to do it in three steps. In the data migration you need to have access to both fields, the one with the old datatype and the one with the new datatype, and you just do the kind of type conversion that PSQL cant do automatically (casting list to single value and vice versa). Backward migration becomes obvious then as well, you just create a list with a single entry.

Make sense thanks for the explanation. I'll leave so and try to do it better next time :-) I think you were right we should slowly start data migration to gain experience with it.

PR #83, opened 2020-11-12
BGDIINF_SB-1434: changing logging config for dev

boecklic:
changing logging config for dev to mitigate logging error


ltshb (approved):
It would be good to have in the commit message the reason of this change, it would help later on when looking at the git history.

rebert (approved):

ltshb:
This PR is out of date and not anymore relevant due to #103 so I suggest to simply close it do you agree @boecklic

PR #84, opened 2020-11-12
 BGDIINF_SB-1411: Show updated and created in the serializers

procrastinatio:


rebert (commented):
app/manage.py makemigrations to get the DDL

boecklic (approved):

ltshb (commented):
I would not add the `# datetime` comment to a `DateTimeField` the comment help for nothing, same for `# string`.
Test could also test if the updated/created value are valid datetime string (e.g. parse it with `stat_api.utils.fromisoformat()`)

PR #85, opened 2020-11-12
Fix_BGDIING_SB-1404 delete case

hansmannj:
Warning: This PR is a bit of a monster.

* It updated the `update_temporal_extent()` to contain also the DELETE case.
* For the sake of performance (on the cost of clearness though), querying all items is reduced to the absolute minimum (at least I hope so).
* special cases were considered and tested (e.g. an item, that was defined by a single `properties.datetime` can be updated to have a range of `properties_start_datetime` to `properties_end_datetime` and vice versa).
* comprehensive unit tests added
* in case of UPDATE checks are done, which values were updated (start_ or end_datetime) and only the minimum necessary update operations are conducted.

Any ideas welcome on how to make the code more clear. Same for the code of the unit tests.
I tried my best so far, but have been staring on the code for too long from too early to too late sometimes ;-)
Thanks :-)


hansmannj:
When trying to split the large `update_temporal_extent()` into smaller subfunctions I noticed, that the DELETE-case is not yet as performant, as could be. Should be a small thing and I'll commit an update to this PR today.

hansmannj:
> When trying to split the large `update_temporal_extent()` into smaller subfunctions I noticed, that the DELETE-case is not yet as performant, as could be. Should be a small thing and I'll commit an update to this PR today.

Done :-)

app/stac_api/models.py
```diff
@@ -197,28 +197,273 @@ def update_geoadmin_variants(self, asset_geoadmin_variant, asset_proj_epsg, asse
...
+                        logger.warning(
+                            'Looping (twice) over all items of collection %s,'
+                            'to update temporal extent, this may take a while',
+                            self.collection_name
+                        )
+                        qs = Item.objects.filter(collection_id=self.pk).exclude(id=item_id)
+                        # get earliest start_datetime or none, in case none exists
+                        # this should be the case, if the item that is currently
+                        # beeing updated is the only item in the collection or
+                        # no item has a range defined (no start_ and end_datetimes)
```
hansmannj, line 270: using `bool()` here, due to this:
"if a some_queryset has not yet been evaluated, but you know that it will be at some point, then using some_queryset.exists() will do more overall work (one query for the existence check plus an extra one to later retrieve the results) than using bool(some_queryset), which retrieves the results and then checks if any were returned."
(see: https://docs.djangoproject.com/en/3.0/ref/models/querysets/)

hansmannj:
Sorry, there still seem to be issues when items are converted from having a single datetime to a range or vice versa.
I am on it and will let you know when I solved it.
Sorry, I only noticed it now when cleaning up my unit tests.

hansmannj:
Alright, now everything should be ready.. won't touch it any more and won't look at it any more today.
I know, this is a perfect example of why small increments are better and adding more details later on makes perfect sense.

The tricky thing with this was, that items can either have a single datetime value or a range.
And hence they can (theoretically) be updated between the two states (either a single datetime --> updated to have a range, or having a range first and being updated to have a single datetime only).
Another tricky thing was, to only do as little loops (resp. queries) over all items, as absolutely necessary, for the sake of performance when a collections can have several 100 thousands of items.

app/stac_api/models.py
```diff
@@ -197,27 +190,469 @@ def update_geoadmin_variants(self, asset_geoadmin_variant, asset_proj_epsg, asse
...
+        :param item_id: the id of the item being treated
+        :param qs: (optional) queryset with all items of the collection excluding the one
+        being updated.
+        '''
+        logger.debug(
+            'Updating item %s in collection (adjusting start_datetime) %s',
+            item_id,
+            self.collection_name
+        )
+        if old_start_datetime != new_start_datetime:
```
hansmannj, line 250: Noticed that this can be removed, as it is redundant. This function is only called, when this condition is true (see line 546)

app/stac_api/models.py
```diff
@@ -197,27 +190,469 @@ def update_geoadmin_variants(self, asset_geoadmin_variant, asset_proj_epsg, asse
...
+        :param item_id: the id of the item being treated
+        :param qs: (optional) queryset with all items of the collection excluding the one
+        being updated.
+        '''
+        logger.debug(
+            'Updating item %s in collection (adjusting end_datetime) %s',
+            item_id,
+            self.collection_name
+        )
+        if old_end_datetime != new_end_datetime:
```
hansmannj, line 329: Noticed that this can be removed, as it is redundant. This function is only called, when this condition is true (see line 578)

app/stac_api/models.py
```diff
@@ -197,27 +190,469 @@ def update_geoadmin_variants(self, asset_geoadmin_variant, asset_proj_epsg, asse
                 "Error when updating collection's summaries values due to asset update."
             ))
 
-    def update_temporal_extent(self, item_properties_start_datetime, item_properties_end_datetime):
+    def insert_item(
+        self, old_start_datetime, new_start_datetime, old_end_datetime, new_end_datetime, item_id
+    ):
```
ltshb, line 195: I would removed unused parameter which don't make sense in this method, `old_*`. And aslo rename the method, we don't insert a item in collection here we update the temporal extent due to item insertion: `update_temporal_extent_insert` it is a bit a long name but much more clearer to the reader in my opinion

app/stac_api/models.py
```diff
@@ -197,27 +190,469 @@ def update_geoadmin_variants(self, asset_geoadmin_variant, asset_proj_epsg, asse
...
+        :param old_start_datetime: item's old value for properties_start_datetime
+        (in this case None, as it is a new item)
+        :param new_start_datetime: item's updated value for properties_start_datetime
+        :param old_end_datetime: item's old value for properties_end_datetime
+        (in this case None, as it is a new item)
+        :param new_end_datetime: item's updated value for properties_end_datetime
+        :param item_id: the id of the item being treated
         '''
-
+        logger.debug('Inserting item %s in collection %s', item_id, self.collection_name)
```
ltshb, line 207: Having the datetimes value in log might be useful in future

app/stac_api/models.py
```diff
@@ -197,27 +190,469 @@ def update_geoadmin_variants(self, asset_geoadmin_variant, asset_proj_epsg, asse
...
+                    # This should be the case, if the item that is currently
+                    # being updated is the only item in the collection or
+                    # no item has properties.datetime defined but all do have
+                    # only ranged (start_ and end_datetimes defined)
+                    if bool(qs.filter(properties_datetime__isnull=False)):
+                        earliest_datetime = qs.filter(
+                            properties_datetime__isnull=False
+                        ).earliest('properties_datetime').properties_datetime
+                    else:
+                        earliest_datetime = None
```
ltshb, line 285: This could also be a function: `earliest_datetime, earliest_start_datetime = get_earliest_datetime(queryset)`

app/stac_api/models.py
```diff
@@ -197,27 +190,469 @@ def update_geoadmin_variants(self, asset_geoadmin_variant, asset_proj_epsg, asse
...
+        if self.cache_end_datetime is None:
+            #first item in collection, as cache_start_datetime is None
+            self.cache_end_datetime = new_end_datetime
+        elif self.cache_end_datetime < new_end_datetime:
+            # new item starts after current collection's range ends
+            self.cache_end_datetime = new_end_datetime
+
+        self.save()
+
+    def update_item_start_datetime(
```
ltshb, line 224: method name is confusing; `update_extent_item_update_start_datetime`

app/stac_api/models.py
```diff
@@ -197,27 +190,469 @@ def update_geoadmin_variants(self, asset_geoadmin_variant, asset_proj_epsg, asse
...
+            elif new_start_datetime < self.cache_start_datetime:
+                # item's start_datetime did not define the left bound of the
+                # collection's temporal extent before update, which does not
+                # matter anyways, as it defines the new left bound after update
+                # and collection's start_datetime can be simply adjusted
+                self.cache_start_datetime = new_start_datetime
+
+        self.save()
+
+    def update_item_end_datetime(
```
ltshb, line 303: Same as above about name

app/stac_api/models.py
```diff
@@ -197,27 +190,469 @@ def update_geoadmin_variants(self, asset_geoadmin_variant, asset_proj_epsg, asse
...
+        end_datetime needs to be updated.
+        :param old_start_datetime: item's old value for properties_start_datetime
+        :param new_start_datetime: item's updated value for properties_start_datetime
+        :param old_end_datetime: item's old value for properties_end_datetime
+        :param new_end_datetime: item's updated value for properties_end_datetime
+        :param item_id: the id of the item being treated
+        :param qs: (optional) queryset with all items of the collection excluding the one
+        being updated.
+        '''
+        logger.debug(
```
ltshb, line 324: datetimes in log message could be useful

app/stac_api/models.py
```diff
@@ -197,27 +190,469 @@ def update_geoadmin_variants(self, asset_geoadmin_variant, asset_proj_epsg, asse
...
+                    # this should be the case, if the item that is currently
+                    # being updated is the only item in the collection or
+                    # no item has properties.datetime defined but all do have
+                    # only ranged (start_ and end_datetimes defined)
+                    if bool(qs.filter(properties_datetime__isnull=False)):
+                        latest_datetime = qs.filter(
+                            properties_datetime__isnull=False
+                        ).latest('properties_datetime').properties_datetime
+                    else:
+                        latest_datetime = None
```
ltshb, line 365: Same as above this could be regular function: `latest_datetime, latest_start_datetime = get_latest_datetime(queryset)

app/stac_api/models.py
```diff
@@ -197,27 +190,469 @@ def update_geoadmin_variants(self, asset_geoadmin_variant, asset_proj_epsg, asse
...
+                # item's end_datetime did not define the right bound of
+                # the collection's temporal extent before update, which
+                # does not matter anyways, as it defines the right bound
+                # after update and collection's end_date can be simply
+                # adjusted
+                self.cache_end_datetime = new_end_datetime
+
+        self.save()
+
+    def delete_item_start_datetime(
```
ltshb, line 383: Name confusing

app/stac_api/models.py
```diff
@@ -197,27 +190,469 @@ def update_geoadmin_variants(self, asset_geoadmin_variant, asset_proj_epsg, asse
...
+                ).earliest('properties_start_datetime').properties_start_datetime
+            else:
+                earliest_start_datetime = None
+
+            # get earliest datetime, or none in case none exists
+            if bool(qs.filter(properties_datetime__isnull=False)):
+                earliest_datetime = qs.filter(properties_datetime__isnull=False
+                                             ).earliest('properties_datetime').properties_datetime
+            else:
+                earliest_datetime = None
```
ltshb, line 435: Could use the same function as above ?

app/stac_api/models.py
```diff
@@ -361,27 +806,94 @@ def save(self, *args, **kwargs):  # pylint: disable=signature-differs
         # This is needed because save() is called during the Item.object.create() function without
         # calling clean() ! and our validation is done within clean() method.
         self.validate_datetime_properties()
+
         if self.properties_datetime is not None:
+            if self.__original_properties_datetime is not None:
+                # This is the case, when the value of properties.datetime has been
+                # updated
+                self.collection.update_temporal_extent(
+                    'up',
```
ltshb, line 815: What do you mean with "up" ? should we not know within the `item.save()` if this is an `insert` or `update`

app/stac_api/models.py
```diff
@@ -361,27 +806,94 @@ def save(self, *args, **kwargs):  # pylint: disable=signature-differs
...
             )
+
         # adding a new item means update the bbox of the collection
         if self.pk is None:
             self.collection.update_bbox_extent('insert', self.geometry, self.pk, self.item_name)
         # update the bbox of the collection only when the geometry of the item has changed
         elif self.geometry != self._original_geometry:
             self.collection.update_bbox_extent('update', self.geometry, self.pk, self.item_name)
-        super().save(*args, **kwargs)
+        super(Item, self).save(*args, **kwargs)
```
ltshb, line 869: Why changing the way we call super here ? Even this is the same, this is dangerous if we move this function in a subclass of Item, then it will still call the `Model.save()` and not the `Item.save()` method. 
`super().method()` should always be preferred to `super(subclass, self).method()`

app/tests/test_collections_model.py
```diff
@@ -0,0 +1,500 @@
...
+    y100 = utc_aware(datetime.strptime('0100-01-01T00:00:00Z', '%Y-%m-%dT%H:%M:%SZ'))
+    y9000 = utc_aware(datetime.strptime('9000-01-01T00:00:00Z', '%Y-%m-%dT%H:%M:%SZ'))
+    y200 = utc_aware(datetime.strptime('0200-01-01T00:00:00Z', '%Y-%m-%dT%H:%M:%SZ'))
+    y8000 = utc_aware(datetime.strptime('8000-01-01T00:00:00Z', '%Y-%m-%dT%H:%M:%SZ'))
+
+    def setUp(self):
+        self.collection = db.create_collection('collection-1')
+
+    def add_range_item(self, start, end, name):
+        item = Item.objects.create(
```
ltshb, line 28: When manually creating object, we should always then call `.full_clean()` method that runs validation

app/tests/test_collections_model.py
```diff
@@ -0,0 +1,500 @@
...
+from stac_api.utils import utc_aware
+
+import tests.database as db
+
+logger = logging.getLogger(__name__)
+
+API_BASE = settings.API_BASE
+
+
+class CollectionsModelTestCase(TestCase):
```
ltshb, line 17: Very good unittest cover :+1:  just a general comment regarding unittest I would not use docstring as it clutter the unittest output IMHO:
```
test_updating_collection (tests.test_collections_endpoint.CollectionsEndpointTestCase) ... ok
test_update_temporal_extent_datetime (tests.test_collections_model.CollectionsModelTestCase)
Tests if the collection's temporal extent is correctly updated, when ... ok
test_update_temporal_extent_datetime_mixed_items (tests.test_collections_model.CollectionsModelTestCase)
Tests, if collection's temporal extent is updated correctly when ... ok
test_update_temporal_extent_deletion_datetime_item (tests.test_collections_model.CollectionsModelTestCase)
Two items with singel datetime values are added and one is deleted ... ok
test_update_temporal_extent_deletion_last_datetime_item (tests.test_collections_model.CollectionsModelTestCase)
An item is added to the collection and deleted again afterwards. ... ok
test_update_temporal_extent_deletion_last_range_item (tests.test_collections_model.CollectionsModelTestCase)
An item is added to the collection and deleted again afterwards. ... ok
test_update_temporal_extent_deletion_range_item (tests.test_collections_model.CollectionsModelTestCase)
Two items are added to the collection and one is deleted afterwards. ... ok
test_update_temporal_extent_range (tests.test_collections_model.CollectionsModelTestCase)
Tests if the collection's temporal extent is correctly updated, when ... ok
```
Uses at best regular comment which are not added to the unittest output

ltshb (commented):
Code has been improved :+1:  but still some improvement possible.
A general remark about comments; `comments are good, but too many comments kills comments`. Too much comments tends not to be read and even worse are usually not updated with the code. Code should be self explanatory.

app/stac_api/models.py
```diff
@@ -229,13 +664,10 @@ def update_bbox_extent(self, action, item_geom, item_id, item_name):
...
         item is being used.
         '''
-
         try:
             # insert (as item_id is None)
             if action == 'insert':
-                logger.debug('Updating collections extent_geometry'
-                             ' as a item has been inserted'
-                             )
+                logger.debug('Updating collections extent_geometry' ' as a item has been inserted')
```
boecklic, line 670: It might be a good idea to add the `collection_name` as an `extra` argument to all log entries where available (as either `collection_name` or simply `collection`) this will allow to easily filter message belonging to a certain collection (that might be causing troubles)

app/tests/test_collections_model.py
boecklic: Is it possible to configure the test runner differently? Cause I like the ocmments

boecklic (approved):
Thanks for the extensive and careful work! 

Somehow this still feels rather complicated. I would leave the mechanism like this for the moment, but in a next phase we'll need to think about a good pattern how to solve this asyncronously and in a simpler manner.

app/stac_api/models.py
hansmannj: Good point. I guess this was because I was following this example here (which was a good catch from @rebert):
https://stackoverflow.com/questions/1355150/when-saving-how-can-you-check-if-a-field-has-changed
You are right, this might cause unexpected behavior, I'll change it today when working on all the comments :+1:

app/tests/test_collections_model.py
ltshb: I don't think so, there isn't much to configure in the original test runner, or its not documented. Using regular comment `# ` instead of docstring works thouh

hansmannj:
Thanks for all the comments and good points you brought up.
I am currently working on the things discussed this morning and will convert to draft for that time.
Also added a few more unit tests.

app/stac_api/models.py
hansmannj: removed redundant if statement

app/stac_api/models.py
hansmannj: removed redundant if statement

app/stac_api/models.py
hansmannj: added datetime values (where applicable) and extra parameter with collection_name in all loggers

app/stac_api/models.py
hansmannj: done

app/stac_api/models.py
hansmannj: renamed the method's here and in all other places

app/stac_api/models.py
hansmannj: done

app/stac_api/models.py
hansmannj: done

app/stac_api/models.py
hansmannj: renamed

app/stac_api/models.py
hansmannj: You are right. I implemented this in analogy to the updating of the spatial extent (I think, at least at some state it was the same there, too). But obviously spatial extent has evolved in the meantime and I also updated the logic for differentiating between `insert` and `update` now.

app/stac_api/models.py
hansmannj: done

app/tests/test_collections_model.py
hansmannj: replaced with regular comments now

app/tests/test_collections_model.py
hansmannj: done

app/stac_api/models.py
hansmannj: Yes, but unfortunately this snippet cannot exactly be reused elsewhere, so it does not make sense to me, to create an extra function for it. Also, the function, to which it currently belongs, is quite short already in current state. So I'll leave it as it is, if OK with you?

app/stac_api/models.py
hansmannj: see my comment above. Unfortunately, there are some important differences in the code (at least after my latest commits, fixing some small issues). So the function could not be reused elsewhere. And as the function, where the snippet is currently located, is quite short already, I'd leave it as it is, if OK with you.

app/stac_api/models.py
hansmannj: see above and below

app/stac_api/models.py
hansmannj: done

app/stac_api/models.py
```diff
@@ -119,7 +121,7 @@ class Provider(models.Model):
     roles = ArrayField(models.CharField(max_length=9), help_text=_(
         "Comma-separated list of roles. Possible values are {}".format(
             ', '.join(allowed_roles)))
-    )
+                      )
```
ltshb, line 124: Seems some formatting issue here

ltshb (approved):
👍

PR #88, opened 2020-11-12
BGDIINF_SB-1427: Validate geometry fields

rebert:
A Validator has been defined, that validates geometries


hansmannj (approved):
:+1:
Nice, looks good to me

ltshb (commented):
:+1:  very nice, but I would add some model unittest to make sure that it work as expected. Unittest should cover also boarder cases (see test_item_model.py as example).

app/stac_api/models.py
```diff
@@ -67,6 +67,20 @@ def float_in(flt, floats, **kwargs):
     return np.any(np.isclose(flt, floats, **kwargs))
 
 
+def validate_valid_geometry(geometry):
```
boecklic, line 70: Detail: `validate_geometry` is probably sufficient as function name, I think it's obvious that we don't want to validate for 'invalid' geometries ;-)

boecklic (approved):

app/tests/test_item_model.py
```diff
@@ -127,3 +128,50 @@ def test_item_property_eo_gsd(self):
...
+
+    def test_item_create_model_empty_geometry(self):
+        # empty geometry should not be allowed
+        with self.assertRaises(ValidationError):
+            item = Item(
+                collection=self.collection,
+                properties_datetime=utc_aware(datetime.utcnow()),
+                item_name='item-empty', geometry=GEOSGeometry(
+                        'POLYGON EMPTY'
+                )
```
ltshb, line 165: code style, don't forget to run `make format` or even better integrate yapf in your IDE

PR #89, opened 2020-11-12
BGDIINF_SB-1382 made admin page also accessible under API_BASE path/admin/

hansmannj:


rebert (changes requested):
It works fine. But when running the make serve, I get the warning:
?: (urls.W005) URL namespace 'admin' isn't unique. You may not be able to reverse all URLs in this namespace

not sure if it has to be fixed:
https://stackoverflow.com/questions/53473027/django-warning-urls-w005-url-namespace-isnt-unique

hansmannj:
Thanks @rebert 
Yeah, I am also not sure, if this is exactly what was asked for in BGDIINF_SB-1382 yet.
I'll have a look at it

boecklic (approved):

rebert (approved):
lgtm

PR #90, opened 2020-11-13
fixed issue with update of the spatial extent

hansmannj:
When working on unit tests for updating the `*temporal* extent`, I noticed problems of the `*spatial* extent`, when the last item of a collection is deleted. This PR is intended to fix it ;-)


rebert (approved):
thanks @hansmannj

PR #92, opened 2020-11-16
BGDIINF_SB-1407 changing Provider-Collection relation

boecklic:
The `Provider` model has no a FK to a collection instead of an M2M relation. This is simpler to manage (validation in particular) but will introduce some redundancy.


app/stac_api/models.py
```diff
@@ -106,12 +106,23 @@ def __str__(self):
...
+        related_name='providers',
+        related_query_name='provider'
+    )
     name = models.CharField(blank=False, max_length=200)
     description = models.TextField(blank=True, null=True)
-    roles = ArrayField(models.CharField(max_length=9))
-    # possible roles are licensor, producer, processor or host. Probably it might sense
+    # possible roles are licensor, producer, processor or host
+    roles = ArrayField(models.CharField(max_length=9), help_text=_(
+        "Comma-separated list of roles. Possible values are 'licensor', "
```
hansmannj, line 119: Probably it would make sense to move the `allowed_roles` from the Provider's clean()-method up here and do something like:
```python
`    allowed_roles = ['licensor', 'producer', 'processor', 'host']
    # possible roles are licensor, producer, processor or host
    roles = ArrayField(models.CharField(max_length=9), help_text=_(
        "Comma-separated list of roles. Possible values are: " +
        ', '.join(allowed_roles)))
```

Just in case the definition of `allowed_roles` should change, it would be automatically updated in the help text.

(in the clean() a `self` would need to be added before the `allowed_roles`:
`if role not in self.allowed_roles:`)

hansmannj (approved):
Nice :smile: 
Looks good to me

app/stac_api/models.py
boecklic: :+1: I changed this

PR #94, opened 2020-11-18
BGDIINF_SB-1368: Added collection, item and asset name validation

ltshb:
This is required because these name are used in the url and we don't
want invalid url characters.


app/stac_api/models.py
```diff
@@ -146,9 +156,10 @@ def clean(self):
 
 
 class Collection(models.Model):
-    created = models.DateTimeField(auto_now_add=True)  # datetime
-    updated = models.DateTimeField(auto_now=True)  # datetime
-    description = models.TextField()  # string  / intentionally TextField and
+    collection_name = models.CharField(unique=True, max_length=255, validators=[validate_name])
```
boecklic, line 159: maybe keep the comment about the different naming in spec ('id') and model ('name') (bc. 'id' is used )

boecklic (approved):

app/stac_api/models.py
ltshb: Done

PR #95, opened 2020-11-18
BGDIINF_SB-1451: Fixed admin page crash when creating new Asset

ltshb:
When creating an Asset from the admin page, django crashed because no collection was given.
    
The collection foreign key is redundant because we have the item foreign key which contains the collection foreign key. Therefore removed the collection FK.

When creating an asset from the admin page with an empty geoadmin:variant, django crashed due to passing None to a regex.  Now instead of doing the validation in the clean() we do it with a validator which always get a string (migth be empty when None).


boecklic (approved):

rebert (approved):
nice

PR #97, opened 2020-11-19
Renaming of collection_name, asset_name and item_name

ltshb:
* Renamed Collection model field collection_name to name
* Renamed Item model field item_name to name
* Renamed Asset model field asset_name to name

These names were redundant.


boecklic (approved):
Great cleanup, thanks!

rebert (approved):

PR #99, opened 2020-11-23
Moved collection and item links to be inlined in admin page

ltshb:
Now collection and item links are inlined in the admin page like the
providers. This way we have a consistent admin interface.

### DIFF
--- a/app/stac_api/admin.py
+++ b/app/stac_api/admin.py
@@ -9,10 +9,7 @@ from .models import Item
 from .models import ItemLink
 from .models import Provider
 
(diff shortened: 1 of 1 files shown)


boecklic (approved):
PR without Jira ticket? :stuck_out_tongue: 

cool!

ltshb:
> PR without Jira ticket? stuck_out_tongue
> 
> cool!

I know, but there isn't any ticket for this. It took me only few minutes to do it and thought ticket was not needed, but I can do one if you think it is best to have one ?

PR #100, opened 2020-11-23
BGDIINF_SB-1456: remove stac extension from collections

hansmannj:
* on collection level, stac_extensions is now []
* sample `collection.json` of the two sample collections is corrected accordingly

### DIFF
--- a/app/stac_api/models.py
+++ b/app/stac_api/models.py
@@ -47,10 +47,9 @@ DEFAULT_SUMMARIES_VALUE = {"eo:gsd": [], "geoadmin:variant": [], "proj:epsg": []
 
 
 def get_default_stac_extensions(is_collection=False):
-    if is_collection: # pylint: disable=no-else-return
+    if is_collection:
         return list()
-    else:
-        return list(DEFAULT_STAC_EXTENSIONS.values())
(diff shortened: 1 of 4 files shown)


app/stac_api/models.py
```diff
@@ -46,8 +46,11 @@
 DEFAULT_SUMMARIES_VALUE = {"eo:gsd": [], "geoadmin:variant": [], "proj:epsg": []}
 
 
-def get_default_stac_extensions():
-    return list(DEFAULT_STAC_EXTENSIONS.values())
+def get_default_stac_extensions(is_collection=False):
+    if is_collection: # pylint: disable=no-else-return
+        return list()
+    else:
+        return list(DEFAULT_STAC_EXTENSIONS.values())
```
ltshb, line 53: Don't disable pylint here simply write code accordingly to it:
```
if is_collection:
    return list()
return list(DEFAULT_STAC_EXTENSIONS.values())
```

ltshb (changes requested):
👍  just fixes pylint issue see comment

app/stac_api/models.py
hansmannj: Makes sense :-)

ltshb (approved):
:1st_place_medal:

PR #101, opened 2020-11-23
BGDIINF_SB-1402: Moved landing page into DB

ltshb:
Now the landing page is part of the DB and aslo uses the Django rest
framework.


app/stac_api/serializers.py
```diff
@@ -85,6 +89,75 @@ def data(self):
...
+            ]),
+            OrderedDict([
+                ("rel", "service-doc"),
+                ("href", request.build_absolute_uri(f"/{api_base}api.html")),
+                ("type", "text/html"),
+                ("title", "The API documentation"),
+            ]),
+            OrderedDict([
+                ("rel", "service-desc"),
+                ("href", request.build_absolute_uri(f"/{api_base}openapi.yaml")),
```
boecklic, line 135: this is an error in the current landing page, should be `/{api_base}/static/openapi.yaml` and `/{api_base}/static/api.html`

app/stac_api/serializers.py
```diff
@@ -85,6 +89,75 @@ def data(self):
...
+        # links already uses OrderedDict, this way we keep consistency between auto link and user
+        # link
+        representation['links'][:0] = [
+            OrderedDict([
+                ('rel', 'self'),
+                ('href', request.build_absolute_uri(f'/{api_base}')),
+                ("type", "application/json"),
+                ("title", "This document"),
+            ]),
+            OrderedDict([
```
boecklic, line 127: I like this read-only serializer, but don't see the point of using django-solo then. Or did miss sth?

boecklic (changes requested):
As mentioned in the comments: what's the point of using the singleton landing page and then using the read-only serializer injecting all the necessary values? If we can get away with just the serializer I'd prefer that.

Sidenote: I don't like that migration files are edited by format or sth. I think they should be untouched by any auto-formatter/linter

app/stac_api/serializers.py
ltshb: Ok I'll change it and also in the spec.

app/stac_api/serializers.py
ltshb: This is not a 100% read only serializer, the `id`, `description` and `title` can be edited. It also have the capability to add other links, like external links for license for example.
I could make it 100% read only and remove the django-solo stuff but then nothing is editable anymore and we loose the flexibility of adding other links. I just thought that having it not 100% static we be a benefit, as we already changed this page twice :-). I let you decide if we keep it as is or make it 100% read only ?

ltshb:
> As mentioned in the comments: what's the point of using the singleton landing page and then using the read-only serializer injecting all the necessary values? If we can get away with just the serializer I'd prefer that.
> 
> Sidenote: I don't like that migration files are edited by format or sth. I think they should be untouched by any auto-formatter/linter

About the migration files I would agree if they would always be generated, but in the future we will have to do data migration that are not auto generated and there is also the case of special migration were they need to be done manually. Due to this I don't see the point of not formatting/linting them, otherwise we might just skip the formatting and linting completely in my opinion.

app/stac_api/serializers.py
ltshb: Done

app/stac_api/serializers.py
boecklic: Ok, let's keep it for the moment. If we see that it will be more or less stable we can still make it 100% static (or if get troubles with django-solo)

boecklic:
> > As mentioned in the comments: what's the point of using the singleton landing page and then using the read-only serializer injecting all the necessary values? If we can get away with just the serializer I'd prefer that.
> > Sidenote: I don't like that migration files are edited by format or sth. I think they should be untouched by any auto-formatter/linter
> 
> About the migration files I would agree if they would always be generated, but in the future we will have to do data migration that are not auto generated and there is also the case of special migration were they need to be done manually. Due to this I don't see the point of not formatting/linting them, otherwise we might just skip the formatting and linting completely in my opinion.

Well, linting/formatting for general code is certainly useful and we must keep it. I just questioned changing of automatically generated files. But I can live with formatting them as well...

boecklic (approved):

app/stac_api/serializers.py
ltshb: :+1:

PR #102, opened 2020-11-23
BGDIINF_SB-1384: Changed 'name' attribute to 'id' in admin page and uses fieldset for item properties

ltshb:
The collection, item and asset attribute name is called 'id' on the spec
therefore uses the verbose_name for these attribute to set the correct
name in the admin page.

Also grouped the item properties into a fieldset.


rebert (approved):
LGTM

PR #105, opened 2020-11-23
BGDIINF_SB-1384: Show read only fields in admin page

rebert:


app/stac_api/admin.py
```diff
@@ -53,7 +63,8 @@ class ItemAdmin(admin.ModelAdmin):
                     'properties_datetime',
                     'properties_start_datetime',
                     'properties_end_datetime',
-                    'properties_title'
+                    'properties_title',
+                    'properties_eo_gsd'
```
ltshb, line 67: This is a readonly field and is not allowed to be here

rebert: isn't it the task to show them?

app/stac_api/templates/admin/openlayers.js
```diff
@@ -0,0 +1,203 @@
...
+    var options = {
+{% autoescape off %}{% for item in map_options.items %}      '{{ item.0 }}' : {{ item.1 }}{% if not forloop.last %},{% endif %}
+{% endfor %}{% endautoescape %}    };{% endblock %}
+    // The admin map for this geometry field.
+    {% block map_creation %}
+    {{ module }}.map = new OpenLayers.Map('{{ id }}_map', options);
+    // Base Layer
+
+     // Standard WMTS options for GeoAdmin...
+    {{ module }}.layers.base = {% block base_layer %}new OpenLayers.Layer.WMTS(
```
rebert, line 119: @procrastinatio is that ok. And do you know how to treat the referer - I get a 403 from WMTS? when developing localy

app/config/settings_prod.py
```diff
@@ -148,6 +148,7 @@
 STATIC_ROOT = os.path.join(BASE_DIR, 'var/www/stac_api/static_files')
 STATICFILES_DIRS = [
     BASE_DIR / "spec/static",
+    BASE_DIR / "app/stac_api/templates"
```
boecklic, line 151: I think the templates itself don't need to be in staticfiles. If it's just for the `*.js` file, pls move the `*.js` file to `stac_api/static`, then it should automatically be found by Django and you can remove this entry.

app/stac_api/admin.py
```diff
@@ -30,8 +31,21 @@ class CollectionLinkInline(admin.TabularInline):
 
 
 @admin.register(Collection)
-class CollectionAdmin(admin.ModelAdmin):
+class CollectionAdmin(admin.GeoModelAdmin):
+    Collection._meta.get_field('extent_geometry').editable = True  # to show the field
```
boecklic, line 35: Hmm, I think this will lead to problems if you change this on class level. I think you can either try to
- define the field in the admin with `editable=True` (e.g. sth like `extent_geometry = models.PolygonField(..., editable=True)
- or overwrite this info in the constructor of the class 
- or change it in the model (why is set to `editable=False` there if we need it editable=True?)

app/stac_api/admin.py
```diff
@@ -30,8 +31,21 @@ class CollectionLinkInline(admin.TabularInline):
...
+    modifiable = False  # to prohibit editing extent_geometry
```
boecklic, line 36: this should apply to the `extent_geometry` field only I guess? What's the behaviour if you add it on `CollectionAdmin` class level?

app/stac_api/admin.py
boecklic: If we can display these read-only fields in the admin it would be nice. If it's not easily possible and needs ugly hacks we leave it.

boecklic (changes requested):
Nice! Added a few comments inline

app/stac_api/admin.py
ltshb: That's sounds definitely a hack here. In the model this field should not be editable as well as in admin page. I agree with @boecklic having read only field in admin would be nice, but I would only do it if django allows it without any hack, so if there is any option in admin class to display non editable fields I would definitely leave it aside.

app/config/settings_prod.py
rebert: ok, I thought that we need this that manager.py collectstatic works

app/stac_api/admin.py
rebert: yes it is a ugly hack to show the extent. But we need it editable=False in the model, as it is just an aggregation from items

rebert: there is the option to display editable=False with readonly_fields = ['list_with_fields']. But for the PolygonField this results in a Testfield: Polygon(.....)... The hack is tho show the extent in the map.

ltshb: Have a look at https://docs.djangoproject.com/en/3.1/ref/contrib/admin/#django.contrib.admin.ModelAdmin.readonly_fields and https://stackoverflow.com/questions/48067882/django-admin-making-a-required-field-read-only. It seems there is a way to display read only fields without hack

rebert: @ltshb I did this already in the PR line: 62

app/stac_api/admin.py
hansmannj: `properties_eo_gsd` will disappear from the Item class, as soon as https://github.com/geoadmin/service-stac/pull/107 is merged ;-)

app/stac_api/admin.py
```diff
@@ -44,8 +45,21 @@ class CollectionLinkInline(admin.TabularInline):
...
+    readonly_fields = ['cache_start_datetime',
+                       'cache_end_datetime',
+                       'summaries'
+                       ]
     inlines = [ProviderInline, CollectionLinkInline]
+    # customisation of the geometry field
+    # https://docs.djangoproject.com/en/1.10/_modules/django/contrib/gis/admin/options/
+    map_template = finders.find('admin/ol_swisstopo.html')  # custom swisstopo
+    if os.environ['APP_ENV'] == 'local':
```
ltshb, line 58: `APP_ENV` is part of the django settings and should be read from it and not directly from os.environ

app/stac_api/admin.py
```diff
@@ -67,8 +82,16 @@ class ItemAdmin(admin.ModelAdmin):
...
-                    'properties_title'
+                    'properties_title',
+                    'properties_eo_gsd'
                 )
             }
         ),
     )
+    # customisation of the geometry field
+    map_template = finders.find('admin/ol_swisstopo.html')  # custom swisstopo
+    if os.environ['APP_ENV'] == 'local':
```
ltshb, line 93: `APP_ENV` is part of the django settings and should be read from it and not directly from os.environ

ltshb (changes requested):
Item property  eo:gsd needs to be removed as it interfere with #107.
The collection extent geometry is buggy (see comment)

app/stac_api/models.py
```diff
@@ -216,7 +216,7 @@ class Collection(models.Model):
     extent_geometry = models.PolygonField(
         default=None,
         srid=4326,
-        editable=False,
+        editable=True,
```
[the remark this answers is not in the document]
rebert: Thanks for testing! In theory it shoud work: https://django.fun/docs/django/en/3.0/ref/contrib/gis/admin/#django.contrib.gis.admin.GeoModelAdmin.modifiable

ltshb (approved):
:+1:

PR #106, opened 2020-11-24
BGDIINF_SB-1401: Fixed DEBUG=True prod image and renamed docker target test

ltshb:
The docker production image could not be started with the DEBUG flag set
to true. This was due to the fact that the DEBUG=True required some
django app that were only configured in the settings_dev.py and not in
prod settings.

Now in the prod image the DEBUG flag is forced to false to avoid issue
when the flag is externally set (e.g. deploying a production image on
dev staging). Anyway we should never deploy an image on production with
django DEBUG enabled.

Also renamed the docker image target test to debug as this target is not
olny used for testing in the CI but also used on the k8s dev staging.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,8 +20,8 @@ SETTINGS_TIMESTAMP = $(TIMESTAMPS)/.settins.timestamp
 DOCKER_BUILD_TIMESTAMP = $(TIMESTAMPS)/.docker-test.timestamp
 
 # Docker variables
-DOCKER_IMG_LOCAL_TAG = swisstopo/$(SERVICE_NAME):latest
-DOCKER_IMG_LOCAL_TAG_TEST = swisstopo/$(SERVICE_NAME):latest-dev
+DOCKER_IMG_LOCAL_TAG = swisstopo/$(SERVICE_NAME):$(USER).latest
+DOCKER_IMG_LOCAL_TAG_TEST = swisstopo/$(SERVICE_NAME):$(USER).latest-dev
 
 # Find all python files that are not inside a hidden directory (directory starting with .)
 PYTHON_FILES := $(shell find $(APP_SRC_DIR) -type f -name "*.py" -print)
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -20,8 +20,8 @@ SETTINGS_TIMESTAMP = $(TIMESTAMPS)/.settins.timestamp
 DOCKER_BUILD_TIMESTAMP = $(TIMESTAMPS)/.docker-test.timestamp
 
 # Docker variables
-DOCKER_IMG_LOCAL_TAG = swisstopo/$(SERVICE_NAME):latest
-DOCKER_IMG_LOCAL_TAG_TEST = swisstopo/$(SERVICE_NAME):latest-dev
+DOCKER_IMG_LOCAL_TAG = swisstopo/$(SERVICE_NAME):$(USER).latest
+DOCKER_IMG_LOCAL_TAG_TEST = swisstopo/$(SERVICE_NAME):$(USER).latest-dev
```
boecklic, line 24: Hmm, this doesn't play well with `docker-compose.yml`, it would be nice to be able to test the `docker-compose.yml` locally

boecklic (changes requested):

Makefile
ltshb: Well this tag is only used when building locally and doesn't have anything to do with docker-compose.yml. This latter uses its own build system and tag, therefore it shouldn't be an issue. But I'll test it to be sure.

ltshb: @boecklic I tried the ``docker-compose up` locally and it worked, so I don't really understand the issue here with using the `$(USER)` in makefile for the `dockerbuild` targets ? The advantage of using the `$(USER)` in local tag is that your can still then manually push it to dockerhub in order to test the image on k8s dev and it gives a hint to people from where come this local image.

boecklic: It's just that docker-compose will download the `:latest-dev` and not the locally built image. If you need to adapt docker-compose (cause e.g. you change sth. in the Dockerfile) you'll want that it uses the locally built image. But we can leave it as you have it now, we just need to remember that when adapting Dockerfile and testing if the CI will work with docker-compose will require temporary manual changes

boecklic (approved):

PR #107, opened 2020-11-24
BGDIINF_SB-1410: stac_extension to [] and no eo-information in item properties

hansmannj:
* set stac_extension to [] for collections and items
* removed eo:gsd from item properties
* adapted sample data
* added migrations file for removal properties_eo_gsd from item model
* renamed `update_geoadmin_variants()` to `update_summaries()`


hansmannj:
Not sure if the specs need to be changed w.r.t. the stac_extensions, that are empty per default.
Probably not, as the specs are more general here.

app/stac_api/models.py
```diff
@@ -36,22 +36,13 @@
 # instance so that it's not shared between all field instances.
 # HINT: Use a callable instead, e.g., use `list` instead of `[]`.""
 # <end quote>
```
ltshb, line 38: This comment (line 29-38) should be removed

app/stac_api/models.py
```diff
@@ -237,7 +228,7 @@ class Collection(models.Model):
     def __str__(self):
         return self.name
 
-    def update_geoadmin_variants(self, asset_geoadmin_variant, asset_proj_epsg, asset_eo_gsd):
+    def update_summaries(self, asset_geoadmin_variant, asset_proj_epsg, asset_eo_gsd):
```
ltshb, line 231: Good catch :-)

app/stac_api/models.py
```diff
@@ -633,17 +594,6 @@ def __str__(self):
...
+        self.item.collection.update_summaries(self.geoadmin_variant, self.proj_epsg, self.eo_gsd)
 
         super().save(*args, **kwargs)
-
-    def delete(self, *args, **kwargs):  # pylint: disable=signature-differs
-        # It is important to use `*args, **kwargs` in signature because django might add dynamically
-        # parameters
-        if self.eo_gsd is not None:
-            self.item.update_properties_eo_gsd(self, deleted=True)
-        super().delete(*args, **kwargs)
```
ltshb, line 649: The delete case for the eo:gsd in collection.summaries was not implemented. As it has an impact on Performance, I would not do it in this PR, but I would suggest to leave the delete method as follow:
```python
def delete(self, *args, **kwargs):  # pylint: disable=signature-differs
    # pylint: disable=fixme
    # TODO implement delete case for updating collection.summaries [BGDIINF_SB-1462]
    super().delete(*args, **kwargs)
```
I've created  a ticket, see https://jira.swisstopo.ch/browse/BGDIINF_SB-1462

app/stac_api/serializers.py
```diff
@@ -359,16 +359,8 @@ class ItemsPropertiesSerializer(serializers.Serializer):
...
     )
-    eo_gsd = serializers.FloatField(source='properties_eo_gsd', allow_null=True, read_only=True)
     title = serializers.CharField(required=True, source='properties_title', max_length=255)
 
-    def get_fields(self):
-        fields = super().get_fields()
-        # This is a hack to allow fields with special characters
-        fields['eo:gsd'] = fields.pop('eo_gsd')
-        # logger.debug('Updated fields name: %s', fields)
-        return fields
```
ltshb, line 370: I'm glad that we can remove this hack :-)

ltshb (approved):
👍  Nice code has been simplified :-), just a comment to remove and another one to add.

app/stac_api/models.py
hansmannj: If I remember correctly, I implemented it during my work for 1410 and can revert deletion from the branch I pushed for that case.
I'll check that tomorrow.

rebert (approved):
works 4 me

app/stac_api/serializers.py
```diff
@@ -257,7 +257,7 @@ def get_crs(self, obj):
         return ["http://www.opengis.net/def/crs/OGC/1.3/CRS84"]
 
     def get_stac_extensions(self, obj):
-        return get_default_stac_extensions(True)
+        return get_default_stac_extensions()
```
boecklic, line 260: see comment above, just `return list()`

app/stac_api/models.py
```diff
@@ -36,22 +36,13 @@
...
 
 DEFAULT_SUMMARIES_VALUE = {"eo:gsd": [], "geoadmin:variant": [], "proj:epsg": []}
 
 
-def get_default_stac_extensions(is_collection=False):
-    if is_collection:
-        return list()
-    return list(DEFAULT_STAC_EXTENSIONS.values())
+def get_default_stac_extensions():
+    return list()
```
boecklic, line 45: this could maybe be removed as well (since afaik it's only called in one place). It's still reference in a migration, this would be a rare case where modifying old migrations is acceptable (replace the function name in the migrations with just `list`)

boecklic (commented):
Nice, simplification is always good though a bit unfortunate that we did all the work for nothing... made a little comment for one more possible simplification

app/stac_api/models.py
hansmannj: Yes, this could be removed. It only was introduced because once upon a time the stac_extensions were stored in an ArrayField I think. And hence it was required to have a callable to set the default values. But now implementation has changed.

I only left the whole construct with the two functions (one in the `models.py` and one in `serializers.py`) just for the case, that in future we might want to have non-empty stac_extensions again. But it is clearer to delete it for now and re-implement in case we need it in future.

boecklic: :+1:

app/stac_api/models.py
hansmannj: Done

app/stac_api/models.py
hansmannj: I'll do that in https://jira.swisstopo.ch/browse/BGDIINF_SB-1462

app/stac_api/serializers.py
hansmannj: done, just returning list() now

app/stac_api/models.py
hansmannj: done

PR #109, opened 2020-11-25
BGDIINF_SB-1429: Updated the collection temporal extent spec

ltshb:
Our current implementation doesn't supports open interval. Because this
is not required and not supported, simply remove it from spec.


rebert (approved):
Thanks

PR #111, opened 2020-11-25
Append trailing slash:

rebert:
- As the service gives an error when calling a url ending with a /

- This setting will append a slash to all urls:

  - collection becomes collection/

  - asset becomes asset/

  - etc.


ltshb:
Personally I would not mind having the endpoints working with AND without the trailing slash, but this needs to be checked with PO and Pasquale. From the SPEC the endpoints (except for the landing page) should be without trailing slash !

rebert:
Actually I'm not happy with the solution in this PR as it `redirects HTTP 301` .... Maybe `re_path with xxxxx/?$` would be a better solution.... or at least being consequent not using trialing slash at all: f.ex. `checker/ `

PR #112, opened 2020-11-25
BGDIINF_SB-1385: Corrected serializer validation & Implemented Collection de-serialization

ltshb:
In a model serializer when explicitely declaring the fields, we need to
add the validation manually and these validations should then match the
one from the models.
Note that when the fields are not manually declared but simply listed in
the Meta.fields then the validation is automatically generated with the
validator from the models.

Also finalize the colleciton de-serialization with unit tests


app/stac_api/serializers.py
```diff
@@ -238,19 +260,24 @@ class Meta:
...
-    crs = serializers.SerializerMethodField()
+    # NOTE: when explicitely declaring fields, we need to had the validation as for the field
+    # in model !
+    id = serializers.CharField(
+        required=True,
+        max_length=255,
+        source="name",
+        validators=[validate_name, UniqueValidator(queryset=Collection.objects.all())]
+    )
+    links = CollectionLinkSerializer(required=False, many=True)
```
hansmannj, line 271: according to specs, links are required (at least `href` and `rel` are required)

app/stac_api/serializers.py
```diff
@@ -416,6 +491,27 @@ class Meta:
...
         ]
 
+    # NOTE: when explicitely declaring fields, we need to had the validation as for the field
+    # in model !
+    type = serializers.CharField(source='media_type', max_length=200)
+    # Here we need to explicitely define these fields with the source, because they are renamed
+    # in the get_fields() method
+    eo_gsd = serializers.FloatField(source='eo_gsd', required=False, allow_null=True)
+    geoadmin_lang = serializers.ChoiceField(
+        source='geoadmin_lang', choices=['de', 'fr', 'it', 'rm', 'en']
```
hansmannj, line 501: probably an `allow_blank = True` is missing here, as in the Language class "NONE" (i.e. "") is a valid option.

app/tests/test_serializer.py
```diff
@@ -128,12 +129,191 @@ def test_collection_serialization(self):
...
         self.check_stac_collection(expected, python_native)
 
-        # back-transate to Python native:
-        stream = io.BytesIO(content)
-        data = JSONParser().parse(stream)
-        # back-translate into fully populated collection instance:
-        serializer = CollectionSerializer(data=data, context={'request': request})
-        self.assertEqual(True, serializer.is_valid(), msg='Serializer data not valid.')
+    def test_collection_deserialization_create_only_required(self):
+        data = OrderedDict([
```
hansmannj, line 133: according to specs `links` should be in the required minimum, too.

hansmannj (changes requested):
Nice solutions and good test coverage :smile: :+1:
Just some minor comments.

app/stac_api/serializers.py
```diff
@@ -213,7 +229,13 @@ class CollectionLinkSerializer(NonNullModelSerializer):
 
     class Meta:
         model = CollectionLink
-        fields = ['href', 'rel', 'link_type', 'title']
+        fields = ['href', 'rel', 'title', 'type']
+
+    # NOTE: when explicitely declaring fields, we need to had the validation as for the field
```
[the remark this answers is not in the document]
ltshb: Copy paste of typo

app/stac_api/serializers.py
ltshb: Yes but, links are automatically generated, here we provide the ability to add extra non-generated links but these ones are not required. I'll add a comment though.

app/tests/test_serializer.py
ltshb: Same as above links are generated (self, root, parent, ...)

rebert (approved):
Nice - lgtm

app/stac_api/serializers.py
ltshb: Good point, I'll add the `required=False` and `allow_blank=True` to match the SPEC and add a comment in https://jira.swisstopo.ch/browse/BGDIINF_SB-1372 which will fully implement the asset deserializer

hansmannj (approved):
:+1:

PR #117, opened 2020-11-26
BGDIINF_SB-1462: collection's update_summaries() on asset insert/update/delete

hansmannj:
**The branch name has 1461 instead of 1462, it has been too late at night when working on it obviously. Sorry for that! The commit messages were updated with the correct number, thanks Brice ;-) **

* implemented delete-case for assets in collection's update_summaries()
* added first unit tests for delete-case
* noticed while unit testing that update-case for assets was not complete yet --> implemented/fixed insert and update-cases as well
* exported collection's update_summaries() to separate file

### DIFF
--- a/app/stac_api/collection_summaries.py
+++ b/app/stac_api/collection_summaries.py
@@ -17,6 +17,16 @@ def float_in(flt, floats, **kwargs):
 
 
 def update_summaries_on_asset_delete(collection, asset):
+    '''
+    updates the collection's summaries on an asset's deletion or raises
+    errors when this fails.
+    :param collection: collection, for which the summaries probably need an update
+    :param asset: asset thats being deleted
+    For all the given parameters this function checks, if the corresponding
+    parameters of the collection need to be updated. If so, they will be either
+    updated or an error will be raised, if updating fails.
+    '''
+
     try:
         asset_qs = asset.__class__.objects.filter(item__collection_id=collection.pk
                                                  ).exclude(id=asset.id)
@@ -54,17 +64,33 @@ def update_summaries_on_asset_delete(collection, asset):
 
 
 def update_summaries_on_asset_insert(collection, asset):
+    '''
+    updates the collection's summaries on an asset's insertion or raises
+    errors when this fails.
+    :param collection: collection, for which the summaries probably need an update
+    :param asset: asset thats being inserted
+    For all the given parameters this function checks, if the corresponding
+    parameters of the collection need to be updated. If so, they will be either
+    updated or an error will be raised, if updating fails.
+    '''
+
     try:
         if asset.geoadmin_variant and \
             asset.geoadmin_variant not in collection.summaries["geoadmin:variant"]:
             collection.summaries["geoadmin:variant"].append(asset.geoadmin_variant)
+            collection.summaries["geoadmin:variant"].sort()
+            collection.save()
 
         if asset.proj_epsg and \
                 asset.proj_epsg not in collection.summaries["proj:epsg"]:
             collection.summaries["proj:epsg"].append(asset.proj_epsg)
+            collection.summaries["proj:epsg"].sort()
+            collection.save()
 
         if asset.eo_gsd and not float_in(asset.eo_gsd, collection.summaries["eo:gsd"]):
             collection.summaries["eo:gsd"].append(asset.eo_gsd)
+            collection.summaries["eo:gsd"].sort()
+            collection.save()
 
     except KeyError as err:
         logger.error(
@@ -76,6 +102,17 @@ def update_summaries_on_asset_insert(collection, asset):
 
 
 def update_summaries_on_asset_update(collection, asset, old_values):
+    '''
+    updates the collection's summaries on an asset's update or raises
+    errors when this fails.
+    :param collection: collection, for which the summaries probably need an update
+    :param asset: asset thats being updated
+    :param old_values: (optional) list with the original values of asset's
+    eo_gsd, geoadmin_variant and proj_epsg.
+    For all the given parameters this function checks, if the corresponding
+    parameters of the collection need to be updated. If so, they will be either
+    updated or an error will be raised, if updating fails.
+    '''
 
     original_eo_gsd = old_values[0]
     original_geoadmin_variant = old_values[1]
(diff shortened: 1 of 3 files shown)


app/stac_api/collection_summaries.py
```diff
@@ -0,0 +1,203 @@
...
+    original_proj_epsg = old_values[2]
+
+    asset_qs = None
+
+    try:
+
+        if original_geoadmin_variant != asset.geoadmin_variant:
+
+            if asset.geoadmin_variant and \
+            asset.geoadmin_variant not in collection.summaries["geoadmin:variant"]:
```
ltshb, line 128: Indentation issue. Might be a bug of `yapf` though. Anyway breaking long if statement can also be done with `()` instead of `/` different style both are acceptable. I've noticed that `yapf` tends to support `()` better than `\`:
e.g. : 
```
            if (asset.geoadmin_variant 
                and asset.geoadmin_variant not in collection.summaries["geoadmin:variant"]):
```

app/stac_api/collection_summaries.py
```diff
@@ -0,0 +1,203 @@
...
+    errors when this fails.
+    :param collection: collection, for which the summaries probably need an update
+    :param asset: asset thats being deleted
+    For all the given parameters this function checks, if the corresponding
+    parameters of the collection need to be updated. If so, they will be either
+    updated or an error will be raised, if updating fails.
+    '''
+
+    try:
+        asset_qs = asset.__class__.objects.filter(item__collection_id=collection.pk
```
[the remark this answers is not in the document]
ltshb: OK I've understood why the need of the `__class__`, it is to avoid circular dependencies. Therefore a short comment might be helpful

app/stac_api/collection_summaries.py
```diff
@@ -0,0 +1,203 @@
...
+            collection.summaries["eo:gsd"] = []
+            collection.save()
+
+    except KeyError as err:
+        logger.error(
+            "Error when updating collection's summaries values due to asset deletion: %s", err
+        )
+        raise ValidationError(_(
+            "Error when updating collection's summaries values due to asset deletion."
+        ))
```
ltshb, line 63: Not sure if this is really needed ? We should always have the `geoadmin:variant`, `proj:espg` and `eo:gsd` keys. Did you have a use case where they were missing ? In any case we should not have a ValidationError here, this error is for user input validation.
From my understanding the only use case that could happen, is during initialization where the summaries haven't been set with default values, but that should not be an error, simply a debug log without raising exception.

app/stac_api/collection_summaries.py
```diff
@@ -0,0 +1,203 @@
...
+    '''
+
+    try:
+        asset_qs = asset.__class__.objects.filter(item__collection_id=collection.pk
+                                                 ).exclude(id=asset.id)
+        if bool(asset_qs):
+
+            if not asset_qs.filter(geoadmin_variant=asset.geoadmin_variant).exists():
+                collection.summaries["geoadmin:variant"].remove(asset.geoadmin_variant)
+                collection.summaries["geoadmin:variant"].sort()
```
boecklic, line 37: out of curiosity: why is the `.sort()` necessary here?

app/stac_api/collection_summaries.py
ltshb: Is the exception happening with the admin page, when creating new Asset or new Collection ?

ltshb (commented):
Very nice. Just a comment about error handling.

ltshb:
The commit message can still be quickly changed with `git rebase -i` ;-)

app/stac_api/collection_summaries.py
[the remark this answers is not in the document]
boecklic: you can use `type(asset)` to the get the model class, so `type(asset).objects.filter()` does work in this case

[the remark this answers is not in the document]
ltshb: I like the solution with `type(asset)`

app/stac_api/collection_summaries.py
hansmannj: For example it makes unit testing simpler. With the `.sort()` you know exactly in which order the variants are to be expected. But we can also drop it, I could live without it :-)

app/stac_api/collection_summaries.py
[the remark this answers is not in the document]
hansmannj: Yes, it was to avoid circular dependencies.
I'll implement the `type(asset)` variant, it is somehow easier to read, I guess.

app/stac_api/collection_summaries.py
hansmannj: Ah, you are right, in the delete-case there's not really a need for `.sort()` after deleting a variant. Make only sense in case of additions

ltshb: I like the sort for unit test and also for human reading read data on the API. From my understanding they are only a few different `geoadmin:variant` ? so the sort() should not have an impact on performance, which could be an issue if we would have several thousands of variant.

boecklic: OK. I'm still not used to relying on `dict`s being ordered

hansmannj: I deleted the `.sort()` in this place, as it is not needed when a variant is removed. But I left it in the other cases, if that's fine for everyone.

app/stac_api/collection_summaries.py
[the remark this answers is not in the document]
hansmannj: replaced `asset.__class__` with `type(asset)`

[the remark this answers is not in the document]
hansmannj: replayed `asset_qs` with `assets` everywhere

app/stac_api/collection_summaries.py
hansmannj: now using `()` instead of `\`

app/stac_api/collection_summaries.py
hansmannj: No, I did not have this error somewhere. Just implemented the exception handling here purely for precaution.
I'll remove it and replace with a debug message, just in case

ltshb (commented):
See question about the exception handling, I would try without it, test then creating a collection, item, asset through admin page and with the populate_testdb

app/stac_api/collection_summaries.py
```diff
@@ -0,0 +1,200 @@
...
+            # asset was the last item in the collection
+            collection.summaries["geoadmin:variant"] = []
+            collection.summaries["proj:epsg"] = []
+            collection.summaries["eo:gsd"] = []
+            collection.save()
+
+    except KeyError as err:
+        logger.debug(
+            "Error when updating collection's summaries values due to asset deletion: %s", err
+        )
```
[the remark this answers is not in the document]
hansmannj: You are right, sorry, I did not realize at first, that you were referring to all `try ... except` in the file. Did not experience any of these errors so far, purely precaution. But I deleted them now. We could still add them again later, if needed.

ltshb (approved):
:+1:

PR #118, opened 2020-11-30
BGDIINF_SB-1454: Remove assetQuery from doc

rebert:
As this query is not implemented yet, it has been removed from the spec.

Removed means in this case, that it has been hidden as a comment #

### DIFF
--- a/spec/static/openapitransactional.yaml
+++ b/spec/static/openapitransactional.yaml
@@ -28,15 +28,6 @@ tags:
   name: STAC
 components:
   parameters:
-    assetQuery:
-      description: >-
-        Query for properties in assets (e.g. mediatype). Use the JSON form of the
-        assetQueryFilter used in POST.
-      in: query
-      name: assetQuery
-      required: false
-      schema:
-        type: string
     bbox:
       description: >-
         Only features that have a geometry that intersects the bounding box are selected.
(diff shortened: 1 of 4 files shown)


spec/static/openapitransactional.yaml
```diff
@@ -2379,8 +2300,8 @@ paths:
         in the link.
       operationId: getSearchSTAC
       parameters:
-      - $ref: "#/components/parameters/assetQuery"
-      - $ref: "#/components/parameters/query"
+      - #- $ref: "#/components/parameters/assetQuery"
```
hansmannj, line 2303: why no "-" here anymore?

hansmannj (approved):
LGTM :+1:
Just one minor comment about a missing "-".

spec/static/openapitransactional.yaml
hansmannj: Referring to line # 2304, sorry

spec/static/openapitransactional.yaml
```diff
@@ -2379,8 +2300,8 @@ paths:
...
+        $ref: "#/components/parameters/query"
```
hansmannj, line 2304: why no "-" here anymore?

spec/static/openapitransactional.yaml
rebert: this has been auto-generated ... I guess, otherwise I would have to remove it completely from the source (instead of hiding it with a comment)

spec/static/openapitransactional.yaml
rebert: the - comes from above ... as autogenerated

ltshb (approved):

PR #120, opened 2020-11-30
BGDIINF_SB-1372: Asset deserializer

hansmannj:
Implementation of asset deserialization and an according unit test


hansmannj:
Waiting on reply of @p1d1d1 on required asset properties. As soon I know the currently complete list of required properties, I'll adapt the asset's admin page, if necessary

app/stac_api/serializers.py
```diff
@@ -514,23 +515,37 @@ class Meta:
...
+    title = serializers.CharField(max_length=255, required=False, allow_blank=True)
     type = serializers.CharField(source='media_type', max_length=200)
     # Here we need to explicitely define these fields with the source, because they are renamed
     # in the get_fields() method
+    description = serializers.CharField(max_length=255, required=False, allow_blank=True)
     eo_gsd = serializers.FloatField(source='eo_gsd', required=False, allow_null=True)
     geoadmin_lang = serializers.ChoiceField(
         source='geoadmin_lang',
-        choices=['de', 'fr', 'it', 'rm', 'en'],
+        choices=['de', 'fr', 'it', 'rm', 'en', ''],
```
rebert, line 535: why an empty string here?

spec/paths.yml
```diff
@@ -124,7 +124,7 @@ paths:
         may not be set explicitly in the link.
       operationId: getSearchSTAC
       parameters:
-        - $ref: "#/components/parameters/assetQuery"
+        # - $ref: "#/components/parameters/assetQuery"
```
rebert, line 127: thanks :-)

app/stac_api/serializers.py
```diff
@@ -514,23 +515,37 @@ class Meta:
...
+        source='name',
+        required=True,
+        max_length=255,
+        validators=[validate_name, UniqueValidator(queryset=Asset.objects.all())]
+    )
+    title = serializers.CharField(max_length=255, required=False, allow_blank=True)
     type = serializers.CharField(source='media_type', max_length=200)
     # Here we need to explicitely define these fields with the source, because they are renamed
     # in the get_fields() method
+    description = serializers.CharField(max_length=255, required=False, allow_blank=True)
```
ltshb, line 531: I would not add the `description` explicitly here, it is not needed. Note that when adding explicitly fields we need to make sure that the validation is done as in the model ! See comments.

app/stac_api/serializers.py
```diff
@@ -514,23 +515,37 @@ class Meta:
...
+    item = serializers.SlugRelatedField(
+        slug_field='name', write_only=True, queryset=Item.objects.all()
+    )
+    id = serializers.CharField(
+        source='name',
+        required=True,
+        max_length=255,
+        validators=[validate_name, UniqueValidator(queryset=Asset.objects.all())]
+    )
+    title = serializers.CharField(max_length=255, required=False, allow_blank=True)
```
ltshb, line 527: I would not add the `title` explicitly here, it is not needed. Note that when adding explicitly fields we need to make sure that the validation is done as in the model ! See comments.

app/stac_api/serializers.py
```diff
@@ -546,6 +561,14 @@ def get_fields(self):
...
         return fields
 
+    def create(self, validated_data):
+        asset = Asset.objects.create(**validated_data)
+        return asset
+
+    def update(self, instance, validated_data):
+        instance.save()
+        return instance
+
```
ltshb, line 571: Are these really needed ? Does the default implementation not sufficient ? When the default implementation works I would rely on it.

app/tests/test_serializer.py
```diff
@@ -800,6 +800,9 @@ def test_asset_serialization(self):
         # translate to Python native:
         serializer = AssetSerializer(self.asset, context={'request': request})
         python_native = serializer.data
+        # hack to deal with the item property, as it is "write_only", it will not appear
+        # in the mocked request's data. So we manually add it here:
+        python_native["item"] = item_name
```
ltshb, line 805: I would move this to the `python_native_back` below because this hack is only for de-serialization needed and not for serialization.

ltshb (changes requested):
See comments in code. I would also suggest to add more unittests e.g.
- test deserialization with invalid data
- test deserialization with missing required fields
-  test deserialization with only required fields

app/stac_api/serializers.py
boecklic: it's a valid choice. But maybe use `Language.NONE` instead (in the rare case this value will change)

boecklic (approved):
nice, made one little comment

app/stac_api/serializers.py
hansmannj: Good point, how about `choices = Asset.Language.values` here? This way we'd always be consistent with the definition in the models. For the "empty language" value as well as all the other languages?

boecklic: even better ;-)

app/stac_api/serializers.py
hansmannj: I made them explicit because I assumed the default value for `required` would be `True` and we want `False` here.
But I did not test, if they default to `False` automatically, as the model defines `blank` or `null` as `True`. 
As discussed, I'll add unit test first, and once they pass successfully, I'll start deleting the (probably unnecessary) explicit definitions and check, if the implicit definition is working.
Implicit would be preferred, if working.

Same for all the other explicit fields below

spec/paths.yml
hansmannj: you're welcome :smile:

app/tests/test_serializer.py
hansmannj: Makes sense, done

app/stac_api/serializers.py
hansmannj: unit tests succeed with the default implicit definition, so I deleted the explicit definitions for `description` and `title`.

app/stac_api/serializers.py
hansmannj: see above

app/stac_api/serializers.py
hansmannj: unit test succeed with the default implementation. So I deleted the explicit definitions here

hansmannj:
Thanks all for your good comments.
They should all be considered/implemented now.

app/stac_api/serializers.py
```diff
@@ -522,23 +523,34 @@ class Meta:
...
+        validators=[validate_name, UniqueValidator(queryset=Asset.objects.all())]
+    )
     type = serializers.CharField(source='media_type', max_length=200)
     # Here we need to explicitely define these fields with the source, because they are renamed
     # in the get_fields() method
     eo_gsd = serializers.FloatField(source='eo_gsd', required=False, allow_null=True)
     geoadmin_lang = serializers.ChoiceField(
         source='geoadmin_lang',
-        choices=['de', 'fr', 'it', 'rm', 'en'],
+        choices=Asset.Language.values,
```
ltshb, line 540: I like that

ltshb (approved):

PR #121, opened 2020-11-30
BGDIINF_SB-1250: Implement conformance endpoint

rebert:
The conformance endpoint is being implemented to be modified

with the admin interface.


app/stac_api/admin.py
```diff
@@ -27,6 +28,12 @@ class LandingPageAdmin(SingletonModelAdmin):
     inlines = [LandingPageLinkInline]
 
 
+@admin.register(ConformancePage)
+class ConformancePageAdmin(SingletonModelAdmin):
+    model = ConformancePage
+    extra = 0
```
ltshb, line 34: I think these are not needed ? Could be simply
```
@admin.register(ConformancePage)
class ConformancePageAdmin(SingletonModelAdmin):
    pass
```
The model is redundant because already in register decorator, but this needs to be quickly tested.

app/stac_api/models.py
```diff
@@ -92,6 +92,20 @@ def validate_geometry(geometry):
     return geometry
 
 
+def set_default_links():
```
ltshb, line 95: Function name is too general, should be `get_conformance_default_links()`

app/stac_api/models.py
```diff
@@ -132,6 +146,22 @@ class Meta:
         unique_together = (('rel', 'landing_page'))
 
 
+class ConformancePage(SingletonModel):
+    conforms_to = ArrayField(
+        models.URLField(
+            blank=True,
+            null=True
```
ltshb, line 153: Empty string should not be accepted. However empty array could be accepted here but don't have to.

app/stac_api/serializers.py
```diff
@@ -153,6 +154,18 @@ class Meta:
...
+class ConformancePageSerializer(serializers.ModelSerializer):
+
+    class Meta:
+        model = ConformancePage
+        fields = ['conforms_to']
+
+    def to_representation(self, instance):
+        representation = super().to_representation(instance)
+        representation['conformsTo'] = representation.pop('conforms_to')  # renaming key
+        return representation
```
ltshb, line 166: This is not needed, simply named the field `conformsTo` in model.

ltshb (changes requested):
:+1:  just small changes required.

ltshb (commented):
Good. I just tried it out and I suggest to add the conformsTo field in the admin page as text area e.g.
```
@admin.register(ConformancePage)
class ConformancePageAdmin(SingletonModelAdmin):
    formfield_overrides = {
        ArrayField: {
            'widget': Textarea(attrs={
                'rows': 10, 'cols': 60
            })
        },
    }
```
This way it is easier to add/modify urls.

ltshb (approved):
:1st_place_medal:

hansmannj (approved):
Nice :smile:

PR #124, opened 2020-12-02
BGDIINF_SB-1497 adding version information

boecklic:
This PR is implementing versioning guideline as described in https://github.com/geoadmin/doc-guidelines/pull/65 for service_stac.

It uses a github action (executed on push to `develop` and `master` only) to automatically tag pushes to `develop` with `vMAJOR.MINOR.PATCH-beta.Y` and pushes to `master` with `vMAJOR.MINOR.PATH`. These tags are used during the build step to overwrite a `version.py` file which itself is used in templates to display the currently deployed version in the admin pages and API landing page.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -4,7 +4,6 @@ on:
     branches:
       - master
       - develop
-      - feature_BGDIINF_SB-1497_versioning
 jobs:
   build:
     runs-on: ubuntu-latest
@@ -12,6 +11,9 @@ jobs:
     - uses: actions/checkout@master
       with:
         fetch-depth: '10'
+    # Details and Docu
+    # https://github.com/anothrNick/github-tag-action
+    # https://github.com/geoadmin/doc-guidelines/blob/master/GIT_FLOW.md#versioning
     - name: Bump version and push tag
       uses: anothrNick/github-tag-action@1.33.0
       env:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,21 @@
+name: Bump version
+on:
+  push:
+    branches:
+      - master
+      - develop
+      - feature_BGDIINF_SB-1497_versioning
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+      with:
+        fetch-depth: '10'
+    - name: Bump version and push tag
+      uses: anothrNick/github-tag-action@1.33.0
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        WITH_V: true
+        RELEASE_BRANCHES: master
+        TAG_CONTEXT: repo
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,7 @@ SUMMON ?= summon --up -p gopass -e service-stac-$(ENV)
 GIT_HASH := `git rev-parse HEAD`
 GIT_BRANCH := `git symbolic-ref HEAD --short 2>/dev/null`
 GIT_DIRTY := `git status --porcelain`
-GIT_TAG := `git describe --tags 2>/dev/null`
+GIT_TAG := `git describe --tags || echo "no version info"`
 AUTHOR := $(USER)
 
 all: help
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ DOCKER_BUILD_TIMESTAMP = $(TIMESTAMPS)/.docker-test.timestamp
 
 # Docker variables
 DOCKER_IMG_LOCAL_TAG = swisstopo/$(SERVICE_NAME):$(USER).latest
-DOCKER_IMG_LOCAL_TAG_TEST = swisstopo/$(SERVICE_NAME):$(USER).latest-dev
+DOCKER_IMG_LOCAL_TAG_DEV = swisstopo/$(SERVICE_NAME):$(USER).latest-dev
 
 # Find all python files that are not inside a hidden directory (directory starting with .)
 PYTHON_FILES := $(shell find $(APP_SRC_DIR) -type f -name "*.py" -print)
@@ -47,6 +47,7 @@ SUMMON ?= summon --up -p gopass -e service-stac-$(ENV)
 GIT_HASH := `git rev-parse HEAD`
 GIT_BRANCH := `git symbolic-ref HEAD --short 2>/dev/null`
 GIT_DIRTY := `git status --porcelain`
+GIT_TAG := `git describe --tags 2>/dev/null`
 AUTHOR := $(USER)
 
 all: help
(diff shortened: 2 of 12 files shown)


Makefile
```diff
@@ -47,6 +47,7 @@ SUMMON ?= summon --up -p gopass -e service-stac-$(ENV)
 GIT_HASH := `git rev-parse HEAD`
 GIT_BRANCH := `git symbolic-ref HEAD --short 2>/dev/null`
 GIT_DIRTY := `git status --porcelain`
+GIT_TAG := `git describe --tags 2>/dev/null || echo 'unknown'`
```
ltshb, line 50: I would only use annotated tag ! I would also avoid writing stderr to null, if an error occur we should have a dirty tag and not a correct  one. Moreover the `git describe` command adds info to the tag if the current commit is not on the tag see https://git-scm.com/docs/git-describe. We should have a bullet proof method of getting the tag to avoid making image with wrong tag.

app/config/settings_prod.py
```diff
@@ -24,6 +24,12 @@
 # the setup
 APP_ENV = os.environ.get('APP_ENV', 'local')
 
+# Determine the APP version
+try:
+    from .version import APP_VERSION  # pylint: disable=unused-import
+except ModuleNotFoundError:
+    APP_VERSION = 'No version info'
```
ltshb, line 31: Maybe add a comment why this could happen and we are ok with it.

buildspec.yml
```diff
@@ -33,19 +33,30 @@ phases:
           export GITHUB_BRANCH=${GITHUB_BRANCH#remotes/origin/};
         fi
       - export COMMIT_HASH=$(echo $CODEBUILD_RESOLVED_SOURCE_VERSION | cut -c 1-7)
-      - export IMAGE_TAG="${GITHUB_BRANCH}.${COMMIT_HASH}"
-      - echo "GITHUB_BRANCH=${GITHUB_BRANCH} COMMIT_HASH=${COMMIT_HASH} IMAGE_TAG=${IMAGE_TAG}"
+      - GIT_TAG="$(git describe --tags 2>/dev/null)"
```
ltshb, line 36: Same comment as in makefile

.github/workflows/main.yml
```diff
@@ -0,0 +1,20 @@
...
+    - uses: actions/checkout@master
+      with:
+        fetch-depth: '10'
+    - name: Bump version and push tag
+      uses: anothrNick/github-tag-action@1.33.0
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        WITH_V: true
+        RELEASE_BRANCHES: master
+        TAG_CONTEXT: repo
```
ltshb, line 20: Maybe a comment would be good here. What does it do exactly ? Does this generate also the version.py, commit it, push it and tag it ?

ltshb (commented):
Why using version with 'v0.0.0' with `v` prefix ? 
I'm a bit concerned with the `git describe --tags` command, see comment.

Makefile
boecklic: What's your reasoning for "only using annotated tag"?

boecklic: And in what way are annotated tags different to leightweight tags when triggering build based on push to branch and not based on creation of tags?

app/config/settings_prod.py
boecklic: It actually shouldn't happen, I changed it

Makefile
boecklic: and btw: the `2>/dev/null` stems from two lines above... plus if you're in local development I think it's ok not to have a tag yet

ltshb: From https://git-scm.com/docs/git-tag 

> Annotated tags are meant for release while lightweight tags are meant for private or temporary object labels. For this
reason, some git commands for naming objects (like git describe) will ignore lightweight tags by default.

That's mainly why I would only used annotated tag. Also annotated tags contains meta data like message, creation date, author while lightweight tag not.

ltshb: > and btw: the `2>/dev/null` stems from two lines above... plus if you're in local development I think it's ok not to have a tag yet

I agree that for local development it is ok not to have a tag, but `git describe` will return a previous tag if the current commit is not tagged ! We should not mark an image with a wrong tag.

ltshb: OK I've just tried the `git describe` command from your branch (which contains already a tag) and it gave `v0.2.0-beta.1-2-g17f5fcc` which is fine for a dirty tag. So I suggest simply to remove the `--tags` option and to use annotated tag

boecklic: Interestingly, github doesn't treat leightweight tags as being very 'private' or 'temporary', they show up in tags list as any other tag (in fact when you create a tag via UI it's a leightweight tag). And for `develop` one could even argue that a leightweight tag is the right thing, since it's more of 'private' and 'temporary' character than a release.

buildspec.yml
boecklic: changed it

app/config/version.py
```diff
@@ -0,0 +1,21 @@
...
+
+# Note that this file is overwritten during the build
+# process by either the git tag of the commit
+# (see Dockerfile for details)
+
+proc = subprocess.Popen(["git", "describe", "--tags"],
+                        stdout=subprocess.PIPE,
+                        stderr=subprocess.PIPE)
+stdout, stderr = proc.communicate()
+git_version = stdout.decode('utf-8').strip()  # pylint: disable=invalid-name
```
ltshb, line 12: Disabling `invalid-name` is really required here ? which name is triggering this pylint issue ? Maybe we could add this name in the pylint good name list, for me all names in this line are really valid.

app/config/version.py
```diff
@@ -0,0 +1,21 @@
...
+if git_version == '':
+    # If theres no git tag found in the history
+    proc = subprocess.Popen(["git", "rev-parse", "--short", "HEAD"],
+                            stdout=subprocess.PIPE,
+                            stderr=subprocess.PIPE)
+    stdout, stderr = proc.communicate()
+    APP_VERSION = f"v_{stdout.decode('utf-8').strip()}"
+else:
+    APP_VERSION = git_version
```
ltshb, line 21: I would comment both case with the output of the git command that we are expecting.
`git describe --tags` => if current commit is tagged then return tag, else return last tag found suffixed with the short hash (e.g. v0.2.0-beta.1-2-g17f5fcc)
When no tag is found uses the current git commit short hash (v_78fec8d)
This way it clearer what version we expect without knowing perfectly these git commands that most people don't know and don't use

app/config/version.py
boecklic: I didn't understand this either, for some reason it expected an UPPERCASE var name...

ltshb: Ok I understand now, missed that. The python guideline expect to have global constant as UPPERCASE :-) and this case it is a constant, so maybe change it to uppercase this way your 100% compliant with our guideline ;-)

app/config/version.py
boecklic: I added some comments.

app/config/version.py
boecklic: Ok

ltshb (approved):

.github/workflows/main.yml
```diff
@@ -0,0 +1,23 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+      with:
+        fetch-depth: '10'
+    # Details and Docu
+    # https://github.com/anothrNick/github-tag-action
+    # https://github.com/geoadmin/doc-guidelines/blob/master/GIT_FLOW.md#versioning
```
hansmannj, line 16: :+1:

PR #125, opened 2020-12-02
BGDIINF_SB-1477: Added admin page unittests 

ltshb:
Due to the BGDIINF_SB-1500 bug the admin page collection removal with
items test case doesn't work, therefore it is commented out until
BGDIINF_SB-1500 is fixed.


boecklic (approved):
Nice!

PR #127, opened 2020-12-02
BGDIINF_SB-1500: Fixed collection delete stackoverflow 

ltshb:
When deleting a collection containing items, django crashed with a
stackoverflow. This was due to the code that saved the original fields
value in the Item.__init__() method. Why this caused a stackoverflow it
is still a mistery to me.

But instead of using this hack I changed the algorithm to use a django
way of keeping track of original as described in https://docs.djangoproject.com/en/3.1/ref/models/instances/#customizing-model-loading
and this solved the issue.


rebert (approved):
thanks

hansmannj:
very nice, thanks :smile:

boecklic (changes requested):
Nice solution! I would maybe also adapt the 'keep-original' logic in `Asset` right away, we'll run into similar troubles there. I think this is a very important issue: we need to be careful with class-variables and instance-variables. Looking at the models it's tempting to just add other class variables. Don't forget that Django does a lot of magic behind the scenes to manage class-vs-instance variables. Unless you're sure that it really MUST be a class variable, use instance variables, it's most likely the right choice.

ltshb:
> Nice solution! I would maybe also adapt the 'keep-original' logic in `Asset` right away, we'll run into similar troubles there. I think this is a very important issue: we need to be careful with class-variables and instance-variables. Looking at the models it's tempting to just add other class variables. Don't forget that Django does a lot of magic behind the scenes to manage class-vs-instance variables. Unless you're sure that it really MUST be a class variable, use instance variables, it's most likely the right choice.

Good point I missed the Asset. I will also change the class var to instance.

boecklic (approved):
:1st_place_medal:

PR #128, opened 2020-12-02
BGDIINF_SB-1503 changing setting for SECURE_PROXY_SSL_HEADER

boecklic:
Django doesn't correctly determine if the original request was made using SSL with our Stack. Therefore we need to adapt the settings value SECURE_PROXY_SSL_HEADER

### DIFF
--- a/app/config/logging-cfg-local.yml
+++ b/app/config/logging-cfg-local.yml
@@ -20,10 +20,6 @@ loggers:
     level: DEBUG
     handlers:
       - console
-  default:
-    level: DEBUG
-    handlers:
-      - console
 
 filters:
   application:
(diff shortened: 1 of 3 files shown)


app/config/logging-cfg-local.yml
```diff
@@ -20,10 +20,6 @@ loggers:
     level: DEBUG
     handlers:
       - console
-  default:
-    level: DEBUG
-    handlers:
-      - console
```
ltshb, line 26: Just for curiosity why did you remove this ?

ltshb (approved):
:+1: 
Just a general comment about git commit message, as good practice I would suggest to squash both commit and add a proper commit message that explain why these changes were needed. This way when looking at the repo we don't have to search in the PR to understand these changes. So also add the jira ticket in commit. This helps a lot when doing git blame on the code to know why lline of code is as it is.

app/config/logging-cfg-local.yml
boecklic: I added this some time ago but I think it's not used (since we use `.getLogger(__name__)` everywhere and not a common default logger)

PR #131, opened 2020-12-03
BGDIINF_SB-1491: authentication

hansmannj:
* added token, basic and session authentication
* set `DjangoModelPermissionsOrAnonReadOnly` as default (non-logged in, anonymous users have read-only access, logged-in users can take actions according to their individual permissions)
* added POST endpoint for users to request a token based on their username and password
(e.g. as follows: `curl -d "username=testuser&password=test12345" http://127.0.0.1:8000/api-token-auth/`)
* added unit tests for authorized POSTs as well as unauthorized POSTs or requets with wrong credentials

The POST/PUT endpoints for collections as well as a lot of the unit tests and test data for the collections_endpoint were borrowed from https://github.com/geoadmin/service-stac/pull/132, thanks @rebert :smiley:

### DIFF
--- a/app/config/settings_prod.py
+++ b/app/config/settings_prod.py
@@ -80,7 +80,7 @@ MIDDLEWARE = [
 
 ROOT_URLCONF = 'config.urls'
 API_BASE = 'api/stac/v0.9'
-LOGIN_URL = f"/{API_BASE}/api-auth/login/"
+LOGIN_URL = f"/{API_BASE}/login/"
 
 TEMPLATES = [
     {
--- a/app/config/settings_prod.py
+++ b/app/config/settings_prod.py
@@ -80,7 +80,7 @@ MIDDLEWARE = [
 
 ROOT_URLCONF = 'config.urls'
 API_BASE = 'api/stac/v0.9'
-LOGIN_URL = f"/{API_BASE}/login/"
+LOGIN_URL = "/api/stac/admin/login/"
 
 TEMPLATES = [
     {
--- a/app/config/settings_prod.py
+++ b/app/config/settings_prod.py
@@ -51,6 +51,9 @@ ALLOWED_HOSTS += os.getenv('ALLOWED_HOSTS', '').split(',')
 INSTALLED_APPS = [
     'rest_framework',
     'rest_framework_gis',
+    'rest_framework.authtoken',
+    #  Note: If you use TokenAuthentication in production you must ensure
+    #  that your API is only available over https.
     'stac_api.apps.StacApiConfig',
     'config.apps.StacAdminConfig',
     'django.contrib.auth',
@@ -77,6 +80,7 @@ MIDDLEWARE = [
 
 ROOT_URLCONF = 'config.urls'
 API_BASE = 'api/stac/v0.9'
+LOGIN_URL = f"/{API_BASE}/api-auth/login/"
 
 TEMPLATES = [
     {
@@ -181,11 +185,21 @@ else:
 TEST_RUNNER = 'tests.runner.TestRunner'
 
 # set default pagination configuration
+# set authentication schemes
+
 REST_FRAMEWORK = {
+    'DEFAULT_AUTHENTICATION_CLASSES': [
+        'rest_framework.authentication.BasicAuthentication',
+        'rest_framework.authentication.TokenAuthentication',
+        'rest_framework.authentication.SessionAuthentication',
+    ],
     'DEFAULT_PAGINATION_CLASS': 'stac_api.apps.CursorPagination',
     'PAGE_SIZE': 100,
     'PAGE_SIZE_LIMIT': 100,
-    'EXCEPTION_HANDLER': 'stac_api.apps.custom_exception_handler'
+    'EXCEPTION_HANDLER': 'stac_api.apps.custom_exception_handler',
+    'DEFAULT_PERMISSION_CLASSES': [
+        'rest_framework.permissions.DjangoModelPermissionsOrAnonReadOnly',
+    ]
 }
 
 # Exception handling
--- a/app/stac_api/urls.py
+++ b/app/stac_api/urls.py
@@ -1,5 +1,8 @@
 from django.conf import settings
+from django.urls import include
 from django.urls import path
+
+from rest_framework.authtoken.views import obtain_auth_token
 
 from stac_api.views import AssetDetail
 from stac_api.views import AssetsList
@@ -40,4 +43,7 @@ urlpatterns = [
         AssetDetail.as_view(),
         name='asset-detail'
     ),
+    # https://www.django-rest-framework.org/tutorial/quickstart/#urls
+    path(f"{API_BASE}/api-auth/", include('rest_framework.urls', namespace='rest_framework')),
+    path('api-token-auth/', obtain_auth_token, name='api-token-auth'),
 ]
--- a/app/stac_api/urls.py
+++ b/app/stac_api/urls.py
@@ -1,5 +1,4 @@
 from django.conf import settings
-from django.urls import include
 from django.urls import path
 
 from rest_framework.authtoken.views import obtain_auth_token
@@ -43,7 +42,5 @@ urlpatterns = [
         AssetDetail.as_view(),
         name='asset-detail'
     ),
-    # https://www.django-rest-framework.org/tutorial/quickstart/#urls
-    path(f"{API_BASE}/", include('rest_framework.urls', namespace='rest_framework')),
     path(f"{API_BASE}/get_token", obtain_auth_token, name='get_token'),
 ]
--- a/app/stac_api/urls.py
+++ b/app/stac_api/urls.py
@@ -44,6 +44,6 @@ urlpatterns = [
         name='asset-detail'
     ),
     # https://www.django-rest-framework.org/tutorial/quickstart/#urls
-    path(f"{API_BASE}/api-auth/", include('rest_framework.urls', namespace='rest_framework')),
-    path('api-token-auth/', obtain_auth_token, name='api-token-auth'),
+    path(f"{API_BASE}/", include('rest_framework.urls', namespace='rest_framework')),
+    path(F"{API_BASE}/get_token/", obtain_auth_token, name='api-token-auth'),
 ]
--- a/app/stac_api/urls.py
+++ b/app/stac_api/urls.py
@@ -45,5 +45,5 @@ urlpatterns = [
     ),
     # https://www.django-rest-framework.org/tutorial/quickstart/#urls
     path(f"{API_BASE}/", include('rest_framework.urls', namespace='rest_framework')),
-    path(F"{API_BASE}/get_token/", obtain_auth_token, name='api-token-auth'),
+    path(f"{API_BASE}/get_token", obtain_auth_token, name='get_token'),
 ]
--- a/app/stac_api/urls.py
+++ b/app/stac_api/urls.py
@@ -42,5 +42,5 @@ urlpatterns = [
         AssetDetail.as_view(),
         name='asset-detail'
     ),
-    path(f"{API_BASE}/get_token", obtain_auth_token, name='get_token'),
+    path(f"{API_BASE}/get-token", obtain_auth_token, name='get-token'),
 ]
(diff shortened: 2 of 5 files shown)


hansmannj:
This is WIP and for now I test authentication for the CollectionListView, which does not make sense. But there are no write endpoints so far to test, yet.

rebert:
@hansmannj for the write endpoint you could use my branch: https://github.com/geoadmin/service-stac/pull/132

app/tests/test_collections_endpoint.py
```diff
@@ -84,3 +118,164 @@ def test_collections_limit_query(self):
...
         self.assertEqual(400, response.status_code)
+
+    def test_valid_collections_post(self):
+        payload_json = self.sample_collections['valid_collection_set_1']
+        grant_permissions('Collection', self.user)
+        self.client.login(username=self.username, password=self.password)
+
+        response = self.client.post(
+            f"/{API_BASE}/collections",
+            data=json.dumps(payload_json),
```
hansmannj, line 129: For some reason I had to use `json.dumps()` here, although the same test seems to work for @rebert in https://github.com/geoadmin/service-stac/pull/132 without it.
When I omit `json.dumps()` here, there would be an issue with single instead of double quotes for the json-payload..

app/tests/test_collections_endpoint.py
```diff
@@ -1,31 +1,65 @@
...
     def setUp(self):
-        self.client = Client()
+        self.client = APIClient()
         self.factory = APIRequestFactory()
         self.collections, self.items, self.assets = db.create_dummy_db_content(4, 4, 4)
         self.maxDiff = None  # pylint: disable=invalid-name
+        self.sample_collections = get_sample_data('collections')
+        self.username = 'SherlockHolmes'
+        self.password = '221B_BakerStreet'
+        self.user = get_user_model().objects.create_user(
```
boecklic, line 59: Hmm, if we say the we currently distinguish two types of users, superuser (can do everything, accessing admin, issuing POST/PUT requests etc.) and 'normal'/anonymous users that can just do GET requests I'd probably create a superuser here and omit the explicit permission granting.
Otherwise it will be difficult to track which permissions are exactly required over all and we should rather consider to create an 'apiuser' group or sth like this where we explicitly define the necessary permissions

boecklic (commented):
Looks good. Made a comment regarding permissions...

One general question: is this PR based on https://github.com/geoadmin/service-stac/pull/132 so that https://github.com/geoadmin/service-stac/pull/132 can be merged individually (before this one) or does this PR include the other one and https://github.com/geoadmin/service-stac/pull/132 can be closed?

app/tests/test_collections_endpoint.py
hansmannj: Makes sense. We can still have some "a bit less than superuser"-users later, when needed.

hansmannj:
> Looks good. Made a comment regarding permissions...
> 
> One general question: is this PR based on #132 so that #132 can be merged individually (before this one) or does this PR include the other one and #132 can be closed?

This one was rebased on #132 at a point. But since then, I guess, the common content of the two PRs diverged a bit and will give us a joyfull time solving merge conflicts :smile:

boecklic:
> > Looks good. Made a comment regarding permissions...
> > One general question: is this PR based on #132 so that #132 can be merged individually (before this one) or does this PR include the other one and #132 can be closed?
> 
> This one was rebased on #132 at a point. But since then, I guess, the common content of the two PRs diverged a bit and will give us a joyfull time solving merge conflicts smile
> 
> I guess merging #132 before this one makes sense, then I'll care about the (hopefully few) merge conflicts later on.

:see_no_evil: :smile:

app/stac_api/urls.py
```diff
@@ -40,4 +43,7 @@
         AssetDetail.as_view(),
         name='asset-detail'
     ),
+    # https://www.django-rest-framework.org/tutorial/quickstart/#urls
+    path(f"{API_BASE}/api-auth/", include('rest_framework.urls', namespace='rest_framework')),
+    path('api-token-auth/', obtain_auth_token, name='api-token-auth'),
```
hansmannj, line 48: Perhaps it might make sense to have the rest_framework.urls directly at `API_BASE` and drop the `api-auth` part. The login URL would then be `API_BASE/login/`.
And perhaps the obtain_auth_token endpoint could also be under `API_BASE/api-token-auth/` or `API_BASE/obtain-token/`.
What do you think?

boecklic: For the `api-toke-auth` I think as well that `API_BASE/get_token/` or sth makes sense. I'm wondering whether the rest of the `api-auth` part is even necessary, since Django itself provides the login facilities?

ltshb: I would keep all endpoints at the same base, this is for the user easier to navigate. Moreover if we need in the future to give the login to external users, we could add a link to it in the landing page and we would have a consistent URL path. I would also move the admin page in the `API_BASE`.

app/stac_api/views.py
```diff
@@ -85,7 +85,7 @@ def checker(request):
     return JsonResponse(data)
 
 
-class CollectionList(generics.ListAPIView, mixins.CreateModelMixin):
+class CollectionList(generics.ListAPIView, mixins.CreateModelMixin):  # pylint: disable=too-many-ancestors
```
ltshb, line 88: This is part from #132 and need to be removed from this PR.

app/stac_api/views.py
```diff
@@ -123,7 +123,7 @@ def get(self, request, *args, **kwargs):
         return Response(data)
 
 
-class CollectionDetail(generics.RetrieveAPIView, mixins.UpdateModelMixin):
+class CollectionDetail(generics.RetrieveAPIView, mixins.UpdateModelMixin):  # pylint: disable=too-many-ancestors
```
ltshb, line 126: This is part from #132 and need to be removed from this PR.

app/tests/test_collections_endpoint.py
```diff
@@ -89,9 +121,13 @@ def test_collections_limit_query(self):
...
+        self.client.login(username=self.username, password=self.password)
 
         response = self.client.post(
-            f"/{API_BASE}/collections", data=payload_json, content_type='application/json'
+            f"/{API_BASE}/collections",
+            data=json.dumps(payload_json),
+            content_type='application/json'
         )
         response_json = response.json()
         self.assertEqual(response.status_code, 201, msg=get_http_error_description(response_json))
```
ltshb: Mixin up PR is the door to the merging hell.

ltshb (changes requested):
We need to be careful when rebasing a PR on another one to not mix them up. It is OK for quick testing but don't mix PRs code.
Here is my suggestion, rebase this PR to develop (removing all code from #132) then add only a unittest for the new login and token endpoints. The current endpoints on `develop` don't require any authentication, this means that this PR can be without conflict been already merged in the develop branch. We need then to synchronize between #132 and [BGDIINF_SB-1482](https://jira.swisstopo.ch/browse/BGDIINF_SB-1482) [BGDIINF_SB-1482] to see who is merging first.

app/stac_api/urls.py
hansmannj: The reason for the admin page to be not in API_BASE was, that it should be independent of the current version. Hence `/v0.9/` was dropped. I think, that still makes sense, doesn't it?
For the rest I'll make it consistent and check, if the `api-auth` part is necessary.

app/tests/test_collections_endpoint.py
hansmannj: That's nicely said :-) And true.
I could rebase on develop and remove all parts from #132. Then I'd have to add the unit tests on the POST/PUT endpoints later on in another PR.

ltshb: Actually you need to add the login to the POST/PUT in this PR as I've already merged POST/PUT/PATCH/DEL of Items ;-)

app/stac_api/urls.py
hansmannj: > For the `api-toke-auth` I think as well that `API_BASE/get_token/` or sth makes sense. I'm wondering whether the rest of the `api-auth` part is even necessary, since Django itself provides the login facilities?

Yes, the `include('restframework.urls', ...)` is necessary. Otherwise there'll be no `log-in` resp. `log-out` link on the nav-bar in the browsable API.

app/tests/test_collections_endpoint.py
hansmannj: I'll add the unit tests for collection POST/PUT endpoint in another PR, soon, after #132 is merged and use superuser and anonymous user for testing.

app/config/settings_prod.py
```diff
@@ -77,6 +80,7 @@
 
 ROOT_URLCONF = 'config.urls'
 API_BASE = 'api/stac/v0.9'
+LOGIN_URL = f"/{API_BASE}/login/"
```
boecklic, line 83: the default from django is `/api/stac/admin/login` so is this additional or replacement? Can we do with just one?

boecklic (approved):

app/config/settings_prod.py
hansmannj: One is enough, you're right.
As discussed, now there's no more login button on the nav-bar and the only login available is `/api/stac/admin/login/`.

app/stac_api/urls.py
```diff
@@ -40,4 +42,5 @@
         AssetDetail.as_view(),
         name='asset-detail'
     ),
+    path(f"{API_BASE}/get_token", obtain_auth_token, name='get_token'),
```
ltshb, line 45: name should be kebabcase like other path names; `name='get-tocken'`

app/tests/test_collections_endpoint.py
```diff
@@ -240,3 +255,39 @@ def test_collection_patch(self):
...
+        # for the start, the collection[1] has to have a different licence than the payload
+        self.assertNotEqual(self.collections[1].license, payload_json['license'])
+        # for start the payload has no description
+        self.assertNotIn('title', payload_json.keys())
+        response = self.client.patch(
+            f"/{API_BASE}/collections/{payload_json['id']}",
+            data=payload_json,
+            content_type='application/json'
+        )
+        self.assertEqual(401, response.status_code, msg="Unauthorized patch was permitted.")
```
ltshb, line 293: I would also add the `DEL` case that should return 404 (because it not implemented) with a comment that if it gets implemented then it needs to return 401. This way we are sure that if DEL gets in future implemented we don't forget to have it also tested against  login.

app/tests/test_get_token_endpoint.py
```diff
@@ -0,0 +1,44 @@
...
+        self.factory = APIRequestFactory()
+        self.username = 'SherlockHolmes'
+        self.password = '221B_BakerStreet'
+        self.user = get_user_model().objects.create_user(
+            self.username, 'top@secret.co.uk', self.password
+        )
+        self.user.save()
+
+    def test_get_token_with_valid_credentials(self):
+        url = reverse('get_token')
```
ltshb, line 28: I like that :-)

app/tests/test_get_token_endpoint.py
```diff
@@ -0,0 +1,44 @@
...
+            generated_token,
+            token_from_db.key,
+            msg="Generated token and token stored in DB do not match."
+        )
+
+    def test_get_token_with_invalid_credentials(self):
+        url = reverse('get_token')
+        response = self.client.post(url, {'username': self.username, 'password': 'wrong_password'})
+        self.assertEqual(
+            400, response.status_code, msg="Token for unauthorized user has been created."
```
ltshb, line 43: Should not be a 401 here ?

ltshb (approved):
Looks good :+1:  just minor comment about naming convention and a use case test.

app/stac_api/urls.py
hansmannj: Done

app/tests/test_collections_endpoint.py
hansmannj: Good point. Added tests for authorized and unauthorized delete requests.

app/tests/test_get_token_endpoint.py
hansmannj: I get a 404 when trying with wrong credentials..

hansmannj: I get a 400 when trying with wrong credentials..

PR #132, opened 2020-12-03
BGDIINF_SB-1481: Collection Endpoint Put and Post

rebert:


app/stac_api/views.py
```diff
@@ -74,13 +75,15 @@ def checker(request):
     return JsonResponse(data)
 
 
-class CollectionList(generics.ListAPIView):
+class CollectionList(generics.ListAPIView, mixins.CreateModelMixin):
```
ltshb, line 78: Here I would use the `GenericAPIView` instead of `ListAPIView` as you are overwritting the `get()` method, the `ListAPIView` doesn't bring any plus value.

app/stac_api/views.py
```diff
@@ -110,11 +113,14 @@ def list(self, request, *args, **kwargs):
         return Response(data)
 
 
-class CollectionDetail(generics.RetrieveAPIView):
+class CollectionDetail(generics.RetrieveAPIView, mixins.UpdateModelMixin):
```
ltshb, line 116: To be consistent I would also use `GenericAPIView` and explicitely add the `get()` method, this way we are consistent in the views. It adds a little bit more of code but it isn't too much and it make the code easier to understand in IMHO.

app/tests/sample_data_test/collections/valid_collection_set_1.json
```diff
@@ -0,0 +1,30 @@
+{
+  "crs": [
+    "http://www.opengis.net/def/crs/OGC/1.3/CRS84"
+  ],
```
ltshb, line 4: crs should be a read only field and not valid in POST/PUT request

app/tests/test_collections_endpoint.py
```diff
@@ -26,6 +27,7 @@ def setUp(self):
         self.factory = APIRequestFactory()
         self.collections, self.items, self.assets = db.create_dummy_db_content(4, 4, 4)
         self.maxDiff = None  # pylint: disable=invalid-name
+        self.sample_collections = get_sample_data('collections')
```
ltshb, line 30: I like that :-)

app/tests/test_collections_endpoint.py
```diff
@@ -84,3 +86,130 @@ def test_collections_limit_query(self):
...
+        )
+
+        response_json = response.json()
+
+        self.assertEqual(response.status_code, 200, msg=get_http_error_description(response_json))
+        self.assertEqual(response_json['title'], payload_json['title'])
+
+    def test_collection_put_change_id(self):
+        payload_json = self.sample_collections['valid_collection_set_3']
+        logger.debug(self.collections[0].name)
```
ltshb, line 185: Maybe an assert to be sure that `collection.id`  in `self.collections[0]` and in `payload_json` would be good here. To avoid changing names in future and breaking this unittest without knowing it.

app/tests/test_collections_endpoint.py
```diff
@@ -84,3 +86,130 @@ def test_collections_limit_query(self):
...
+
+        # the old collection shouldn't exist any more
+        response = self.client.get(f"/{API_BASE}/collections/{self.collections[0].name}")
+        response_json = response.json()
+        self.assertEqual(response.status_code, 404, msg=get_http_error_description(response_json))
+
+    def test_collection_put_none_values(self):
+        collection_name = self.collections[1].name
+        payload_json = self.sample_collections['valid_min_collection_set_2']
+        payload_json['id'] = collection_name
```
ltshb, line 207: Maybe test name should be `test_collection_put_remove_title_provider` or `test_collection_put_remove_optional_fields`. Also before the PUT some assert to make sure that `title` and `provider` are not already empty would increase the unittest safety.

app/tests/utils.py
```diff
@@ -11,3 +19,22 @@ def get_http_error_description(json_response):
...
+
+
+def get_sample_data(topic):
+    '''Get a dictionary of sample data as json
+
+    This function takes a string describing the subpath to the sample data and returns a
+    dictionary with key = filename, value = json.
+    f.ex. get_sample_data('collections')
+    :param topic: the name of the topic (f.ex. collections)
+    :return: dict
```
ltshb, line 31: I like this way of defining param and return values in docstrings however it doesn't match our guidelines. So I would make it match the guidelines but we might want also to rediscussed this guideline and change it ? @boecklic what is your opinion about this docstring guideline ?

ltshb (commented):
Very nice unittesting especially with the sample data I like that :+1: 
Few minor comments about styling and improvement in unit test for long term.

app/tests/sample_data_test/collections/valid_collection_set_1.json
rebert: @ltshb 
`  "crazy-stuff": [
    "adding this as well"
  ],`

this works as well. Shall we explicitly give a error, when read only fields have been added to the request?

app/stac_api/serializers.py
```diff
@@ -329,6 +329,7 @@ class Meta:
     summaries = serializers.JSONField(read_only=True)
     stac_extensions = serializers.SerializerMethodField(read_only=True)
     stac_version = serializers.SerializerMethodField(read_only=True)
+    title = serializers.CharField(allow_blank=False, allow_null=False, default=None)
```
hansmannj, line 332: This is to avoid empty titles to be shown, I guess.
I am not sure, if the `allow_blank=False` and `allow_null=False` are needed here.
But I guess, a `required=False` could make sense here. Perhaps additionally or to replace the `allow_blank` and `allow_null` options.

http://www.tomchristie.com/rest-framework-2-docs/api-guide/serializers
"ote: When validation is applied to a ModelSerializer, both the serializer fields, and their corresponding model fields must correctly validate. If you have optional fields on your model, make sure to correctly set blank=True on the model field, as well as setting required=False on the serializer field."

app/tests/sample_data_test/collections/invalid_collection_set_1.json
```diff
@@ -0,0 +1,30 @@
+{
+  "crss": [
+    "http://www.opengis.net/def/crs/OGC/1.3/CRS84"
+  ],
+  "description": "A European Affrican Swallow",
+  "id": "europen-swallow",
```
hansmannj, line 6: any of these can carry coconuts ;-)

hansmannj:
Very nice, thanks :smiley: 
And yes, very good unit test coverage :+1:

app/tests/sample_data_test/collections/valid_collection_set_1.json
ltshb: That's a good point. @boecklic what is your opinion ? Should we still silently ignore unknown attributes or return a 400 BAD REQUEST ?

Advantage of returning 400:

- Clear usage of API, we will not have people complaining that the API is buggy when trying to set the CRS and nothing is happening

Advantage of silently ignoring unknown/invalid attribute:

- In the use case where we would switch between new/old version of the API with new attributes, we will get less troubles. Not sure if this use case will ever happen.

ltshb: In any case I would remove the crs from the data sample and if we need to return a 400 in this case, this should be done in a separate PR.

app/stac_api/serializers.py
rebert: @hansmannj This change has been made that the endpoint only returns a title attribute, when it has content

app/tests/sample_data_test/collections/valid_collection_set_1.json
rebert: @ltshb In any case I will remove the crs from the data sample

app/stac_api/serializers.py
ltshb: Good point we must set `required=False` in the serializer for optional fields. @hansmannj having here `required=False`, `allow_blank=False` (or `allow_null=False` depending on the field type) and `default=None` allow the framework to have the field optional and to have the PUT method behave correctly, meaning that if the field is not in the PUT request, it is set to None in the DB, which removes the field from the answer due to our `NonNullModelSerializer` base class. Whithout this config, the PUT request will behave like a PATCH in case an optional field is missing in the request body.
That being said to be consistent with the Item serializer this title field should be:
`title = serializers.CharField(required=False, allow_blank=False, default=None, max_length=255)`
It may be good to have a comment about this in code (adding it also for the Item.properties.title), this is for the moment only commented in git commit https://github.com/geoadmin/service-stac/pull/134/commits/c07541989c19a1cf348f1737ccaa48d431399096

I would remove the `allow_null=False` because it is anyway the default value and because of this comment from the Rest Framework docu (https://www.django-rest-framework.org/api-guide/fields/#charfield)

> The allow_null option is also available for string fields, although its usage is discouraged in favor of allow_blank. It is valid to set both allow_blank=True and allow_null=True, but doing so means that there will be two differing types of empty value permissible for string representations, which can lead to data inconsistencies and subtle application bugs.

rebert:
Validation of JSON as a new task: https://jira.swisstopo.ch/browse/BGDIINF_SB-1525

app/stac_api/serializers.py
```diff
@@ -338,6 +338,7 @@ class Meta:
     summaries = serializers.JSONField(read_only=True)
     stac_extensions = serializers.SerializerMethodField(read_only=True)
     stac_version = serializers.SerializerMethodField(read_only=True)
+    title = serializers.CharField(required=False, allow_blank=False, default=None)
```
hansmannj, line 341: probably add `max_length=255` here?

PR #133, opened 2020-12-08
BGDIINF_SB-1510: fixed invalid datetime range | added unit tests

hansmannj:
* now end_date cannot be before start_date in items datetime ranges
* added unit tests for item model and item serializer
* in case the item's (start_, end_)datetime property is not an instance of datetime, it is now checked, if it is a valid datetime string.

### DIFF
--- a/app/stac_api/models.py
+++ b/app/stac_api/models.py
@@ -1,5 +1,6 @@
 import logging
 import re
+from datetime import datetime
 
 from django.contrib.gis.db import models
 from django.contrib.gis.geos import GEOSGeometry
@@ -128,6 +129,20 @@ def validate_item_properties_datetimes_dependencies(
             raise ValidationError(_(message))
 
 
+def validate_datetime_string(datetime_string):
+    '''
+    Check if a given string is a datetime string in the expected format or
+    raise an error, if not.
+
+    Raises:
+        ValueError
+    '''
+    try:
+        datetime.strptime(datetime_string, '%Y-%m-%dT%H:%M:%SZ')
+    except:
+        raise ValueError(_('Invalid datetime string'))
+
+
 def validate_item_properties_datetimes(
     properties_datetime, properties_start_datetime, properties_end_datetime, partial=False
 ):
@@ -142,6 +157,16 @@ def validate_item_properties_datetimes(
             properties_start_datetime,
             properties_end_datetime,
         )
+
+        if properties_datetime is None:
+            if not isinstance(properties_start_datetime, datetime):
+                validate_datetime_string(properties_start_datetime)
+            if not isinstance(properties_end_datetime, datetime):
+                validate_datetime_string(properties_end_datetime)
+            if properties_end_datetime < properties_start_datetime:
+                message = "Property end_datetime can't refer to a date earlier than property "\
+                "start_datetime"
+                raise ValidationError(_(message))
 
 
 def get_conformance_default_links():
--- a/app/stac_api/models.py
+++ b/app/stac_api/models.py
@@ -15,6 +15,7 @@ from solo.models import SingletonModel
 
 from stac_api.collection_summaries import update_summaries
 from stac_api.temporal_extent import update_temporal_extent
+from stac_api.utils import fromisoformat
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


boecklic (approved):

app/stac_api/models.py
```diff
@@ -510,6 +511,24 @@ def delete(self, *args, **kwargs):  # pylint: disable=signature-differs
...
+            self.properties_start_datetime, datetime
+        ):
+            message = "Property start_datetime must be valid datetime.datetime or None."
+            raise ValidationError(_(message))
+
+        if self.properties_end_datetime is not None and not isinstance(
+            self.properties_end_datetime, datetime
+        ):
+            message = "Property end_datetime must be valid datetime.datetime or None."
+            raise ValidationError(_(message))
```
ltshb, line 530: Not a good idea, the fields itself should do the validation, and as it is a DatetimeField it will checks for valid datetime string. This function might be called with valid datetime string instead of datetime object and the transformation happens later. How and where did you had an issue that needed this code ?

ltshb (changes requested):

app/stac_api/models.py
hansmannj: As discussed I added a check (in a separate function `validate_datetime_string()`) that validates in case the datetime_property is not an instance of datetime if it is a valid datetime string.

app/stac_api/models.py
```diff
@@ -128,6 +129,20 @@ def validate_item_properties_datetimes_dependencies(
...
+def validate_datetime_string(datetime_string):
+    '''
+    Check if a given string is a datetime string in the expected format or
+    raise an error, if not.
+
+    Raises:
+        ValueError
+    '''
+    try:
+        datetime.strptime(datetime_string, '%Y-%m-%dT%H:%M:%SZ')
```
ltshb, line 141: 2020-12-14T08:00:00+00:00 is also valid. Please use here the helper function `utils.fromisoformat()` which handle correctly the timezone.

app/stac_api/models.py
```diff
@@ -128,6 +129,20 @@ def validate_item_properties_datetimes_dependencies(
...
+    except:
```
ltshb, line 142: very bad practice `except:`, always catch explicit Exceptions, `expcept ValueError as error:`

app/stac_api/models.py
```diff
@@ -143,6 +158,16 @@ def validate_item_properties_datetimes(
...
 
+        if properties_datetime is None:
+            if not isinstance(properties_start_datetime, datetime):
+                validate_datetime_string(properties_start_datetime)
+            if not isinstance(properties_end_datetime, datetime):
+                validate_datetime_string(properties_end_datetime)
+            if properties_end_datetime < properties_start_datetime:
+                message = "Property end_datetime can't refer to a date earlier than property "\
+                "start_datetime"
+                raise ValidationError(_(message))
```
ltshb, line 169: This should be inside `validate_item_properties_datetimes_dependencies()` as it is also a dependency check. This check won't work if the datetime is a string, the `<` will failed with strings. Instead of using `validate_datetime_string()` we should use the method `utils.fromisoformat()` that return a datetime object. This function raises a `ValueError` in case of invalid string, therefore the code should be wrapped around with a `try: ...; except ValueError as error; raise ValidationError() from error`

ltshb (changes requested):
Maybe also add a unittest case only for `validate_item_properties_datetimes()` method with passing once valid datetime string and once with invalid datetime string to test the datetime to string convertion.

app/stac_api/models.py
```diff
@@ -111,6 +113,21 @@ def validate_item_properties_datetimes_dependencies(
...
+    try:
+        if not isinstance(properties_datetime, datetime) and properties_datetime is not None:
+            properties_datetime = fromisoformat(properties_datetime)
+        if not isinstance(
+            properties_start_datetime, datetime
+        ) and properties_start_datetime is not None:
+            properties_start_datetime = fromisoformat(properties_start_datetime)
+        if not isinstance(
+            properties_end_datetime, datetime
+        ) and properties_end_datetime is not None:
```
ltshb, line 125: Just minor code style comment, having `if .. and ..:` enclosed in brackets make the code easier to read; 
```
if (not isinstance(properties_start_datetime, datetime) and
    properties_start_datetime is not None
):
```
Yapf does a better wrapping with brackets as without.

ltshb (approved):
:+1:  just a minor comment about styling but its all good

PR #135, opened 2020-12-09
BGDIINF_SB-1482: Updated Transactional SPEC

ltshb:
* Some clean up in the error schema and examples
* Added ETag and Location headers to Collection
* Corrected Item POST/PUT/PATCH responses to give back the created/updated item
* Corrected the DEL of Asset/Item with the 200 OK instead of 204 No Content

NOTE: The ETag might still requires some changes based on up coming implementation.

### DIFF
--- a/spec/transaction/transaction.yml
+++ b/spec/transaction/transaction.yml
@@ -34,6 +34,10 @@ paths:
       responses:
         "201":
           $ref: "#/components/responses/Collection"
+        "403":
+          $ref: "#/components/responses/PermissionDenied"
+        "500":
+          $ref: "#/components/responses/ServerError"
   "/collections/{collectionId}":
     put:
       tags:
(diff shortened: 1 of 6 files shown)


spec/transaction/transaction.yml
```diff
@@ -939,6 +839,12 @@ components:
     #       items:
     #         $ref: "#/components/schemas/link"
     #       type: array
+    itemIdUpdate:
+      description: >-
+        Item identifier (unique per collection. If it doesn't match the `featureId` in path
+        parameters, then the Item is renamed.
+      example: smr200-200-4-2019
+      type: string
```
boecklic, line 847: To be discussed whether we want/should/must support this case, same with renaming assets.

spec/transaction/transaction.yml
```diff
@@ -743,129 +713,59 @@ components:
...
+    updateItem:
+      allOf:
+        - type: object
+          required:
+            - id
+          properties:
+            id:
+              $ref: "#/components/schemas/itemIdUpdate"
+        - $ref: "#/components/schemas/writeItem"
+    partialItem:
```
boecklic, line 746: nice solution!

boecklic (approved):
:+1: 
one thing we'll need to discuss with Business: do we want/should we support renaming of items/assets or not? In principle I would say yes, but potential consequences must be clear to everybody.

rebert (approved):
reasonable changes for me

spec/transaction/transaction.yml
ltshb: I've asked Pasquale and Dave on Jira about this. I'll wait on their answer.

ltshb: I'll merge as is and if the rename needs to be removed we can create a new PR, we will anyway need to change the code.

PR #138, opened 2020-12-10
BGDIINF_SB-1515: Collection Endpoint: Do not show empty provider

rebert:
When the provider is None, this attribute will not to be shown

in the collection endpoint any more.

### DIFF
--- a/app/stac_api/serializers.py
+++ b/app/stac_api/serializers.py
@@ -99,6 +99,8 @@ class NonNullModelSerializer(serializers.ModelSerializer):
             for key, value in obj.items():
                 if isinstance(value, dict):
                     filtered_obj[key] = filter_null(value)
+                # then links array might be empty at this point,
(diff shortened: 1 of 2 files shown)


app/stac_api/serializers.py
```diff
@@ -99,8 +99,9 @@ def filter_null(obj):
             for key, value in obj.items():
                 if isinstance(value, dict):
                     filtered_obj[key] = filter_null(value)
-                elif isinstance(value, list) and len(value) > 0:
-                    filtered_obj[key] = value
+                elif isinstance(value, list) and key != 'links':
```
ltshb, line 102: Maybe add a comment that links array might be empty at this point but that in the view auto generated links are added anyway.

ltshb (approved):
:+1:  just a minor comment

hansmannj (approved):
:+1:

PR #139, opened 2020-12-10
BGDIINF_SB-1521: Added ETag support and updated SPEC accordingly

ltshb:
Added ETag, If-Match and If-None-Match headers support and updated SPEC accordingly The If-None-Match on
PUT/PATH/DEL has not been documented, because it doesn't make sense to
use it, although it is supported by the framework.

For this I had to rework quite a lot the collection updates functions (extent and summaries) in order to having saving to DB when not needed but also avoiding saving several times in a row. This was required because now saving is changing the ETag and we don't want to change the Etag is nothing changed. The rework also brings more code clarity and consistency.
There is still improvement, especially by moving the temporal extent and summaries update functions within a mixin class, this could/should be done within [BGDIINF_SB-1545](https://jira.swisstopo.ch/browse/BGDIINF_SB-1545)


rebert (approved):
By doing some functional tests, this did not work for me anymore:



Request Method: | GET
-- | --
http://localhost:8000/api/stac/v0.9/collections/ch.swisstopo.swisstlmregio
3.1.4
KeyError
'item_name'
/home/ltrea/service_stac/app/stac_api/views.py, line 36, in generate_item_etag
/home/ltrea/service_stac/.venv/bin/python3
3.7.5
['/home/ltrea/service_stac/app',  '/usr/lib/python37.zip',  '/usr/lib/python3.7',  '/usr/lib/python3.7/lib-dynload',  '/home/ltrea/service_stac/.venv/lib/python3.7/site-packages']
Fri, 11 Dec 2020 08:37:28 +0000

rebert (changes requested):
Request Method: | GET
-- | --
http://localhost:8000/api/stac/v0.9/collections/ch.swisstopo.swisstlmregio
3.1.4
KeyError
'item_name'
/home/ltrea/service_stac/app/stac_api/views.py, line 36, in generate_item_etag
/home/ltrea/service_stac/.venv/bin/python3
3.7.5
['/home/ltrea/service_stac/app',  '/usr/lib/python37.zip',  '/usr/lib/python3.7',  '/usr/lib/python3.7/lib-dynload',  '/home/ltrea/service_stac/.venv/lib/python3.7/site-packages']
Fri, 11 Dec 2020 08:37:28 +0000

ltshb:
> item_name

Good catch, I quickly added the eTag to the collection but didn't changed the logic of the hash that should be done on the collection and not on the item for collection.

Makefile
```diff
@@ -198,7 +198,8 @@ dockerrun: dockerbuild-debug
 
 .PHONY: clean_venv
 clean_venv:
-	pipenv --rm
+	# ignore pipenv errors by adding command prefix -
+	-pipenv --rm
```
boecklic, line 202: what kind of errors are these? is the env still removed or do we have to verify that it's actually removed?

app/stac_api/views.py
```diff
@@ -29,6 +31,55 @@
...
+
+    Args:
+        queryset: QuerySet
+            Queryset from which to compute an ETag
+
+    Returns:
+        ETag string
+    '''
+    return hashlib.sha256(''.join([str(i) for i in queryset.values()[0].values()]).encode('utf-8')
+                         ).hexdigest()
```
boecklic, line 45: Since the etag is valid for a single object, shouldn't we better pass directly the object itself? also in the methods below we always expect to have only one matching object, wouldn't it be better to directly `.get()` the object?

Second question: are we sure that this doesn't duplicate the DB requests? (once to receive the object in the view function, once to calculate the etag in the decorator)?

boecklic (commented):
I understand the implementation of the `generate_*_etag` functions is such for being able to use the third-party package; is it optimal that way (see comment)?

Makefile
ltshb: Yes it happens if the environment has already been removed, then the following clean step are not done because `pipenv --rm` act as `rm -r` and not as `rm -rf`

app/stac_api/views.py
ltshb: I could use `get()` instead of `filter()` and enclosed the code in a `try ... except DoesNotExist as err` different approach, don't know if it has different impact on performance ?

You are right this duplicates the DB request, another idea that I had was to compute the ETag in the save as a hidden DB field and to only retrive this field, but I don't know if this has a better performance as retrieving the whole object ? Any other idea on how to compute the ETag ?

ltshb: From this https://www.willmcgugan.com/blog/tech/post/etag-magic-with-django/ we could use the cache (although I would not use a random number, but either the hash of the object content), but not sure if the cache would work with our k8s setup ?

ltshb: After some googling and django docu reading here are my propositions:

1. Use cache as mention in previous comment (add some complexity, don't know how it will work with CloudFront)
1. Use `Collection.objects.get(name=kwargs['collection_name']).only('etag').values('etag')` to retrieve the ETag, this one is then in a hidden DB field and calculated on `save()`. This does a small and quick DB query which only retrieves the Etag
1. Leave it as now (only using the `get()` instead of `filter()`. The use of `values()` is a improvement in performance as the ORM is not used and the fields are not converted in python native)

I suggest using 2., what are your opinions @boecklic @rebert  @hansmannj  ? or any other suggestions ?

Makefile
```diff
@@ -139,7 +139,7 @@ lint:
 test:
 	# Collect static first to avoid warning in the test
 	$(PYTHON) $(DJANGO_MANAGER) collectstatic --noinput
-	$(PYTHON) $(DJANGO_MANAGER) test --verbosity=2 $(TEST_DIR)
+	$(PYTHON) $(DJANGO_MANAGER) test --verbosity=2 --parallel 8 $(TEST_DIR)
```
hansmannj, line 142: :+1:

app/stac_api/views.py
rebert: I'm not in the details. As it is only a lookup and won't impact the performance a lot, I tend to persist the ETag in the DB: 2

hansmannj: Without having had too much time to dig into details here, I'd also +1 for option 2. 
Would a re-computation of the ETag also be required and triggered, if for example in case of an asset only the data file on S3 changes, but none of the meta data?

ltshb: @hansmannj  you bring a good point ! Currently I compute the ETag only based on the Model Fields which means that if a Collection.Provider or a Link or an Asset changed the ETag stays the same, which is not what we want ! So solution 3. don't work as we would expect. I will then try to implement solution 2. with taking care about model relations and add unit test for this use case.

rebert (approved):
@ltshb yes, quite a lot of refactoring. I tested the PR functionally and could not find any bugs.

rebert:
detail: I noticed that there was no rebuild of the specs with the last changes

ltshb:
> detail: I noticed that there was no rebuild of the specs with the last changes

good catch I've rebuild the spec.

PR #140, opened 2020-12-11
BGDIINF_SB-1478: field type of Asset has to be required and validated

rebert:
Putting all validators into a sepperate file


rebert:
@here It seems that the validator _validate_item_properties_datetimes_dependencies_ is not being used any more. Is it ok, when I remove it?

hansmannj:
> @here It seems that the validator _validate_item_properties_datetimes_dependencies_ is not being used any more. Is it ok, when I remove it?

From my understanding it is still in use. It is called from within `validate_item_properties_datetimes()` which is called from within the item's `clean()`. So I'd suggest to leave it as is.

hansmannj:
P.S. I guess the ticket still says, that `type` (called `media_type` in `models.py`) should be set to required. This has already been done ;-)

rebert:
@here:
- I did not use the field type Choices, as the media types are sometimes with, sometimes without space
- Even though it has defined diffined:
  `image/tiff; application=geotiff; profile=cloud-optimized` 
  I left it:
  `image/tiff; application=geotiff; profile=cloud-optimize` (without **d**)

app/stac_api/validators.py
```diff
@@ -0,0 +1,196 @@
...
+from django.core.exceptions import ValidationError
+from django.utils.translation import gettext_lazy as _
+from stac_api.utils import fromisoformat
+
+logger = logging.getLogger(__name__)
+
+MEDIA_TYPES = [
+    ('application/x.asc+zip', 'Zipped ASCII (.asc)', '.zip'),
+    ('application/x.asc-grid+zip', 'Zipped ASCII-GRID (.xyz)', '.zip'),
+    ('application/x.e00+zip 	Zipped', 'e00', '.zip'),
```
ltshb, line 15: Remove tab in media type

app/stac_api/validators.py
```diff
@@ -0,0 +1,196 @@
...
+
+def validate_item_properties_datetimes_dependencies(
+    properties_datetime, properties_start_datetime, properties_end_datetime
+):
+    '''
+    Validate the dependencies between the Item datetimes properties
+	This makes sure that either only the properties.datetime is set or
+	both properties.start_datetime and properties.end_datetime
+	Raises:
+		django.core.exceptions.ValidationError
```
ltshb, line 141: TABS ! only space in code.

app/stac_api/models.py
```diff
@@ -656,6 +531,12 @@ def from_db(cls, db, field_names, values):
     def __str__(self):
         return self.name
 
+    def clean(self):
+        # as the media type sometimes is with space, sometimes without
+        # it will alway be stored without
+        if self.media_type:
+            self.media_type = self.media_type.replace(' ', '')
```
ltshb, line 538: This is a field only clean step and should be placed in `clean_media_type()` method (see https://docs.djangoproject.com/en/3.1/ref/forms/validation/#form-and-field-validation)

ltshb (changes requested):
Good only issue regarding space and TAB.

In my opinion we have to decide the separator for the media type:

1. SEMI-COLON + SPACE (e.g. `; `)
1. SEMI-COLON (e.g. `;`)

Then the media type list should be done accordingly.

Then the question do we want to support extra space separator and automatically strip them ? I would not, in the admin page the field is a choice field anyway and the API will be used by machine with correct type string. @boecklic what's your opinion ?

app/stac_api/validators.py
```diff
@@ -0,0 +1,196 @@
+import logging
```
hansmannj, line 1: :+1:
Good idea to source-out the validators to a separate file :smile:

app/stac_api/validators.py
hansmannj: They use `; ` (semi-colon + space) here: https://github.com/radiantearth/stac-spec/blob/v0.9.0/item-spec/item-spec.md#media-types.
So I'd suggests to stick to that, unless it causes problems with the choice field. In that case I'd drop the extra space and the world will still keep on turning ;-)

hansmannj: P.S. and they call it `cloud-optimized` (with d)

hansmannj (approved):
Thanks, LGTM :smiley:

PR #142, opened 2020-12-14
BGDIINF_SB-1482: Fixed Asset SPEC POST/PUT/PATCH

ltshb:
Same issue as for the GET Asset SPEC, see  #141


rebert (approved):
thanks

hansmannj (approved):
:smile:

PR #144, opened 2020-12-14
BGDIINF_SB-1483 asset write endpoints

hansmannj:
* added asset POST, PUT, PATCH and DELETE
* added unit tests

### DIFF
--- a/app/stac_api/views.py
+++ b/app/stac_api/views.py
@@ -287,10 +287,26 @@ class ItemDetail(
         return self.destroy(request, *args, **kwargs)
 
 
-class AssetsList(generics.GenericAPIView):
+class AssetsList(generics.GenericAPIView, views_mixins.CreateModelMixin):
     serializer_class = AssetSerializer
     queryset = Asset.objects.all()
     pagination_class = None
+
+    def get_write_request_data(self, request, *args, **kwargs):
+        data = request.data.copy()
+        data['item'] = kwargs['item_name']
+        return data
+
+    def get_success_headers(self, data):
+        asset_link_self = self.request.build_absolute_uri() + "/" + self.request.data["id"]
+        try:
+            return {'Location': asset_link_self}
+        except KeyError as err:
+            logger.error('Failed to set the Location header for asset creation: %s', err)
+            return {}
+
+    def post(self, request, *args, **kwargs):
+        return self.create(request, *args, **kwargs)
 
     def get_queryset(self):
         # filter based on the url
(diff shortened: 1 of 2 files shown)


app/stac_api/views_mixins.py
```diff
@@ -26,21 +32,32 @@ def get_write_request_data(self, request, *args, **kwargs):
...
+            try:
+                return {'Location': get_link(data['links'], 'self', raise_exception=True)['href']}
+            except KeyError as err:
+                logger.error('Failed to set the Location header for item creation: %s', err)
+                return {}
+        else:
+            try:
+                return {'Location': asset_link_self}
+            except KeyError as err:
+                logger.error('Failed to set the Location header for asset creation: %s', err)
```
ltshb, line 59: The handling of location could be easily done within `AssetsList` class  (and should be) by overriding `get_success_headers()` method. This method in `CreateModelMixin` should simply be extended with the `request` parameter.

But a fundamental questions we may be better add `links` to the Asset like in Collection and Item, this way our API is more consistent, what do you think @boecklic and @rebert  ? The asset is anyway swisstopo specific and not part of the original STAC we can define it as we want.

app/tests/test_assets_endpoint.py
```diff
@@ -76,3 +87,256 @@ def test_single_asset_endpoint(self):
...
+
+    def test_asset_endpoint_patch_rename_asset(self):
+        collection_name = self.collections[0].name
+        item_name = self.items[0][0].name
+        asset_name = self.assets[0][0][0].name
+        new_asset_name = "new-name-123abc"
+        data = {
+            "id": new_asset_name,
+            "checksum:multihash": "01205c3fd6978a7d0b051efaa4263a09",
+            "type": "image/tiff; application=geotiff; profile=cloud-optimize",
```
ltshb, line 255: To better test patch I would only add the id in the payload to only rename it.

ltshb (changes requested):
Good just an issue with the location handling and open question about it.

When it then done I would recommand to squash all commits together, IMHO one commit is enough for this work ;)

app/stac_api/views_mixins.py
hansmannj: Good point, I'll implement the handling of the location in the `get_success_headers()` inside `AssetList`.
As for the general question of including links to the assets like in collection and item, I'd suggest we discuss this and - if agreed - create a separate ticket for this (in order to get this PR merged soon).
From my point of view we could add links to be consistent. It shouldn't be a big deal and probably is helpful for users of the browseable API as well.

app/tests/test_assets_endpoint.py
hansmannj: Makes sense, I'll change that.

ltshb (approved):
Simply remove the try catch and is all good. I'll create a ticket to discuss about the links in the assets.

app/stac_api/views.py
```diff
@@ -287,11 +287,27 @@ def delete(self, request, *args, **kwargs):
...
+        data = request.data.copy()
+        data['item'] = kwargs['item_name']
+        return data
+
+    def get_success_headers(self, data):
+        asset_link_self = self.request.build_absolute_uri() + "/" + self.request.data["id"]
+        try:
+            return {'Location': asset_link_self}
+        except KeyError as err:
+            logger.error('Failed to set the Location header for asset creation: %s', err)
```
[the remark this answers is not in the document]
hansmannj: done

PR #145, opened 2020-12-15
BGDIINF_SB-1512 manage assets on s3

boecklic:
This PR adds some helpers to handle
* moving assets on s3 when changing asset name/id
* deleting assets on s3 when deleting asset metadata object
  (Note that this has to be done using pre_save signal)

Furthermore it
* improves logging to log stacktrace in case of unhandled exception
* adapts docker-compose to run postgres and minio as unpriviledged
user


ltshb:
@boecklic @rebert @hansmannj this PR is ready for a first review, here below a summary on what's working and what's not working plus some future considerations.

**Admin page:**
- File upload is mandatory for creating asset
- `checksum:multihash` is automatically computed, using a *sha2-256*
  - **ISSUE:** currently file upload is apparently not done using S3 Signature V4, which means (at least from my understanding) that the `HEAD collection_name/item_name/asset_name` request doesn't returns the `x-amz-meta-sha256` header but only a md5 ETag ! So we need to either make the `x-amz-meta-sha256` work for file upload or change the checksum to md5 in the auto compute.
  - I used a workaround/hack for the `x-amz-meta-sha256` for the file upload from admin page, not sure if this is the correct way thouth.

**POST Asset:**
- *href* if given, is only checked to match `https://service_domain/collection_name/item_name/asset_name` and has been removed from SPEC.
  - Using *href* in post to move temporary asset to the correct location has an issue with the admin page. Currently the S3 asset management code is in `models.py`, there we cannot differentiate between an asset insertion from Admin page or from POST API and we cannot know without complex logic or dirty hack if the asset needs to be moved on S3 or not. Therefore for the move from temp location I would suggest to use the initial SPEC *copyfromhref* writeonly attribute in POST API, this would make the code much simpler.
- `checksum:multihash` field if provided is used to check if the asset match this hash, currently it supports md5sum and sha2-256 checksum. The asset checksum is retrieved with `HEAD collection_name/item_name/asset_name` request using either the `x-amz-meta-sha256` header or the md5 ETag header depending on the availability of the first one.
  - I would suggest to make this field mandatory for security, but I'm not sure if this would give more trouble to GDWH ?
- When no `checksum:multihash` field is provided, then it is set from the answer of the HEAD asset requests header.
- Independently of `href` and/or `checksum:multihash` fields, the asset on S3 is checked and if no asset exists then the POST is refused

**General issues**
- I'm not sure to have fully understand from where the `x-amz-meta-sha256` header comes from, I did a workaround/hack to have it in the HEAD request of the asset being uploaded via the admin page, but don't how this will work with the Asset uploaded by GDWH ?
- Unittest is becoming more and more messy, I didn't extend nor did any improvement there and I suggest to merge this PR asap (only fixing small issues first) and then I'll focused on the unittest refactoring.

ltshb:
NOTE: this PR is currently running on DEV if you want to test it

boecklic:
Hej @ltshb thanks for the effort and the great work with this PR! Some comments to your comments/questions:
> checksum:multihash is automatically computed, using a sha2-256
>   - ISSUE: currently file upload is apparently not done using S3 Signature V4,
I think we should force our code to always use S3V4, since Frankfurt is only supporting this signature version (https://aws.amazon.com/blogs/aws/amazon-s3-update-sigv2-deprecation-period-extended-modified/)
As far as I understand, the `x-amz-meta-sha256` is not used to verify the upload, but is only available via HTTP GET from the object on s3 (computed by AWS). Checking integrity of the upload still uses MD5 (which I guess is used under the hood by django_storages, so we don't have to bother with that) (https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Object.put) If you want we can try to make a call these days.

> Unittest is becoming more and more messy, I didn't extend nor did any improvement there and I suggest to merge this PR asap (only fixing small issues first) and then I'll focused on the unittest refactoring.
+1 for merging this PR and then continuing on unittests!

ltshb:
> Hej @ltshb thanks for the effort and the great work with this PR! Some comments to your comments/questions:
> 
> > checksum:multihash is automatically computed, using a sha2-256
> > 
> > * ISSUE: currently file upload is apparently not done using S3 Signature V4,
> 
> I think we should force our code to always use S3V4, since Frankfurt is only supporting this signature version (https://aws.amazon.com/blogs/aws/amazon-s3-update-sigv2-deprecation-period-extended-modified/)
> As far as I understand, the `x-amz-meta-sha256` is not used to verify the upload, but is only available via HTTP GET from the object on s3 (computed by AWS). Checking integrity of the upload still uses MD5 (which I guess is used under the hood by django_storages, so we don't have to bother with that) (https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Object.put) If you want we can try to make a call these days.
> 
> > Unittest is becoming more and more messy, I didn't extend nor did any improvement there and I suggest to merge this PR asap (only fixing small issues first) and then I'll focused on the unittest refactoring.
> 
> +1 for merging this PR and then continuing on unittests!

I think with my implementation it forces `s3v4`, at least it is configured so, but this didn't add the `x-amz-meta-sha-256` to the GET/HEAD header when retrieving an Asset, when this one was uploaded via Admin page. So the signature V4 didn't seems to have any impact on the GET/HEAD headers! To have the correct header, I had to hack the django-storage to send the `sha-256` with the `upload_fileobj()`. I think we can have a look at this after your holiday and made any changes if required then in another PR.

boecklic:
> I think with my implementation it forces `s3v4`, at least it is configured so, but this didn't add the `x-amz-meta-sha-256` to the GET/HEAD header when retrieving an Asset, when this one was uploaded via Admin page. So the signature V4 didn't seems to have any impact on the GET/HEAD headers! To have the correct header, I had to hack the django-storage to send the `sha-256` with the `upload_fileobj()`. I think we can have a look at this after your holiday and made any changes if required then in another PR.
I don't see the issue: if I upload a file via admin and then do a `curl --head <url of s3 object>` I get the correct `x-amz-meta-sha256` in the response..?

app/stac_api/models.py
```diff
@@ -579,13 +605,13 @@ class Asset(models.Model):
     )
     # using "name" instead of "id", as "id" has a default meaning in django
     name = models.CharField('id', unique=True, max_length=255, validators=[validate_name])
-    file = models.FileField(upload_to=get_upload_to_asset_path, null=True, blank=True)
+    file = models.FileField(upload_to=upload_asset_to_path_hook, max_length=255)
```
hansmannj, line 608: Would it make sense to limit the allowed file types (or at least to blacklist some file types) to prevent potentially malicious files from being uploaded?

hansmannj (approved):
Thanks @ltshb and @boecklic for this impressive effort!

As I have my day off tomorrow and don't want to block you, I'll approve, even if I'd have liked to dig deeper into this PR review than my time allowed so far.
If this can wait, I'll continue on Thursday, otherwise feel free to merge.
As far as I could see yet, things look good!

hansmannj:
> > 
> > +1 for merging this PR and then continuing on unittests!

+ (another) 1

ltshb:
> > I think with my implementation it forces `s3v4`, at least it is configured so, but this didn't add the `x-amz-meta-sha-256` to the GET/HEAD header when retrieving an Asset, when this one was uploaded via Admin page. So the signature V4 didn't seems to have any impact on the GET/HEAD headers! To have the correct header, I had to hack the django-storage to send the `sha-256` with the `upload_fileobj()`. I think we can have a look at this after your holiday and made any changes if required then in another PR.
> 
> @ltshb I don't see the issue: if I upload a file via admin and then do a `curl --head <url of s3 object>` I get the correct `x-amz-meta-sha256` in the response..?

The current version deployed on DEV is with my hack that fix this issue ;)

app/stac_api/models.py
ltshb: The file type would not block malicious, as a malware could be renamed with an allowed file extension. However this could be a good idea to have a validation, that's validate the file extension with the media type and also having the asset name validated with the media type, I've just opened a jira ticket for this: https://jira.swisstopo.ch/browse/BGDIINF_SB-1566

PR #146, opened 2020-12-15
Corrected the SPEC and unit tests for the error description type

ltshb:
Some django errors are encapsulated in object or list in order to
provide a good and detail error descrition in an organized structure. In
order not to loose this structure and organization we allow in the spec
to have the 'description' error field not only be string, but also array
and object.


rebert (approved):
LGTM

hansmannj (approved):
LGTM, too ;-)

PR #148, opened 2020-12-16
Run test in parallel to speed up testing

ltshb:
Testing took more than 200s, now with using parallel testing it is reduced to ~75s.
The number of parallel test has been set to a maximum value for
performance improving (adding more doesn't reduce the execution time).

**IMPORTANT NOTE:** For subTest the error message is not useful when running in parallel, which means that if a subTest fails we need to manually restart the tests in sequence to know what went wrong. With parallel testing and subTest we only know which test failed but don't have the error message. This is only for subTest not for regular test.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -139,7 +139,7 @@ lint:
 test:
 	# Collect static first to avoid warning in the test
 	$(PYTHON) $(DJANGO_MANAGER) collectstatic --noinput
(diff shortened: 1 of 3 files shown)


boecklic (approved):
:+1:

rebert:
@ltshb works for me. What about adding a _make test-serially_ for debugging?

rebert (approved):
thanks for this

PR #149, opened 2020-12-16
BGDIINF_SB-1525 validation of JSON in transactions

hansmannj:
* added validate_JSON_payload_write_requests() in validators.py to ensure that requests with extra-payload or read-only fields in payload fail with a 400
* added calls of new validator function in views.py by overriding get_serializer() in collections, items and assets list- and detail-view
* added unit tests to ensure that write requests for collections, items and assets return 400 when extra content or read-only fields are in payload


app/stac_api/views.py
```diff
@@ -349,6 +382,14 @@ def get_success_headers(self, data):
         asset_link_self = self.request.build_absolute_uri() + "/" + self.request.data["id"]
         return {'Location': asset_link_self}
 
+    def get_serializer(self, *args, **kwargs):
```
rebert, line 385: As this function is being uses several times, it could be defined once outside the class?

app/tests/test_assets_endpoint.py
```diff
@@ -153,6 +155,36 @@ def test_asset_endpoint_post_full(self):
         self.assertStatusCode(200, response)
         self.check_stac_item(data, json_data)
 
+    def test_asset_endpoint_post_extra_payload(self):
```
rebert, line 158: cool, but then you validate even extra "crazy:stuff"?

app/tests/test_collections_endpoint.py
```diff
@@ -117,6 +134,26 @@ def test_valid_collections_post(self):
         response_json = response.json()
         self.assertStatusCode(400, response)
 
+    def test_collections_post_extra_payload(self):
+        payload_json = self.sample_collections['valid_collection_set_1']
+        payload_json["crazy:stuff"] = "woooohoooo"
```
rebert, line 139: yes, it is xmas time :-)

rebert (commented):
my 5 cents for the code. I'm going to do some functional testing as well

app/stac_api/validators.py
```diff
@@ -187,3 +187,29 @@ def validate_item_properties_datetimes(
             properties_start_datetime,
             properties_end_datetime,
         )
+
+
+def validate_JSON_payload_write_requests(serializer_class, request_data):  # pylint: disable=invalid-name
```
ltshb, line 192: please stick to the naming convention, all lower case and don't disable invalid-name

app/stac_api/validators.py
```diff
@@ -187,3 +187,29 @@ def validate_item_properties_datetimes(
...
+    Args:
+         serializer_class: serializer class for validating and deserializing input and
+         for serializing output.
+
+         request_data: data of the request to be validated
+    Raises:
+        ValidateionError: Either in case of unexpected payload or if the payload contains
+        values for read-only fields.
+    '''
+    expected_payload = list(serializer_class().fields.keys())
```
ltshb, line 204: only create one serializer instance, but for the future asset management it would be better to already have the view instance here.

app/stac_api/validators.py
```diff
@@ -187,3 +187,29 @@ def validate_item_properties_datetimes(
...
+    expected_payload_read_only = list()
+
+    for field in serializer_class().fields.keys():
+        if serializer_class().fields[field].read_only:
+            expected_payload_read_only.append(field)
```
ltshb, line 209: This could be done in list comprehension `[field for field in serializer.fields if field.read_only]`

app/stac_api/views.py
```diff
@@ -126,6 +127,14 @@ class CollectionList(generics.GenericAPIView, mixins.CreateModelMixin):
     serializer_class = CollectionSerializer
     queryset = Collection.objects.all()
 
+    def get_serializer(self, *args, **kwargs):
+        serializer_class = self.get_serializer_class()
+        kwargs.setdefault('context', self.get_serializer_context())
+
+        validate_JSON_payload_write_requests(serializer_class, self.request.data)
+
+        return serializer_class(*args, **kwargs)
```
ltshb, line 136: This method should be generalized either in a mixin or a new based class. I would use a mixin. Also it would be better to create the serializer instance then pass it to the validator and after validation return the instance. This would be more efficient and might be required for asset management were we might need serializer dynamic field.

ltshb (changes requested):
Nice solution, just some improvement comment ;)

rebert (commented):
nice. First I tested extra payload with the forms (http://localhost:8000/api/stac/v0.9/collections/ch.swisstopo.pixelkarte-farbe-pk200.noscale) - here I got a **500**er error with extra payload. 

So I build a test case in test_collections_endpoint with the same payload. And here I get a 400 as expected...

app/stac_api/views.py
hansmannj: Makes sense, I'll see if it works to put it in the `mixins_views.py`

app/stac_api/validators.py
hansmannj: I knew it! ;-)

app/stac_api/validators.py
hansmannj: :+1:

hansmannj: Done

app/stac_api/validators.py
hansmannj: I changed it so that now the serializer instance is passed to this function. Did not really get the point with the view class so far. For what is it needed here?

hansmannj:
> First I tested extra payload with the forms (http://localhost:8000/api/stac/v0.9/collections/ch.swisstopo.pixelkarte-farbe-pk200.noscale) - here I got a **500**er error with extra payload.
> 
> So I build a test case in test_collections_endpoint with the same payload. And here I get a 400 as expected...

Thanks for testing via API, I only tested via unit tests so far. Will have a look at this!

app/stac_api/views.py
hansmannj: :+1:

app/stac_api/views.py
hansmannj: Done. Introduced class `BaseAPIView` in `views.py` with overriden `get_serializer()`.

app/stac_api/views.py
hansmannj: Done.

ltshb (commented):
I had a quick look at the first commit (validation done within `get_serializer()` method from the view). The issue that you were having with the browsable view (returning HTTP 500 instead of 400) was due to a misused of the ‘get_serialzer()’ method which was also called by the browsable API for rendering. It turns out that  `get_serializer()` method is used in several place in the code and is not an appropriate place to do validation. Also important to mention that the 500 instead of 400 issue was only on the Browsable view, not using the JSON api (e.g. with curl).
Using the serializer `validate()` method is much better, even if we have to do a small hack by using `.initial_data` because `attrs` doesn't contains unexpected attributes nor read only attributes.
So we learn a lot about the framework ;) and I suggest to keep the first commit as is and clean up the second commit (using `validate_json_payload_write_requests()` method within `validate()`) with a good commit message explaining the issue of the first implementation and the actual solution. This way we have very well documented within the code (git commit message) and within github.

hansmannj:
In a first commit the validation of JSON payload for extra fields or read-only fields in payload was done by overriding the view's `get_serializer()`. This worked, but when posting extra payload or read-only fields in payload by using the browsable API, a 500 instead of a 400 was returned.

Hence the validation was moved to `serializers.py` by creating a common serializer class with an overriden `validate()` for collections, items and assets. This works and yields the correct error codes.

app/stac_api/serializers.py
```diff
@@ -582,6 +599,21 @@ def get_href(self, obj):
...
+    def validate(self, attrs):
+        # see: https://www.django-rest-framework.org/api-guide/serializers/#object-level-validation
+        expected_payload = list(self.fields.keys())
+        expected_payload_read_only = [
+            field for field in self.fields if self.fields[field].read_only
+        ]
+
+        for key in self.initial_data.keys():
+            if key not in expected_payload:
+                raise ValidationError(_(f"Found unexpected property in payload: {key}"))
```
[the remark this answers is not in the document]
hansmannj: works!

app/tests/test_serializer.py
```diff
@@ -857,6 +857,11 @@ def test_asset_serialization(self):
         logger.debug('python native:\n%s', pformat(python_native))
 
         # translate to JSON:
+        # remove read-only properties from python_native
+        read_only_fields = ["href", "created", "updated"]
+        for key in read_only_fields:
+            del python_native[key]
+
```
ltshb, line 864: This should be done later just before de-serialization in line 890 on the `python_native_back`

app/stac_api/serializers.py
```diff
@@ -84,6 +86,36 @@ def update_or_create_links(model, instance, instance_type, links_data):
...
+        expected_payload = list(self.fields.keys())
+        expected_payload_read_only = [
+            field for field in self.fields if self.fields[field].read_only
+        ]
+
+        for key in self.initial_data.keys():
+            if key not in expected_payload:
+                raise ValidationError(detail=_("Unexpected property in payload"), code=key)
+            if key in expected_payload_read_only:
+                raise ValidationError(detail=_("Found read-only property in payload"), code=key)
```
ltshb, line 105: This could have/should have been left in the validators as a function.

app/stac_api/serializers.py
```diff
@@ -84,6 +86,36 @@ def update_or_create_links(model, instance, instance_type, links_data):
...
+        # add specific validation for ItemSerializer
+        if self.__class__.__name__ == "ItemSerializer":
+            validate_item_properties_datetimes(
+                attrs.get('properties_datetime', None),
+                attrs.get('properties_start_datetime', None),
+                attrs.get('properties_end_datetime', None),
+                partial=self.partial
+            )
+
+        return attrs
```
ltshb, line 116: I don't really like the uses of inherithance here, and if used it should be renamed as mixin, as it follow the mixin pattern. I would keep it simple by simply overriding the `validate()` method of the three main serializer (Collection, Item and Asset) which would call the validator function by passing `self.initial_data`

ltshb (commented):
I would not use here multiple inheritance it makes the architecture more complex without bringing advantage. Also using a mixin pattern to overwrite a base method as your have seen with the view doesn't always work, this means that if we need in future for a serializer to add an object validation we will get into trouble. And that is the case with asset management which will overwrite the validate to add more validation. We should also keep all validation for code clarity in validators.py. Note that moving it back to validator will introduce a small issue that I've already fixed in PR #145, the `ValidationError` in validators.py is from django, it also support the `id` parameter and the exception gets translated in rest_framework.ValidationError` in `apps.custom_exception_handler()`, however the `id` is not translated yet. I've done it but not pushed it yet. Code should still work, we just loose this id information until I merge my changes. In PR #145 I've already changed all field ValidationError to make use of this `id`.

app/stac_api/serializers.py
hansmannj: moved it back ;-)

app/stac_api/serializers.py
hansmannj: No inheritance here anymore, wrote three separate `validate()` now

ltshb (approved):
:+1:  just minor comments.

In PR #145  I've created a new module for serializer validator `validators_serializer.py`. I separated the validator that are and can only be used by the serializer for a better code comprehension and to avoid mixins django and rest_framework exceptions import within modules. On of the validator that I use in #145 requires an `rest_framework.APIException` in order to return an HTTP 500, that's why I separated it in another module. The `validators.py` module contains then only `django.core.exception.ValidationError` and can be used in models while the `validators_serializer.py` can only be used in serializers.
So depending who is merging first (probably you ;-) ) one of us needs to move your validator function in `validators_serializer.py` or you can do it anyway :-)
Merry Christmas

app/stac_api/serializers.py
```diff
@@ -674,10 +683,14 @@ def update(self, instance, validated_data):
         return super().update(instance, validated_data)
 
     def validate(self, attrs):
-        validate_item_properties_datetimes(
-            attrs.get('properties_datetime', None),
-            attrs.get('properties_start_datetime', None),
-            attrs.get('properties_end_datetime', None),
-            partial=self.partial
-        )
+        if self.__class__.__name__ == "ItemSerializer":
```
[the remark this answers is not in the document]
hansmannj: True, that's a relict from having this code elsewhere ;-)

app/stac_api/validators.py
```diff
@@ -187,3 +187,27 @@ def validate_item_properties_datetimes(
...
+    '''
+    Validate JSON payload and raise error, if extra payload or read-only fields
+    in payload are found
+    Args:
+        serializer: serializer for which payload is checked
+
+    Raises:
+        ValidationError if extra or read-only fields in payload are found
+    '''
+    # see: https://www.django-rest-framework.org/api-guide/serializers/#object-level-validation
```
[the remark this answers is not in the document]
hansmannj: If you mean the comment pointing to the drf doc, I think it still does make sense here.
It points to parts of docu that explain overriding of validate() and also accessing `.initial_data`.
So it might be useful from my point of view.

PR #151, opened 2020-12-21
BGDIINF_SB-1484: Realise the /search endpoint

rebert:
- Added a unittestfile

- Copy of view ItemList to SearchList as a base for search endpoint

- Separation of functions used in several views

### DIFF
--- a/app/stac_api/views.py
+++ b/app/stac_api/views.py
@@ -1,13 +1,11 @@
+import json
 import logging
 from collections import OrderedDict
 from datetime import datetime
-from json import JSONDecodeError
-from json import loads
-from json import dumps
 
 from django.conf import settings
-from django.contrib.gis.geos import Polygon
 from django.contrib.gis.geos import GEOSGeometry
+from django.contrib.gis.geos import Polygon
 from django.db.models import Q
 from django.http import JsonResponse
 from django.shortcuts import get_object_or_404
@@ -132,11 +130,11 @@ def filter_by_bbox(queryset, bbox):
         from the bbox string is invalid.
     '''
     try:
-        logger.debug('Item query parameter bbox = %s', bbox)
+        logger.debug('Query parameter bbox = %s', bbox)
         query_bbox_polygon = Polygon.from_bbox(bbox.split(','))
     except ValueError as error:
         logger.error(
-            'Invalid bbox parameter: '
+            'Invalid bbox query parameter: '
             'Could not transform bbox "%s" to a polygon; %s'
             'f.ex. bbox=5.96, 45.82, 10.49, 47.81',
             bbox,
@@ -145,7 +143,7 @@ def filter_by_bbox(queryset, bbox):
         raise ValidationError(
             _('Invalid bbox query parameter, '
               ' has to contain 4 values. f.ex. bbox=5.96,45.82,10.49,47.81'),
-            code='bbox'
+            code='bbox-invalid'
         )
     return queryset.filter(geometry__intersects=query_bbox_polygon)
 
@@ -185,8 +183,7 @@ def _filter_by_datetime_range(queryset, start_datetime, end_datetime):
     if start_datetime == '..':
         # open start range
         return queryset.filter(
-            Q(properties_datetime__lte=end_datetime) |
-            Q(properties_end_datetime__lte=end_datetime)
+            Q(properties_datetime__lte=end_datetime) | Q(properties_end_datetime__lte=end_datetime)
         )
     if end_datetime == '..':
         # open end range
@@ -197,8 +194,8 @@ def _filter_by_datetime_range(queryset, start_datetime, end_datetime):
         # else fixed range
     return queryset.filter(
         Q(properties_datetime__range=(start_datetime, end_datetime)) | (
-                Q(properties_start_datetime__gte=start_datetime) &
-                Q(properties_end_datetime__lte=end_datetime)
+            Q(properties_start_datetime__gte=start_datetime) &
+            Q(properties_end_datetime__lte=end_datetime)
         )
     )
 
@@ -313,7 +310,6 @@ class ItemsList(generics.GenericAPIView, views_mixins.CreateModelMixin):
 
         return queryset
 
-
     def list(self, request, *args, **kwargs):
         queryset = self.filter_queryset(self.get_queryset())
 
@@ -411,10 +407,9 @@ class SearchList(generics.GenericAPIView, mixins.ListModelMixin):
             if 'collections' in data:
                 queryset = self.filter_by_collections(queryset, data['collections'])
             if 'query' in data:
-                queryset = self.filter_by_query(queryset, dumps(data['query']))
+                queryset = self.filter_by_query(queryset, json.dumps(data['query']))
             if 'intersects' in data:
-                queryset = self.filter_by_intersects(queryset, dumps(data['intersects']))
-
+                queryset = self.filter_by_intersects(queryset, json.dumps(data['intersects']))
 
         return queryset
 
@@ -455,8 +450,8 @@ class SearchList(generics.GenericAPIView, mixins.ListModelMixin):
 
         # validate json
         try:
-            json_query = loads(query)
-        except JSONDecodeError as error:
+            json_query = json.loads(query)
+        except json.JSONDecodeError as error:
             message = f"The application could not decode the JSON." \
                       f"Please check the syntax ({error})." \
                       f"{query}"
@@ -464,7 +459,7 @@ class SearchList(generics.GenericAPIView, mixins.ListModelMixin):
             logger.error(message)
             raise ValidationError(_(message))
 
-        for attribute in json_query: # pylint: disable=too-many-nested-blocks
+        for attribute in json_query:  # pylint: disable=too-many-nested-blocks
             # iterate trough the fields given in the query parameter
             if attribute in queriable_fields:
                 logger.debug("attribute: %s", attribute)
@@ -504,7 +499,6 @@ class SearchList(generics.GenericAPIView, mixins.ListModelMixin):
                                 message = f"Invalid dateformat: ({error})"
                                 logger.error(message)
                                 raise ValidationError(_(message))
-
 
                         # __eq does not exist, but = does it as well
                         if operator == 'eq':
--- a/app/stac_api/views.py
+++ b/app/stac_api/views.py
@@ -1,9 +1,13 @@
 import logging
 from collections import OrderedDict
 from datetime import datetime
+from json import JSONDecodeError
+from json import loads
+from json import dumps
 
 from django.conf import settings
 from django.contrib.gis.geos import Polygon
+from django.contrib.gis.geos import GEOSGeometry
 from django.db.models import Q
 from django.http import JsonResponse
 from django.shortcuts import get_object_or_404
@@ -26,7 +30,9 @@ from stac_api.serializers import CollectionSerializer
 from stac_api.serializers import ConformancePageSerializer
 from stac_api.serializers import ItemSerializer
 from stac_api.serializers import LandingPageSerializer
+from stac_api.utils import fromisoformat
 from stac_api.utils import utc_aware
+from stac_api.validators import validate_geometry
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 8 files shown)


rebert:
TODO:
- [ ] intesects
- [ ] test post (pagination)
- [ ] nicer filter_by_query function (too-many-nested-blocks, too-many-locals, too-many-branches)

spec/static/openapitransactional.yaml
```diff
@@ -2137,11 +2135,6 @@ paths:
       responses:
         "200":
           description: Returns the updated Collection
-          headers:
```
rebert, line 2140: @ltshb is ok?

ltshb: Good catch :+1:

app/stac_api/views.py
```diff
@@ -344,6 +389,208 @@ def delete(self, request, *args, **kwargs):
...
+                        message = f"Invalid operator in query argument. The operator {operator} " \
+                                  f"is not supported. Use: {operators}"
+                        logger.error(message)
+                        raise ValidationError(_(message))
+            else:
+                message = f"Invalid field in query argument. The field {attribute} is not " \
+                          f"a propertie. Use one of these {queriable_fields}"
+                logger.error(message)
+                raise ValidationError(_(message))
+        return queryset
```
ltshb, line 524: FOR NEXT PR: 
- [ ] Separate the parsing logic with the queryset logic. I mean first do the parsing with validation of the json in a method looping over all attributes, then in a second method loop again in the attributes to generate the queryset. Looping twice here is totally acceptable because we should not have big query string with thousand attributes and it would make the code cleaner and easier to understand. 
- [ ] In the parsing method then make sure to not allow duplicate attributes and to only have valid attributes (this would automatically prevent looping over a big structure twice :-))
- [ ] In the parsing method make uses of the multiple errors, instead of stopping at the first errors save the error in a list and at the end of the parsing return `ValidationError(error_list, code='query-invalid')`
- [ ] Due to the preceding we need to avoid to loop over a big numbers of attributes, otherwise a malicious could generate a huge query string for a DoS. So a first check that the `len(json_query) < (number_max_attribute + 5)` and directly stop the validation if this is not true. This way we can make uses of multiples error without the risk of a big loop.
- [ ] Make more explicit debug message, so that we can use them on Kibana (add a context). For example the request could be added as extra.

app/stac_api/views.py
```diff
@@ -344,6 +389,208 @@ def delete(self, request, *args, **kwargs):
...
+
+    def filter_by_intersects(self, queryset, intersects):
+        try:
+            logger.debug('Item query parameter intersects = %s', intersects)
+            the_geom = GEOSGeometry(intersects)
+        except ValueError as error:
+            message = f"Invalid intersects parameter: " \
+                f"Could not transform {intersects} to a geometry; {error}"
+            logger.error(message)
+            raise ValidationError(_(message))
```
ltshb, line 534: FOR NEXT PR:  make use of ValidationError code.

app/stac_api/views.py
```diff
@@ -344,6 +389,208 @@ def delete(self, request, *args, **kwargs):
...
+    def filter_by_ids(self, queryset, ids):
+        ids_array = ids.split(',')
+        queryset = queryset.filter(name__in=ids_array)
+        return queryset
+
+    def list(self, request, *args, **kwargs):
+        if request.method == 'POST':
+            queryset = self.filter_queryset(self.parse_request_body_for_queryset())
+        else:
+            queryset = self.filter_queryset(self.get_queryset())
```
ltshb, line 554: FOR NEXT PR: 
Proposition for better code comprehension and code reuse:
```python
if request.method == 'POST':
   query = self.parse_query_from_payload()
else:
   query = self.parse_query_from_param()

queryset = self.filter_queryset(self.get_queryset(query))
```
Both parse method would also do validation. `query` would be a tuple `(ids, collections, bbox, date_time, query, intersects)` with intersects being `None` in case of GET.

The `get_queryset()` method could be changed and reused: `get_queryset(self, bbox, date_time, collections, ids, query)`

app/tests/test_search_endpoint.py
```diff
@@ -0,0 +1,155 @@
...
+    def test_query(self):
+        # get match
+        title = "My item 1"
+        query = '{"title":{"eq":"%s"}}' % title
+        response = self.client.get(
+            f"/{API_BASE}/search"
+            f"?query={query}&limit=100")
+        self.assertStatusCode(200, response)
+        json_data_get = response.json()
+        self.assertEqual(json_data_get['features'][0]['properties']['title'], title)
```
ltshb, line 54: FOR NEXT PR: check that only one features is found.

app/tests/test_search_endpoint.py
```diff
@@ -0,0 +1,155 @@
...
+            }
+        }
+        """
+        response = self.client.post(self.path, data=payload, content_type="application/json")
+        self.assertStatusCode(200, response)
+        json_data_post = response.json()
+        self.assertEqual(json_data_post['features'][0]['properties']['title'], title)
+
+        # compare get and post
+        self.assertEqual(len(json_data_post['features']), len(json_data_get['features']))
```
ltshb, line 70: FOR NEXT PR: would check the payload equality and not the size. Both payload response should be equal instead of the timestamp.

app/tests/test_search_endpoint.py
```diff
@@ -0,0 +1,155 @@
...
+        {"query":
+            {"title":
+                {"eq":"My item 1"}
+            }
+        }
+        """
+        response = self.client.post(self.path, data=payload, content_type="application/json")
+        self.assertStatusCode(200, response)
+        json_data_post = response.json()
+        self.assertEqual(json_data_post['features'][0]['properties']['title'], title)
```
ltshb, line 67: FOR NEXT PR: same as above make sure that only one item is found.

app/tests/test_search_endpoint.py
```diff
@@ -0,0 +1,155 @@
...
+        payload = """
+        {"query":
+            {"datetime":
+                {"in":["2020-10-28T13:05:10Z","2525-10-19T00:00:00Z"]}
+            }
+        }
+        """
+        response = self.client.post(self.path, data=payload, content_type="application/json")
+        json_data = response.json()
+        self.assertEqual(len(json_data['features']), 2)
```
ltshb, line 83: FOR NEXT PR: make sure that the id of the items found are the expected ones

app/tests/test_search_endpoint.py
```diff
@@ -0,0 +1,155 @@
...
+            if link['rel'] == 'next':
+                cursor = link['href'].split('?')[1]
+        item_id1 = json_data['features'][0]['id']
+
+        # pagination
+        response = self.client.get(f"{self.path}?{cursor}")
+        self.assertStatusCode(200, response)
+        json_data = response.json()
+        item_id2 = json_data['features'][0]['id']
+        self.assertNotEqual(item_id1, item_id2)
```
ltshb, line 110: FOR NEXT PR: check the item ids and use a loop over the next and previous link, this can be copied from unittest added in PR #164

app/tests/test_search_endpoint.py
```diff
@@ -0,0 +1,155 @@
...
+        data = """
+        { "intersects":
+            { "type": "POINT",
+              "coordinates": [6, 47]
+            }
+        }
+        """
+        response = self.client.post(f"{self.path}", data=data, content_type="application/json")
+        json_data = response.json()
+        self.assertGreater(len(json_data['features']), 0)
```
ltshb, line 123: FOR NEXT PR: check the expected items ids.

app/tests/test_search_endpoint.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def test_collections(self):
+        # match
+        response = self.client.get(
+            f"/{API_BASE}/search"
+            f"?collections=collection-1,har")
+        self.assertStatusCode(200, response)
+        json_data = response.json()
+        self.assertGreater(len(json_data['features']), 0)
```
ltshb, line 132: FOR NEXT PR: test the expected items id.

app/tests/test_search_endpoint.py
```diff
@@ -0,0 +1,155 @@
...
+        json_data = response.json()
+        self.assertEqual(len(json_data['features']), 0)
+
+    def test_ids(self):
+        response = self.client.get(
+            f"/{API_BASE}/search"
+            f"?ids=item-1,item-2")
+        self.assertStatusCode(200, response)
+        json_data = response.json()
+        self.assertEqual(len(json_data['features']), 2)
```
ltshb, line 147: FOR NEXT PR: check the expected item ids

app/tests/test_search_endpoint.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def test_ids_first_and_only_prio(self):
+        response = self.client.get(
+            f"/{API_BASE}/search"
+            f"?ids=item-1,item-2&collections=not_exist")
+        self.assertStatusCode(200, response)
+        json_data = response.json()
+        self.assertEqual(len(json_data['features']), 2)
```
ltshb, line 155: FOR NEXT PR: test the expected item ids

ltshb (changes requested):
Some important issues, see comments.

I've already added comment for refactoring and improving that can be done in another PR, each comment is prefixed with `FOR NEXT PR: `.

app/stac_api/views.py
```diff
@@ -106,6 +112,92 @@ def parse_datetime_query(date_time):
     return start, end
 
 
+def filter_by_bbox(queryset, bbox):
+    '''Filter a querystring with a given bbox
+
+    This function is a helper function, for some views, to add a bbox filter to the queryset.
```
boecklic, line 118: FOR NEXT PR: following the object-oriented approach of django, this would be a candidate for creating custom queryset class: https://docs.djangoproject.com/en/3.1/topics/db/managers/#calling-custom-queryset-methods-from-the-manager

app/stac_api/views.py
```diff
@@ -106,6 +112,92 @@ def parse_datetime_query(date_time):
...
+                                ' has to contain 4 values. f.ex. bbox=5.96,45.82,10.49,47.81'))
+    query_bbox_polygon.srid = 4326
+    validate_geometry(query_bbox_polygon)
+    return queryset.filter(geometry__intersects=query_bbox_polygon)
+
+
+def filter_by_datetime(queryset, date_time):
+    '''Filter a queryset by datetime
+
+    Args:
```
boecklic, line 154: FOR NEXT PR: dito, custom queryset class

app/stac_api/views.py
```diff
@@ -106,6 +112,92 @@ def parse_datetime_query(date_time):
...
+    start, end = parse_datetime_query(date_time)
+    if end is not None:
+        return _filter_by_datetime_range(queryset, start, end)
+    return queryset.filter(properties_datetime=start)
+
+
+def _filter_by_datetime_range(queryset, start_datetime, end_datetime):
+    '''Filter a queryset by datetime range
+
+    Helper function of filter_by_datetime
```
boecklic, line 171: FOR NEXT PR: dito

boecklic (approved):
Nice! I have two general comments:

For this PR: the current example for POST /search is not very good (the POST payload). If you could improve that it would help inser I think.

For the next PR: as mentioned in the inline comments: we haven't much worked with customizing querysets and mangers yet. This is the occasion to do so, so we stay in the "spirit" of Django. So the filter functions should be methods of a custom QuerySet class.

ltshb (approved):

app/stac_api/views.py
rebert: https://github.com/geoadmin/service-stac/pull/175

app/stac_api/views.py
rebert: https://github.com/geoadmin/service-stac/pull/175

app/stac_api/views.py
rebert: https://github.com/geoadmin/service-stac/pull/181

rebert: Do not allow duplicate attributes: They will be treated with AND... so I will allow 50 times the same a attribte

PR #153, opened 2021-01-04
BGDIINF_SB-1551: added docu on authentication in specs

hansmannj:


spec/base.yml
```diff
@@ -23,3 +23,14 @@ tags:
     name: Data
   - description: Extension to OGC API - Features to support STAC metadata model and search API
     name: STAC
+  - description: >-
+      All write requests require authentication. There are several options for a user to
+      authenticate himself:
+
+      * When using the browsable API the user can simply use the admin interface for logging in
+
+      * Using basic authentication for the write requests.
```
hansmannj, line 32: Would it make sense to add command line examples (e.g. using curl et al.) here?

spec/base.yml
```diff
@@ -23,3 +23,14 @@ tags:
...
+
+      * Using token authentication for the write requests. See description of the get-token POST endpoint
```
hansmannj, line 34: Command line examples here as well?

spec/base.yml
boecklic, line 32: Pls add this to `transaction/transaction.yml`, not `base.yml` since it only concerns the transactional part.

Note that you have to repeat all tags, so you need  to add (uncomment):
```
tags:
  - description: Essential characteristics of this API
    name: Capabilities
  - description: Access to data (features)
    name: Data
  - description: Extension to OGC API - Features to support STAC metadata model and search API
    name: STAC
  - name: 'Data Management'
    description: >-
      STAC-specific operations to add, remove, and edit items within OGC API - Features
      collections.
```

Yeah, curl examples would be nice, either here or in one of the Markdown files.

boecklic (changes requested):

spec/base.yml
hansmannj: added examples

spec/base.yml
hansmannj: added examples

spec/base.yml
hansmannj: Makes sense, moved it

spec/transaction/transaction.yml
```diff
@@ -893,8 +945,66 @@ components:
             description: "Permission denied"
 # Note: this is defined as an overwrite in 'overwrites', since defining it here would violate
```
hansmannj, line 946: This comment might be obsolete, right? Should I delete it?

boecklic: Jup

boecklic (approved):
:+1:

PR #156, opened 2021-01-07
#patch prod logging settings

boecklic:
logging settings changed for prod: DEBUG logs from app should be printed as well, WARNING for all other loggers

Note: this is done as a `hotfix` (branched off `master`) to check if hotfix versioning (patch position in semver) is working correctly. A hotfix branch has to be merged back in both, `develop` and `master`

### DIFF
--- a/app/config/logging-cfg-prod.yml
+++ b/app/config/logging-cfg-prod.yml
@@ -4,10 +4,14 @@ disable_existing_loggers: False # this allow to get logger at module level
 root:
   handlers:
     - console
-  level: INFO
+  level: WARNING
   propagate: True
 
 loggers:
+  stac_api:
+    level: DEBUG
+  middleware:
+    level: DEBUG
   django:
     level: INFO
   gunicorn.error:


app/config/logging-cfg-prod.yml
```diff
@@ -4,10 +4,14 @@ disable_existing_loggers: False # this allow to get logger at module level
...
+  level: WARNING
   propagate: True
 
 loggers:
+  stac_api:
+    level: DEBUG
+  middleware:
+    level: DEBUG
   django:
     level: INFO
```
ltshb, line 16: You might want to remove these lines, otherwise django will use INFO level instead of WARNING.

ltshb (approved):
See inline comment about django logging level.

app/config/logging-cfg-prod.yml
boecklic: Yeah, thought about that. Not sure if these messages will be useful or not. I don't think that we'll use them... but we can leave it for the moment and remove once we see that things are stable and we never use them.

procrastinatio (approved):

PR #157, opened 2021-01-07
Hotfix change prod logging

boecklic:
logging settings changed for prod: DEBUG logs from app should be printed as well, WARNING for all other loggers


app/config/logging-cfg-prod.yml
```diff
@@ -4,10 +4,14 @@ disable_existing_loggers: False # this allow to get logger at module level
...
+  level: WARNING
   propagate: True
 
 loggers:
+  stac_api:
+    level: DEBUG
+  middleware:
+    level: DEBUG
   django:
     level: INFO
```
ltshb, line 16: Same as for #156

ltshb (approved):
Same as for #156

procrastinatio (approved):
> Do not forget to revert when really in prod

I wouldn't change this for prod. Debug log messages from libraries such as boto3 are imho of little use for debugging, important are log messages from the app itself (the ones from us)

PR #160, opened 2021-01-08
BGDIINF_SB-1534: avoid duplicate validation in items

ltshb:
The item properties datetimes were validated twice in a row. This was
unecessary and due to a wrong uses of the django/rest_framework in
unittest. Creating an object with
ModelClass.objects.create(**validated_data) should
only be done with validated data, like for example in the serializer.
This method doesn't do any validation. If we don't have validated data,
then we need to use ModelClass(**unvalidated_data).full_clean().save()

So the item model unittest has been corrected in order to remove the
self.clean() in the item.save() method. This introduce a sligthly
performance improvement, but more important it makes the code cleaner.

### DIFF
--- a/app/stac_api/serializers.py
+++ b/app/stac_api/serializers.py
@@ -735,12 +735,21 @@ class ItemSerializer(NonNullModelSerializer):
         return super().update(instance, validated_data)
 
     def validate(self, attrs):
-        validate_item_properties_datetimes(
-            attrs.get('properties_datetime', None),
-            attrs.get('properties_start_datetime', None),
-            attrs.get('properties_end_datetime', None),
-            partial=self.partial
-        )
+        if (
+            not self.partial or \
+            'properties_datetime' in attrs or \
+            'properties_start_datetime' in attrs or \
+            'properties_end_datetime' in attrs
+        ):
+            validate_item_properties_datetimes(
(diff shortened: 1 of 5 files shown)


hansmannj (approved):
Looks good to me :+1:

boecklic (approved):

app/stac_api/serializers.py
```diff
@@ -735,12 +735,21 @@ def update(self, instance, validated_data):
...
-            partial=self.partial
-        )
+        if (
+            not self.partial or \
+            'properties_datetime' in attrs or \
+            'properties_start_datetime' in attrs or \
+            'properties_end_datetime' in attrs
+        ):
+            validate_item_properties_datetimes(
+                attrs.get('properties_datetime', None),
```
procrastinatio, line 745: So `None`is a valid date?

procrastinatio (commented):
Would also test if `date`is `None`

app/stac_api/serializers.py
ltshb: None is not a valide date, but in properties there is 3 dates:
- properties_datetime
- properties_start_datetime
- properties_end_datetime

and either the first one is required or both start and end datetime, that's what validate the function.

PR #161, opened 2021-01-11
BGDIINF_SB-1568: Fixed serializer hide_fields and added asset mgmt tests

ltshb:
The serializer hide_fields did not work for the asset put/patch
endpoints because the parameter was not passed in the view mixins. This
has been fixed.

In data_factory, allow to use extra attribute with the
required_only=True. This way we can retrieve the minimal attributes plus
an extra.

Completed the asset endpoint unittest to test the asset management. What
still missing is testing the asset managment on S3 (uploading/moving/deleting
asset) because S3 is currently mocked in unittest.


rebert (approved):
lgtm

PR #162, opened 2021-01-12
Feature BGDIINF_SB-1545 query set caching

hansmannj:
* made use of QuerySet caching in collection_temporal_extent.py
* optimized performance by using exists() insteadof bool() in collection_summaries.py
* optimized performance by re-using already evaluated quersets in collection_temporal_extent.py
  (The readability of code has suffered a bit under this, but I documented and commented, so I hope, it still is OK)

### DIFF
--- a/app/stac_api/collection_temporal_extent.py
+++ b/app/stac_api/collection_temporal_extent.py
@@ -84,8 +84,14 @@ def update_start_temporal_extent_on_item_update(
 
     Returns:
         bool: True if temporal extent has been updated, false otherwise
+        return_qs: QuerySet | None
+            Queryset containing all items (but the one currently updated) that have
+            non-null properties_datetime values. This queryset might be used in
+            update_end_temporal_extent_on_item_update() and can be passed in already
+            evaluated state to save one DB hit.
     '''
     updated = False
+    return_qs = None
     logger.debug(
         "Updating start of item %s in collection %s (old start: %s, new start: %s, "
         "old end: %s, new end: %s) ",
@@ -114,22 +120,41 @@ def update_start_temporal_extent_on_item_update(
             # set earliest start_datetime to min(earliest_start_datetime
             # of all items but the one currently updated and
             # new_start_datetime).
-            if bool(qs.filter(properties_start_datetime__isnull=False)):
+            logger.warning(
+                'Looping over all items of collection %s,'
+                'to update temporal extent, this may take a while',
+                collection.name
+            )
+            qs_other_items_with_properties_start_datetime = qs.filter(
+                properties_start_datetime__isnull=False
+            )
+            if bool(qs_other_items_with_properties_start_datetime):
+                other_items_earliest_properties_start_datetime = (
+                    qs_other_items_with_properties_start_datetime.
+                    earliest('properties_start_datetime').properties_start_datetime
+                )
                 earliest_start_datetime = min(
-                    new_start_datetime,
-                    qs.filter(properties_start_datetime__isnull=False
-                             ).earliest('properties_start_datetime').properties_start_datetime
+                    new_start_datetime, other_items_earliest_properties_start_datetime
                 )
             else:
                 earliest_start_datetime = new_start_datetime
             # set earliest datetime to min(earliest_datetime of all items
-            # bute the one currently updated and new_start_datetime)
-            if bool(qs.filter(properties_datetime__isnull=False)):
-                earliest_datetime = min(
-                    new_start_datetime,
-                    qs.filter(properties_datetime__isnull=False
-                             ).earliest('properties_datetime').properties_datetime
+            # but the one currently updated and new_start_datetime)
+            logger.warning(
+                'Looping over all items of collection %s,'
+                'to update temporal extent, this may take a while',
+                collection.name
+            )
+            qs_other_items_with_properties_datetime = qs.filter(properties_datetime__isnull=False)
+            if bool(qs_other_items_with_properties_datetime):
+                other_items_earliest_properties_datetime = (
+                    qs_other_items_with_properties_datetime.earliest('properties_datetime'
+                                                                    ).properties_datetime
                 )
+                earliest_datetime = min(
+                    new_start_datetime, other_items_earliest_properties_datetime
+                )
+                return_qs = qs_other_items_with_properties_datetime
             else:
                 earliest_datetime = new_start_datetime
             updated |= True
@@ -142,7 +167,7 @@ def update_start_temporal_extent_on_item_update(
         updated |= True
         collection.extent_start_datetime = new_start_datetime
 
-    return updated
+    return updated, return_qs
 
 
 def update_end_temporal_extent_on_item_update(
@@ -152,7 +177,8 @@ def update_end_temporal_extent_on_item_update(
     old_end_datetime,
     new_end_datetime,
     item_id,
-    qs=None
+    qs=None,
+    qs_other_items_with_properties_datetime=None
 ):
     '''This function is called from within update_temporal_extent() when an
     item in the collection is updated to check if the collection's
@@ -173,6 +199,9 @@ def update_end_temporal_extent_on_item_update(
             the id of the item being treated (pk)
         qs: QuerySet | None
             queryset with all items of the collection excluding the one being updated. (optional)
+        qs_other_items_with_properties_datetimes: QuerySet | None
+            Already evaluated queryset with all items (but the one currently updated) that have
+            non-null properties_datetime values (optional).
 
     Returns:
         bool: True if temporal extent has been updated, false otherwise
@@ -205,22 +234,42 @@ def update_end_temporal_extent_on_item_update(
             # to be determined:
             # set latest end_datetime to max(new_end_datetime and
             # end_datetime of all items but the one currently updated).
-            if bool(qs.filter(properties_end_datetime__isnull=False)):
-                latest_end_datetime = max(
-                    new_end_datetime,
-                    qs.filter(properties_end_datetime__isnull=False
-                             ).latest('properties_end_datetime').properties_end_datetime
+            logger.warning(
+                'Looping over all items of collection %s,'
+                'to update temporal extent, this may take a while',
+                collection.name
+            )
+            qs_other_items_with_properties_end_datetime = qs.filter(
+                properties_end_datetime__isnull=False
+            )
+            if bool(qs_other_items_with_properties_end_datetime):
+                other_items_latest_end_datetime = (
+                    qs_other_items_with_properties_end_datetime.latest('properties_end_datetime'
+                                                                      ).properties_end_datetime
                 )
+                latest_end_datetime = max(new_end_datetime, other_items_latest_end_datetime)
             else:
                 latest_end_datetime = new_end_datetime
             # set latest datetime to max(new_end_datetime and
             # end end_datetime of all items but the one currently updated)
-            if bool(qs.filter(properties_datetime__isnull=False)):
-                latest_datetime = max(
-                    new_end_datetime,
-                    qs.filter(properties_datetime__isnull=False
-                             ).latest('properties_datetime').properties_datetime
+            logger.warning(
+                'Looping over all items of collection %s,'
+                'to update temporal extent, this may take a while',
+                collection.name
+            )
+            # get latest datetime, or none in case none exists and
+            # use the already evaluated qs_other_items_with_properties_datetime, if it has
+            # been passed to this function to save a DB hit.
+            if qs_other_items_with_properties_datetime is None:
+                qs_other_items_with_properties_datetime = qs.filter(
+                    properties_datetime__isnull=False
                 )
+            if bool(qs_other_items_with_properties_datetime):
+                other_items_latest_properties_datetime = (
+                    qs_other_items_with_properties_datetime.latest('properties_datetime'
+                                                                  ).properties_datetime
+                )
+                latest_datetime = max(new_end_datetime, other_items_latest_properties_datetime)
             else:
                 latest_datetime = new_end_datetime
             updated |= True
@@ -254,8 +303,14 @@ def update_start_temporal_extent_on_item_delete(collection, old_start_datetime, 
 
     Returns:
         bool: True if temporal extent has been updated, false otherwise
+        return_qs: QuerySet | None
+            Queryset containing all items (but the one currently updated) that have
+            non-null properties_datetime values. This queryset might be used in
+            update_end_temporal_extent_on_item_update() and can be passed in already
+            evaluated state to save one DB hit.
     '''
     updated = False
+    return_qs = None
     logger.debug(
         "Deleting item %s from collection %s and updating the collection's start date",
         item_id,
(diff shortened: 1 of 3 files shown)


app/stac_api/collection_temporal_extent.py
```diff
@@ -114,22 +120,41 @@ def update_start_temporal_extent_on_item_update(
...
-            if bool(qs.filter(properties_start_datetime__isnull=False)):
+            logger.warning(
+                'Looping over all items of collection %s,'
+                'to update temporal extent, this may take a while',
+                collection.name
+            )
+            qs_other_items_with_properties_start_datetime = qs.filter(
+                properties_start_datetime__isnull=False
+            )
+            if bool(qs_other_items_with_properties_start_datetime):
```
ltshb, line 131: Uses `exists()` instead of bool(). Because if it exists we reconstruct a new query, using exists might have a performance impact here.

app/stac_api/collection_temporal_extent.py
```diff
@@ -114,22 +120,41 @@ def update_start_temporal_extent_on_item_update(
...
-            # bute the one currently updated and new_start_datetime)
-            if bool(qs.filter(properties_datetime__isnull=False)):
+            # but the one currently updated and new_start_datetime)
+            logger.warning(
+                'Looping over all items of collection %s,'
+                'to update temporal extent, this may take a while',
+                collection.name
+            )
+            qs_other_items_with_properties_datetime = qs.filter(properties_datetime__isnull=False)
+            if bool(qs_other_items_with_properties_datetime):
```
ltshb, line 149: Same as above use `exists()`

app/stac_api/collection_temporal_extent.py
```diff
@@ -173,6 +199,9 @@ def update_end_temporal_extent_on_item_update(
             the id of the item being treated (pk)
         qs: QuerySet | None
             queryset with all items of the collection excluding the one being updated. (optional)
+        qs_other_items_with_properties_datetimes: QuerySet | None
+            Already evaluated queryset with all items (but the one currently updated) that have
```
ltshb, line 203: Is not necessary already evaluated if used with exists() see comment above

app/stac_api/collection_temporal_extent.py
```diff
@@ -173,6 +199,9 @@ def update_end_temporal_extent_on_item_update(
             the id of the item being treated (pk)
         qs: QuerySet | None
```
ltshb, line 200: maybe renaming `qs` with `qs_items` or `qs_other_items` would make the code clearer

app/stac_api/collection_temporal_extent.py
```diff
@@ -205,22 +234,42 @@ def update_end_temporal_extent_on_item_update(
...
-                             ).latest('properties_end_datetime').properties_end_datetime
+            logger.warning(
+                'Looping over all items of collection %s,'
+                'to update temporal extent, this may take a while',
+                collection.name
+            )
+            qs_other_items_with_properties_end_datetime = qs.filter(
+                properties_end_datetime__isnull=False
+            )
+            if bool(qs_other_items_with_properties_end_datetime):
```
ltshb, line 245: same, use `exists()`

app/stac_api/collection_temporal_extent.py
```diff
@@ -205,22 +234,42 @@ def update_end_temporal_extent_on_item_update(
...
+                collection.name
+            )
+            # get latest datetime, or none in case none exists and
+            # use the already evaluated qs_other_items_with_properties_datetime, if it has
+            # been passed to this function to save a DB hit.
+            if qs_other_items_with_properties_datetime is None:
+                qs_other_items_with_properties_datetime = qs.filter(
+                    properties_datetime__isnull=False
                 )
+            if bool(qs_other_items_with_properties_datetime):
```
ltshb, line 267: same uses exists()

app/stac_api/collection_temporal_extent.py
```diff
@@ -270,25 +325,34 @@ def update_start_temporal_extent_on_item_delete(collection, old_start_datetime,
...
-
         # get earliest start_datetime or none, in case none exists
-        if bool(qs.filter(properties_start_datetime__isnull=False)):
-            earliest_start_datetime = qs.filter(
-                properties_start_datetime__isnull=False
-            ).earliest('properties_start_datetime').properties_start_datetime
+        qs_other_items_with_properties_start_datetime = qs.filter(
+            properties_start_datetime__isnull=False
+        )
+        if bool(qs_other_items_with_properties_start_datetime):
```
ltshb, line 337: Uses exists()

app/stac_api/collection_temporal_extent.py
```diff
@@ -270,25 +325,34 @@ def update_start_temporal_extent_on_item_delete(collection, old_start_datetime,
...
+            'Looping over all items of collection %s,'
+            'to update temporal extent, this may take a while',
+            collection.name
+        )
         # get earliest datetime, or none in case none exists
-        if bool(qs.filter(properties_datetime__isnull=False)):
-            earliest_datetime = qs.filter(properties_datetime__isnull=False
-                                         ).earliest('properties_datetime').properties_datetime
+        qs_other_items_with_properties_datetime = qs.filter(properties_datetime__isnull=False)
+        if bool(qs_other_items_with_properties_datetime):
```
ltshb, line 351: Uses exists()

app/stac_api/collection_temporal_extent.py
```diff
@@ -318,6 +384,9 @@ def update_end_temporal_extent_on_item_delete(collection, old_end_datetime, item
             the id of the item being treated (pk)
         qs: QuerySet | None
             queryset with all items of the collection excluding the one being updated. (optional)
+        qs_other_items_with_properties_datetimes: QuerySet | None
+            Already evaluated queryset with all items (but the one currently updated) that have
```
ltshb, line 388: See comment about exists(), then it is not fully evaluated.

app/stac_api/collection_temporal_extent.py
```diff
@@ -339,22 +408,35 @@ def update_end_temporal_extent_on_item_delete(collection, old_end_datetime, item
...
         )
         # get latest end_datetime or none, in case none exists
-        if bool(qs.filter(properties_end_datetime__isnull=False)):
-            latest_end_datetime = qs.filter(
-                properties_end_datetime__isnull=False
-            ).latest('properties_end_datetime').properties_end_datetime
+        qs_other_items_with_properties_end_datetime = qs.filter(
+            properties_end_datetime__isnull=False
+        )
+        if bool(qs_other_items_with_properties_end_datetime):
```
ltshb, line 419: Uses exists()

app/stac_api/collection_temporal_extent.py
```diff
@@ -339,22 +408,35 @@ def update_end_temporal_extent_on_item_delete(collection, old_end_datetime, item
...
+        # use the already evaluated qs_other_items_with_properties_datetime, if it has
+        # been passed to this function to save a DB hit.
+        logger.warning(
+            'Looping over all items of collection %s,'
+            'to update temporal extent, this may take a while',
+            collection.name
+        )
+        if qs_other_items_with_properties_datetime is None:
+            qs_other_items_with_properties_datetime = qs.filter(properties_datetime__isnull=False)
+        if bool(qs_other_items_with_properties_datetime):
```
ltshb, line 436: Use exists()

app/stac_api/collection_temporal_extent.py
```diff
@@ -438,33 +517,29 @@ def update_temporal_extent(
                     item.pk,
                     qs
                 )
+                )
```
ltshb, line 520: Code style, indentation issue

app/stac_api/collection_temporal_extent.py
```diff
@@ -438,33 +517,29 @@ def update_temporal_extent(
...
                     collection,
                     old_start_datetime,
                     new_start_datetime,
                     old_end_datetime,
                     new_end_datetime,
-                    item.pk
+                    item.pk,
+                    qs=None
+                )
                 )
```
ltshb, line 534: Code style indentation

ltshb (changes requested):
Some missing replacement of `bool()` to exists()

app/stac_api/collection_temporal_extent.py
hansmannj: `bool()` evaluates a qs, so that when the qs is accessed the next time, the cache is used.
`exist()` does not evaluate the qs. So I used `bool()` instead of `exist()` in all those places, where the qs is accessed after calling `bool()`. In other cases I used `exist()`

ltshb: I don't know the detail of the implementation of the queryset, but if we would later loop over the qs, then bool() would be for sure correct. However when we use the qs to do a new qs, which is here the case with the `.earliest()`, django might improve performance with using another query and we don't need to evaluate the complete initial query.

hansmannj: My understanding is, that `.earliest()` accesses the already existing qs. If using `.filter()` then I agree, `exists()` would be better here. But I thought, using `.earliest()` is just a kind of sorting and returning the first hit and not creating a new qs.
Since this in an important point I'll have to double check here. This would impact a lot of `exists()` and `bools()` of this PR.

ltshb: Mmm based on the docu, I don't think that `earliest()` and `latest()` evaluates the whole queryset like `bool()` does: https://docs.djangoproject.com/en/3.1/ref/models/querysets/#when-querysets-are-evaluated

hansmannj: This is what I thought initially as well.
But checking the docu for `filter()` it is explicitly stated, that a new QuerySet is returned:
https://docs.djangoproject.com/en/3.1/ref/models/querysets/#filter
--> "Returns a new QuerySet containing objects..."

Which is not true for the docu on `earliest()` or `latest()` respectively:
https://docs.djangoproject.com/en/3.1/ref/models/querysets/#django.db.models.query.QuerySet.latest
--> "Returns the latest object in the table..."*

This is why I thought using `bool()` first, which evaluates the qs, followed by an `earliest()` on this already evaluated qs is more performant than using `exists()` instead of `bool()` in this cases.
In all other cases I sticked to using `exists()`. I'll have to dig deeper here, but I did not find any more useful docu so far.

ltshb: Maybe this would be better : 
```python
try:
    earliest = qs.filter(properties_start_datetime__isnull=False).earliest('properties_start_datetime')
except DoesNotExists as error:
   ...
```
Not 100% sure if this impact performance, but here we only retrieve one object, while with `bool(qs.filter(properties_start_datetime__isnull=False))` you potentially evaluates a big queryset. @boecklic  what is your opinion based on your experience with django ?

hansmannj: Funny, this is pretty much what I first had in mind, too.
But again, I guess the first part `qs.filter(properties_start_datetime__isnull=False)` is the expensive part here, which will be evaluated anyway and this is the part where there can be many many hits. And afterwards the `earliest()` simply sorts and returns one of the many. So I guess this will not significantly improve performance.

hansmannj: Perhaps it might make sense to create a test db with a very large number of sample items, so I can check the costs of these queries..

app/stac_api/collection_temporal_extent.py
hansmannj: I am not sure if I fully understand your point here.

My reasoning for passing the `qs_other_items_with_properties_datetime` as an optional argument to `update_end_temporal_extent_on_item_update()` is, that probably this qs has already been evaluated earlier on in `update_start_temporal_extent_on_item_update`.

So `update_start_temporal_extent_on_item_update` returns either this very qs or None. If it has returned the qs, then this qs is already evaluated and doesn't need to be evaluated again in `update_end_temporal_extent_on_item_update`.
So we can save on DB hit here, for items, where both, starts and ends have been udated.

Similar implementation is done for updating on item_delete instead of item_update further below.

app/stac_api/collection_temporal_extent.py
hansmannj: good point, I'll change the naming

app/stac_api/collection_temporal_extent.py
hansmannj: See comment above when updating on item_update

app/stac_api/collection_temporal_extent.py
hansmannj: I'll change that. Strange that linting didn't complain

app/stac_api/collection_temporal_extent.py
hansmannj: dito

app/stac_api/collection_temporal_extent.py
hansmannj: OK, so now there is  a ticket for possible improvements/refactoring of `collection_temporal.extent.py` (https://jira.swisstopo.ch/browse/BGDIINF_SB-1582).

As for the point of using `bool()` or `exists()` here, I did some test on a collection with 150'000 items:

**exists()**
```
qs_exists = Item.objects.filter(collection_id="3").exclude(id="0")
qs_exists.exists()
```
this yields: 
```(0.001) SELECT (1) AS "a" FROM "stac_api_item" WHERE ("stac_api_item"."collection_id" = 3 AND NOT ("stac_api_item"."id" = 0)) LIMIT 1; args=(3, 0)
```
fetching the earliest datetime with `qs_exists.earliest('properties_datetime').properties_datetime` yields:
```
(0.170) SELECT "stac_api_item"."id", "stac_api_item"."name", "stac_api_item"."collection_id", "stac_api_item"."geometry"::bytea, "stac_api_item"."created", "stac_api_item"."updated", "stac_api_item"."properties_datetime", "stac_api_item"."properties_start_datetime", "stac_api_item"."properties_end_datetime", "stac_api_item"."properties_title", "stac_api_item"."etag", "stac_api_collection"."id", "stac_api_collection"."name", "stac_api_collection"."created", "stac_api_collection"."updated", "stac_api_collection"."description", "stac_api_collection"."extent_geometry"::bytea, "stac_api_collection"."extent_start_datetime", "stac_api_collection"."extent_end_datetime", "stac_api_collection"."license", "stac_api_collection"."summaries", "stac_api_collection"."title", "stac_api_collection"."etag" FROM "stac_api_item" INNER JOIN "stac_api_collection" ON ("stac_api_item"."collection_id" = "stac_api_collection"."id") WHERE ("stac_api_item"."collection_id" = 3 AND NOT ("stac_api_item"."id" = 0)) ORDER BY "stac_api_item"."properties_datetime" ASC LIMIT 1; args=(3, 0)
True
```
During this fetch with `earliest()` the qs is evaluated and the time needed is 0.170 ms.
After evaluation, the cached qs is used. Average time per fetching the `earliest()` after initial evaluation is 0.124 (average of 10 fetches).


**bool()**
```
qs_bool = Item.objects.filter(collection_id="3").exclude(id="0")
bool(qs_bool)
```
yields:
(0.931) SELECT "stac_api_item"."id", "stac_api_item"."name", "stac_api_item"."collection_id", "stac_api_item"."geometry"::bytea, "stac_api_item"."created", "stac_api_item"."updated", "stac_api_item"."properties_datetime", "stac_api_item"."properties_start_datetime", "stac_api_item"."properties_end_datetime", "stac_api_item"."properties_title", "stac_api_item"."etag", "stac_api_collection"."id", "stac_api_collection"."name", "stac_api_collection"."created", "stac_api_collection"."updated", "stac_api_collection"."description", "stac_api_collection"."extent_geometry"::bytea, "stac_api_collection"."extent_start_datetime", "stac_api_collection"."extent_end_datetime", "stac_api_collection"."license", "stac_api_collection"."summaries", "stac_api_collection"."title", "stac_api_collection"."etag" FROM "stac_api_item" INNER JOIN "stac_api_collection" ON ("stac_api_item"."collection_id" = "stac_api_collection"."id") WHERE ("stac_api_item"."collection_id" = 3 AND NOT ("stac_api_item"."id" = 0)); args=(3, 0)
True

The check for existence takes 0.931 ms, as the qs is fully evaluated at the same time, compared to 0.170ms when using `exists()` (where no evaluation is done).

Fetching earliest datetime using `earliest()` on the already evaluated qs with qs_bool.earliest('properties_datetime').properties_datetime` takes 0.118 ms.
Average of 10 subsequent executions: 0.13 ms.

**Executive summary ;-)**
So this is purely empirical guess-work, but `exist()` seems to be a bit faster in this setting, as the initial evaluation is quicker in comparison to `bool()`. So I'll use `exist()` here. And when we run into performance or other issues in future, I guess we'll do the refactoring and probably have to think about asynchronous jobs then.

ltshb (approved):
:+1:

PR #163, opened 2021-01-12
BGDIINF_SB-1546: Improved error messages

ltshb:
Improved user error message in validator.

Also removed duplicate key in unittest error message (the key is already
part of the path)


hansmannj (approved):
:+1:

PR #164, opened 2021-01-12
BGDIINF_SB-1575: Fixed items datetime query pagination

ltshb:
When using the next link from a items datetime query result, the API
returned a 500: "Calling QuerySet.filter() after union() is not supported."

So instead of using union() I use Q(). This is now working (see previous
unittest commit)


hansmannj (approved):
Looks good to me :+1:

PR #165, opened 2021-01-12
BGDIINF_SB-1564 perf test dummy data, prefetch_related

boecklic:
make use of `prefetch_related` in list and detail views to reduce the number of queries that are made to fetch related objects

added a mgmt command to generate diverse (i.e. with all expected values) dummy data for performance testing

### DIFF
--- a/app/config/settings_dev.py
+++ b/app/config/settings_dev.py
@@ -20,6 +20,8 @@ DEBUG = bool(strtobool(os.getenv('DEBUG', 'False')))
 
 ALLOWED_HOSTS = ['*']
 
+SECRET_KEY = 'fJZUSfuyvLGR5roQklcqBrvc/ceFPk8dcePzZIVqOKc='
+
 # django-extensions
 # ------------------------------------------------------------------------------
 if DEBUG:
--- a/app/config/settings_dev.py
+++ b/app/config/settings_dev.py
@@ -20,8 +20,6 @@ DEBUG = bool(strtobool(os.getenv('DEBUG', 'False')))
 
 ALLOWED_HOSTS = ['*']
 
-SECRET_KEY = 'fJZUSfuyvLGR5roQklcqBrvc/ceFPk8dcePzZIVqOKc='
-
 # django-extensions
 # ------------------------------------------------------------------------------
 if DEBUG:
--- a/app/config/logging-cfg-local.yml
+++ b/app/config/logging-cfg-local.yml
@@ -12,6 +12,8 @@ loggers:
     level: DEBUG
   middleware:
     level: DEBUG
+  # django.db:
+  #   level: DEBUG
   django:
     level: INFO
   django.utils.autoreload:
--- a/app/stac_api/admin.py
+++ b/app/stac_api/admin.py
@@ -85,6 +85,7 @@ class CollectionAdmin(admin.ModelAdmin):
         'extent_start_datetime', 'extent_end_datetime', 'summaries', 'extent_geometry'
     ]
     inlines = [ProviderInline, CollectionLinkInline]
+    search_fields = ['name']
 
 
 class ItemLinkInline(admin.TabularInline):
@@ -95,6 +96,8 @@ class ItemLinkInline(admin.TabularInline):
 @admin.register(Item)
 class ItemAdmin(admin.GeoModelAdmin):
     inlines = [ItemLinkInline]
+    autocomplete_fields = ['collection']
+    search_fields = ['name']
     fieldsets = (
         (None, {
             'fields': ('name', 'collection', 'geometry')
--- a/app/stac_api/management/commands/dummydata.py
+++ b/app/stac_api/management/commands/dummydata.py
@@ -1,0 +1,132 @@
+import datetime
+import logging
+import random
+import string
+
+from dateutil.parser import isoparse
+
+from django.contrib.gis.geos import Polygon
+from django.core.files.uploadedfile import SimpleUploadedFile
+from django.core.management.base import BaseCommand
+
+from stac_api.models import Asset
+from stac_api.models import Collection
+from stac_api.models import Item
+from stac_api.validators import MEDIA_TYPES
+
+logger = logging.getLogger(__name__)
+
+# Min/Max extent (roughly) of CH in LV95
+XMIN = 2570000
+XMAX = 2746000
+YMIN = 1130000
+YMAX = 1252000
+
+MIN_DATETIME = isoparse('1970-01-01T00:00:01Z')
+MAX_DATETIME = isoparse('2020-12-31T23:59:10Z')
+
+
+def random_datetime(start, end):
+    """Generate a random datetime between `start` and `end`"""
+    return start + datetime.timedelta(
+        # Get a random amount of seconds between `start` and `end`
+        seconds=random.randint(0, int((end - start).total_seconds())),
+    )
+
+
+class Command(BaseCommand):
+    help = """Manage dummy data for performance testing.
+
+    The command populates the database by default with
+    30 collections, 300 items per collection and 2 assets per item.
+    Number of collections, items and assets can be changed.
+
+    The generated data is randomized where necessary, i.e. the field
+    that are also likely to be queried.
+    """
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            'action',
+            type=str,
+            choices=['populate', 'clean'],
+            default='populate',
+            help=
+            'Define the action to be performed, either "populate" (default) to create '
+            'dummy data or "clean" to delete it',
+        )
+
+        parser.add_argument(
+            '--collections',
+            type=int,
+            default=30,
+            help="Number of collections to create (default 30)"
+        )
+
+        parser.add_argument(
+            '--items',
+            type=int,
+            default=300,
+            help="Number of items per collection to create (default 300)"
+        )
+
+        parser.add_argument(
+            '--assets', type=int, default=2, help="Number of assets per item to create (default 2)"
+        )
+
+    def handle(self, *args, **options):
+
+        if options['action'] == 'clean':
+            Collection.objects.filter(name__startswith='perftest').delete()
+        elif options['action'] == 'populate':
+
+            for collection_id in [
+                f'perftest-collection-{x}' for x in range(options['collections'])
+            ]:
+                collection, _ = Collection.objects.get_or_create(
+                    name=collection_id,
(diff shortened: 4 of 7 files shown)


app/config/settings_dev.py
```diff
@@ -20,6 +20,8 @@
 
 ALLOWED_HOSTS = ['*']
 
+SECRET_KEY = 'fJZUSfuyvLGR5roQklcqBrvc/ceFPk8dcePzZIVqOKc='
```
ltshb, line 23: Why overriding the SECRET_KEY here ? It should already be set in `.env.local` from `.env.default` or in `docker-compose-ci.yml`.

app/stac_api/management/commands/perftest.py
```diff
@@ -0,0 +1,118 @@
...
+
+            for collection_id in [f'perftest-collection-{x}' for x in range(30)]:
+                collection, _ = Collection.objects.get_or_create(
+                    name=collection_id,
+                    defaults={
+                        'description': 'This is a description',
+                        'license': 'test',
+                        'summaries': {
+                            "eo:gsd": [], "geoadmin:variant": [], "proj:epsg": []
+                        },
```
ltshb, line 78: Why adding summaries here ? Summaries should be read-only and auto-generated.

app/stac_api/management/commands/perftest.py
```diff
@@ -0,0 +1,118 @@
...
+                        for asset_id in [f'perftest-asset-{x}' for x in range(3)]:
+                            asset, _ = Asset.objects.get_or_create(
+                                item=item,
+                                name=asset_id,
+                                defaults={
+                                    'title': 'my-title',
+                                    'checksum_multihash': "01205c3fd6978a7d0b051efaa4263a09",
+                                    'description': "this an asset",
+                                    'eo_gsd': random.choice([2, 2.5, 5, 10]),
+                                    'geoadmin_lang': random.choice(['de', 'fr', 'it', 'rm', 'en', '']),
```
ltshb, line 108: empty language should not be allowed, see https://jira.swisstopo.ch/browse/BGDIINF_SB-1576, either don't generate asset without language or for some asset simply generates the asset without `geoadmin_lang` attribute

app/stac_api/management/commands/perftest.py
```diff
@@ -0,0 +1,118 @@
...
+            help='Clean dummy data',
+        )
+
+    def handle(self, *args, **options):
+
+        geo = Polygon.from_bbox((2570000.0, 1129970.0, 2580000.0, 1139990.0))
+        geo.srid = 2056
+        geo.transform(4326, clone=False)
+        range(2570000, 2746000, 1000)
+        range(1130000, 1252000, 1000)
```
ltshb, line 64: What is the purpose of this code ?

app/stac_api/management/commands/perftest.py
```diff
@@ -0,0 +1,118 @@
...
+def random_datetime(start, end):
+    """Generate a random datetime between `start` and `end`"""
+    return start + datetime.timedelta(
+        # Get a random amount of seconds between `start` and `end`
+        seconds=random.randint(0, int((end - start).total_seconds())),
+    )
+
+
+class Command(BaseCommand):
+    help = """Populates the local test database with dummy data
```
ltshb, line 41: Maybe a more explicit help here and how to use it. For example `--clean` don't populate anything it just deletes the generated stuff. Also the command doesn't do nothing when no option is given, so maybe a check that al least 1 option is given would be good. If I understand correctly the name of the command is `perftest` however it doesn't don any performance tests, it just populate a lot of random data for performance testing, so `populate_perftestdb` would be more correct and inline with `populate_testdb`.

app/stac_api/management/commands/perftest.py
```diff
@@ -0,0 +1,118 @@
...
+        geo.srid = 2056
+        geo.transform(4326, clone=False)
+        range(2570000, 2746000, 1000)
+        range(1130000, 1252000, 1000)
+
+        if options['clean']:
+            Collection.objects.filter(name__startswith='perftest').delete()
+        elif options['populate']:
+
+            for collection_id in [f'perftest-collection-{x}' for x in range(30)]:
```
ltshb, line 70: The number of collection and asset could be added as option with a default value. e.g. `--collections=50  --assets=10`

app/stac_api/management/commands/perftest.py
```diff
@@ -0,0 +1,118 @@
...
+logger = logging.getLogger(__name__)
+
+XMIN = 2570000
+# this would be the max extent for CH
+# XMAX = 2746000
+XMAX = 2580000
+YMIN = 1130000
+# this would be the max extent for CH
+# YMAX = 1252000
+YMAX = 1140000
```
ltshb, line 26: Maybe a better short comment on the algorithm that generates the geometry for the items here would be good, how many item would be generated ? Why these numbers ? If I understand your comment, some item would be out of switzerland ? Sorry I'm a totally beginner with geo stuff :-P

app/stac_api/management/commands/perftest.py
```diff
@@ -0,0 +1,118 @@
...
+                            }
+                        )
+
+                        for asset_id in [f'perftest-asset-{x}' for x in range(3)]:
+                            asset, _ = Asset.objects.get_or_create(
+                                item=item,
+                                name=asset_id,
+                                defaults={
+                                    'title': 'my-title',
+                                    'checksum_multihash': "01205c3fd6978a7d0b051efaa4263a09",
```
ltshb, line 105: Is this checksum corresponding to the file content ? I would suggest to use the helper `stac_api.utils.get_sha256_multihash()` here.

app/stac_api/management/commands/perftest.py
```diff
@@ -0,0 +1,118 @@
...
+                                    'geoadmin_variant': random.choice(['var1', 'var2', 'var3']),
+                                    'proj_epsg': random.choice([2056, 4326, 21781]),
+                                    'media_type': random.choice([typ[0] for typ in MEDIA_TYPES]),
+                                    'file': SimpleUploadedFile(
+                                        f'{item.collection.name}/{item.name}/{asset_id}',
+                                        b"dummy content"
+                                    )
+                                }
+                            )
+                            asset.save()
```
ltshb, line 118: The `save()` here is not needed, `get_or_create()` already save the asset.

app/stac_api/management/commands/perftest.py
```diff
@@ -0,0 +1,118 @@
...
+                                'properties_datetime': random_datetime(MIN_DATETIME, MAX_DATETIME),
+                                'properties_title': "My Title",
+                                'geometry': geo
+                            }
+                        )
+
+                        for asset_id in [f'perftest-asset-{x}' for x in range(3)]:
+                            asset, _ = Asset.objects.get_or_create(
+                                item=item,
+                                name=asset_id,
```
ltshb, line 102: Asset ID should have the media type extension, otherwise this won't work anymore when https://jira.swisstopo.ch/browse/BGDIINF_SB-1566 will be implemented

app/stac_api/management/commands/perftest.py
```diff
@@ -0,0 +1,118 @@
...
+
+def random_datetime(start, end):
+    """Generate a random datetime between `start` and `end`"""
+    return start + datetime.timedelta(
+        # Get a random amount of seconds between `start` and `end`
+        seconds=random.randint(0, int((end - start).total_seconds())),
+    )
+
+
+class Command(BaseCommand):
```
ltshb, line 40: Measuring the execution time of this command and printing (using `print()` not logger) it at the end might be a good idea. This way we could see how our code change impact the DB creation performance by checking out earlier version and executing this command

ltshb (changes requested):
See my comments

app/stac_api/management/commands/perftest.py
hansmannj: I'll have to search where we documented it, but for some reason, we intentionally implemented the "empty language" choice. So "empty language" is a valid choice, from what I have in the back of my mind. I'll just need to do some research for the reasoning behind that. (Will do that tomorrow, today I am only working for a very short time..)

hansmannj: Probably it was because `geoadmin:lang` is optional. So in the admin interface the dropdown list should provide an "empty" choice as a valid option. But it might make sense to not return empty languages in the API

ltshb: I've found and fixed an issue with the geoadmin:lang in model, still working on it see https://jira.swisstopo.ch/browse/BGDIINF_SB-1576, I'll put a link to this PR as soon as I'm ready for a PR for this issue. But geoadmin:lang empty should definitely not be valid, for optional the field should be simply set to None in model or not added to the payload for the POST method.

app/stac_api/management/commands/perftest.py
boecklic: removed

app/stac_api/management/commands/perftest.py
boecklic: checksum is now calculated by the service, as  it would be if the asset was uploaded via admin

app/stac_api/management/commands/perftest.py
boecklic: done

app/stac_api/management/commands/perftest.py
```diff
@@ -0,0 +1,118 @@
...
+    help = """Populates the local test database with dummy data
+
```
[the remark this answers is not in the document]
boecklic: removed

app/stac_api/management/commands/perftest.py
boecklic: well, if you wanna be strict with naming then also `populate_testdb` doesn't make sense as we don't populate the testdb but we create sample data. So more appropriate would in my opinion then be `manage.py sampledata --populate` and `manage.py dummydata --populate`.

boecklic: `app/manage.py dummydata --help` does print the help along with the arguments

ltshb (approved):
:1st_place_medal:  very nice

app/stac_api/management/commands/perftest.py
ltshb: I know that this would not give a reliable measure, but more kind of a hint. I mean if suddenly it takes twice the time or even more, then we might have an issue, but small times difference it won't help.

PR #166, opened 2021-01-13
Bugfix BGDIINF_SB-1412: A pseudo point in the bbox should be allowed

rebert:
This fix treats a bbox pseudo point as a point and a
regular bbox as a polygone for intersects with items

### DIFF
--- a/app/stac_api/views.py
+++ b/app/stac_api/views.py
@@ -4,9 +4,8 @@ from collections import OrderedDict
 from datetime import datetime
 
 from django.conf import settings
-from django.contrib.gis.geos import Polygon
-from django.contrib.gis.geos import Point
 from django.contrib.gis.geos import GEOSGeometry
+from django.contrib.gis.geos import Point
 from django.contrib.gis.geos import Polygon
 from django.db.models import Q
 from django.http import JsonResponse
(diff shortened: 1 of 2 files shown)


app/stac_api/views.py
```diff
@@ -220,8 +222,15 @@ def get_queryset(self):
     def filter_by_bbox(self, queryset, bbox):
         try:
             logger.debug('Item query parameter bbox = %s', bbox)
-            query_bbox_polygon = Polygon.from_bbox(bbox.split(','))
-        except ValueError as error:
+            list_bbox_values = bbox.split(',')
+            if list_bbox_values[0] == list_bbox_values[2] and \
+            list_bbox_values[1] == list_bbox_values[3]:
```
ltshb, line 227: I would uses `()` instead of `\` for line breaking in if statement, it makes the code cleaner.

app/stac_api/views.py
```diff
@@ -220,8 +222,15 @@ def get_queryset(self):
...
+                bbox_geometry = Point(float(list_bbox_values[0]), float(list_bbox_values[1]))
```
ltshb, line 228: This code has a potential `IndexError`, either try...expect or test the array size.

ltshb (changes requested):
See comment

ltshb (approved):
:+1:

PR #167, opened 2021-01-13
BGDIINF_SB-1576: Removed empty asset fields from GET request

ltshb:
The description, title, geoadmin_variant and geoadmin_lang were not
always stripped in the GET output when they were empty. That was the
case for the description when it was generated from the admin page (but
not when generated via POST). This was also the case for geoadmin_lang
when generated via Admin page but not from POST.


app/stac_api/admin.py
```diff
@@ -22,6 +22,14 @@
 class AssetAdmin(admin.ModelAdmin):
     readonly_fields = ['href', 'checksum_multihash']
 
+    def save_model(self, request, obj, form, change):
+        if obj.description == '':
+		    # The admin interface with TextArea uses empty string instead
```
rebert, line 27: Flyspeck (space)

app/stac_api/migrations/0050_data_20210113_0953.py
```diff
@@ -0,0 +1,56 @@
...
+                    'collection': asset.item.collection.name,
+                    'item': asset.item.name,
+                    'asset': asset.name
+                }
+            )
+            asset.save()
+
+
+def reverse_migration(apps, schema_editor):
+    # this method allow reverse migration and does nothing
```
rebert, line 44: we will never go back :-)

rebert (approved):
LGTM. I made the migrations locally and tested in the admin interface.

PR #169, opened 2021-01-13
BGDIINF_SB-1537: Better POST search examples in spec

rebert:
And some fixes of POST

### DIFF
--- a/spec/components/schemas.yml
+++ b/spec/components/schemas.yml
@@ -314,8 +314,8 @@ components:
       type: object
       example:
         collections:
-        - collection-1
-        - collection-2
+        - ch.swisstopo.swisstlmregio
+        - ch.bfe.energieschweiz
     confClasses:
       properties:
         conformsTo:
@@ -525,8 +525,8 @@ components:
       type: object
       example:
         ids:
-        - item-1
-        - item-2
+        - swisstlmregio-2019
+        - swisstlmregio-2020
     intersectsFilter:
       description: Only returns items that intersect with the provided polygon.
       properties:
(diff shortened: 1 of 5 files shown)


spec/components/schemas.yml
```diff
@@ -306,6 +312,8 @@ components:
         collections:
           $ref: "#/components/schemas/collectionsArray"
       type: object
+      example:
+        collections: "collection-1,collection-2"
```
rebert, line 316: [ ... ]

spec/components/schemas.yml
```diff
@@ -513,6 +521,8 @@ components:
         ids:
           $ref: "#/components/schemas/ids"
       type: object
+      example:
+        ids: "item-1,item-2"
```
rebert, line 525: [ .. ]

ltshb (commented):
I would use realistic values for the examples (not 'collection-1' or 2525-01-01T00:00:00.000Z, ...) all other examples uses realistic values. Same for the bbox, I haven't checked if the value are realistic but they should be.

Looking at the SPEC if found that the description should be updated as well: 
**GET**:

_Retrieve Items matching filters. Intended as a shorthand API for simple queries._

~~_This method is optional, but you MUST implement POST /search if you want to implement this method._~~

**_If this endpoint is implemented on a server, it is required to add a link referring to this endpoint with rel set to search to the links array in GET /. As GET is the default method, the method may not be set explicitly in the link._**

About the last part the text should be modified and maybe be more clear on what it means. Actually I don't really understand it, maybe the SPEC md provide more infos on this ? And we currently don't have any search link, so this probably needs to be implemented ? @boecklic ?

OK I've just had a look at the MD of the SPEC, so from my understanding the of the second parts of the description above in bold is that we need to have in the landing page two links to the search endpoint; one with an attribute `method=GET` and one with an attribute `method=POST`. I would then remove this comment in the SPEC as it doesn't really help the user in my opinion.

boecklic (approved):
:+1:

PR #175, opened 2021-01-15
BGDIINF_SB-1484: Own QueryClass for fiter functions

rebert:
Putting general filter functions to become methods of a custom QuerySet class

### DIFF
--- a/app/stac_api/managers.py
+++ b/app/stac_api/managers.py
@@ -16,7 +16,7 @@ logger = logging.getLogger(__name__)
 
 class ItemQuerySet(models.QuerySet):
 
-    def by_bbox(self, bbox):
+    def filter_by_bbox(self, bbox):
         '''Filter a querystring with a given bbox
 
         This function is a helper function, for some views, to add a bbox filter to the queryset.
@@ -62,7 +62,7 @@ class ItemQuerySet(models.QuerySet):
 
         return self.filter(geometry__intersects=bbox_geometry)
 
-    def by_datetime(self, date_time):
+    def filter_by_datetime(self, date_time):
         '''Filter a queryset by datetime
 
         Args:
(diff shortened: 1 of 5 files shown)


app/stac_api/managers.py
```diff
@@ -0,0 +1,163 @@
...
+from rest_framework.exceptions import ValidationError
+
+from stac_api.validators import validate_geometry
+
+logger = logging.getLogger(__name__)
+
+
+class ItemQuerySet(models.QuerySet):
+
+    def by_bbox(self, bbox):
```
boecklic, line 19: I'd maybe call it `filter_by_bbox`, slightly more descriptive what the method is doing.

boecklic (commented):
Nice! Made a little comment

app/stac_api/managers.py
hansmannj: If this change is implemented, it might make sense for consistence to also rename the other `by_something` filters, e.g. `by_datetime`.

app/stac_api/managers.py
```diff
@@ -0,0 +1,163 @@
...
+        return start, end
+
+
+class ItemManager(models.Manager):
+
+    def get_queryset(self):
+        return ItemQuerySet(self.model, using=self._db).select_related('collection')
+
+    def by_bbox(self, bbox):
+        return self.get_queryset().by_bbox(bbox)
```
ltshb, line 163: The `by_datetime()` method is missing.

app/stac_api/managers.py
ltshb: That was actually my idea to be more concise, but I'm also OK with `filter_by_bbox`, like @hansmannj  mentioned we need then to be consistent everywhere.

ltshb (approved):
I'll let you choose if you want to use `filter_by_bbox` or `by_bbox`, but don't forget to be consistent. Otherwise looks good, only the `(filter_)by_datetime` method in the manager is missing.

hansmannj (approved):
:+1: 
LGTM, just the naming issue that needs to be decided upon

PR #176, opened 2021-01-18
 BGDIINF_SB-1571: Sets appropriate cache headers in GET and HEAD responses

ltshb:
Adds the following headers to the django views:

- `Expires: DATE_600_SEC_AHEAD_FROM_NOW`
- `Cache-control: max-age=600, public`
- `Vary: Accept, Cookie` (Added automatically by Django)

The max-age and expire is configurable via `HTTP_CACHE_SECONDS`

For the static files the following headers are added (added automatically by Whitenoise):

- `cache-control: max-age=3600, public, immutable`
- `Expires: DATE_600_SEC_AHEAD_FROM_NOW`

The max-age and expire is configurable via `HTTP_STATIC_CACHE_SECONDS`

_NOTES:_
- it would be useful to have caching enabled in /search POST endpoint, however CloudFront apparently doesn't supports caching on POST requests.
- I did not use the Django CacheMiddleware, because it first install a default cache ([Local-memory caching](https://docs.djangoproject.com/en/3.1/topics/cache/#local-memory-caching) which is not a good choice for PROD) and because it has conflict with the ETag Precondition. If we want to also add caching in Django and make uses of CacheMiddleware we would need first to install a proper cache backend and then to move the ETag functionality in a Middleware as well (it is currently done per view using a decorator).
- Currently because we use the Django CSRF Middleware, the `Vary: Coockie` header is added to each responses. In my opinion this header for our caching is not required and depending on the CloudFront configuration, could even reduce the cache hit ratio of CloudFront. Once deployed on INT we will need to review the CloudFront configuration and do some testing.
- See https://github.com/geoadmin/infra-kubernetes/pull/42 for k8s configuration

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -143,7 +143,7 @@ lint:
 test:
 	# Collect static first to avoid warning in the test
 	$(PYTHON) $(DJANGO_MANAGER) collectstatic --noinput
-	$(PYTHON) $(DJANGO_MANAGER) test --verbosity=2 --parallel 8 $(TEST_DIR)
+	$(PYTHON) $(DJANGO_MANAGER) test --verbosity=2 --parallel 20 $(TEST_DIR)
 
 
 ###################
--- a/app/middleware/cache_headers.py
+++ b/app/middleware/cache_headers.py
@@ -1,0 +1,42 @@
+import logging
+
+from django.conf import settings
+from django.utils.cache import add_never_cache_headers
+from django.utils.cache import patch_cache_control
+from django.utils.cache import patch_response_headers
+
+logger = logging.getLogger(__name__)
+
+
+class CacheHeadersMiddleware:
+    '''Middleware that adds appropriate cache headers to GET and HEAD methods.
+
+    NOTE: /checker and /get-token endpoints are marked as never cache.
+    '''
+
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        # Code to be executed for each request before
+        # the view (and later middleware) are called.
+
+        response = self.get_response(request)
+
+        # Code to be executed for each request/response after
+        # the view is called.
+
+        if request.path in ['/checker', '/get-token']:
+            # never cache the /checker and /get-token endpoints.
+            add_never_cache_headers(response)
+        elif request.method in ('GET', 'HEAD'):
+            logger.debug(
+                "Patching cache headers for request %s %s",
+                request.method,
+                request.path,
+                extra={"request": request}
+            )
+            patch_response_headers(response, settings.CACHE_MIDDLEWARE_SECONDS)
+            patch_cache_control(response, public=True)
+
+        return response
--- a/app/stac_api/admin.py
+++ b/app/stac_api/admin.py
@@ -114,7 +114,7 @@ class ItemAdmin(admin.GeoModelAdmin):
             }
         ),
     )
-    # customisation of the geometry field
+    # customization of the geometry field
     map_template = finders.find('admin/ol_swisstopo.html')  # custom swisstopo
     wms_layer = 'ch.swisstopo.pixelkarte-farbe-pk1000.noscale'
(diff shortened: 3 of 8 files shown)


ltshb:
Just find out that the `get-tocken` endpoint should not be cached

ltshb:
Ready for review :-)

ltshb:
mmh caching is breaking ETag... I'll have a look at it and also add unittest for header cache.

Makefile
```diff
@@ -143,7 +143,7 @@ lint:
 test:
 	# Collect static first to avoid warning in the test
 	$(PYTHON) $(DJANGO_MANAGER) collectstatic --noinput
-	$(PYTHON) $(DJANGO_MANAGER) test --verbosity=2 --parallel 8 $(TEST_DIR)
+	$(PYTHON) $(DJANGO_MANAGER) test --verbosity=2 --parallel 20 $(TEST_DIR)
```
rebert, line 146: glad having a stronger workstation now :-)

app/stac_api/admin.py
```diff
@@ -114,7 +114,7 @@ class ItemAdmin(admin.GeoModelAdmin):
             }
         ),
     )
-    # customisation of the geometry field
+    # customization of the geometry field
```
rebert, line 117: Why not... US English. Personally I prefer British English

rebert (approved):
LGTM - these tweaks are hard to test locally ...

app/middleware/cache_headers.py
```diff
@@ -0,0 +1,42 @@
...
+    def __call__(self, request):
+        # Code to be executed for each request before
+        # the view (and later middleware) are called.
+
+        response = self.get_response(request)
+
+        # Code to be executed for each request/response after
+        # the view is called.
+
+        if request.path in ['/checker', '/get-token']:
```
hansmannj, line 29: What about the admin interface? Would it also make sense not to cache that one, too?

ltshb: Admin interface are not cached. The middleware don't act on the admin interface, but only on the Django views. For the Admin interface the statics files are then cached via Whitenoise while the admin api is not cached (django adds never cache headers)

hansmannj (approved):
:+1: LGTM, as @rebert said, difficult to test locally.
And probably also difficult to unit test. Each test would take 600 or 3600 seconds :wink:
Probably the default value could be overridden in a unit test with a much shorter time..

ltshb:
> +1 LGTM, as @rebert said, difficult to test locally.
> And probably also difficult to unit test. Each test would take 600 or 3600 seconds wink
> Probably the default value could be overridden in a unit test with a much shorter time..

@hansmannj  Oops good point unittest, I can add unittest to check the headers, I forgot that, I'll add them.

app/tests/test_generic_api.py
```diff
@@ -293,3 +296,79 @@ def test_delete_precondition(self):
...
+                    response['Cache-Control'],
+                    'max-age=3600, public',
+                    msg='Wrong cache-control values'
+                )
+
+                self.assertTrue(response.has_header('Expires'), msg="Expires header missing")
+
+                expires = datetime.strptime(response['Expires'], '%a, %d %b %Y %H:%M:%S GMT')
+                self.assertAlmostEqual((expires - now).total_seconds(),
+                                       timedelta(seconds=3600).total_seconds(),
```
hansmannj, line 373: Nice, thanks for adding the unit tests.
Just have to small points here:

* Probably the 3600 should be a variable and also be uses in the `override_settings` decorator, so in case the value will be changed in future in one place, it will be changed everywhere and the test will not crash
* It is probably very unlikely that a delta of 2 is not enough, but would there be an alternative to using `AlmostEqual` here? Because what we want to check here is, that the `Expires` is something between 3600 and 0 (or simply `le` 3600, right?) and not that this check can be done in less than 2 seconds after the request was send.

hansmannj: Resolved :-)

PR #177, opened 2021-01-18
BGDIINF_SB-1484: filter_by_datetime in manager

rebert:


ltshb (changes requested):
Renaming issues in views.py:313 and views.py:316. I would also add unittest to fully cover the GET query request.

ltshb (approved):

PR #181, opened 2021-01-21
BGDIINF_SB-1545: Refactoring search endpoint

rebert:
- Own ValidationClass
- Own QuerySets


app/stac_api/managers.py
```diff
@@ -153,6 +162,73 @@ def _parse_datetime_query(self, date_time):
             )
         return start, end
 
+    def filter_by_ids(self, ids_array):
+        '''Filter by ids parameter
```
ltshb, line 166: Maybe `Filter by item name` would be more explicit and won't introduced missunderstanding with the real id (pk)

app/stac_api/managers.py
```diff
@@ -153,6 +162,73 @@ def _parse_datetime_query(self, date_time):
...
+        json_query = json.loads(query)
+        for attribute in json_query:
+            for operator in json_query[attribute]:
+                value = json_query[attribute][operator]  # get the values given by the operator
+                # treat date
+                if attribute in queriabel_date_fields:
+                    if isinstance(value, list):
+                        value = [fromisoformat(i) for i in value]
+                    else:
+                        value = fromisoformat(value)
```
ltshb, line 222: Either adds a comment in docstring that the function can raises `ValueError` in case of invalid datetime string, or add a try except around `fromisoformat()` to handle invalid datetime string.

app/stac_api/managers.py
```diff
@@ -153,6 +162,73 @@ def _parse_datetime_query(self, date_time):
...
+
+        Args:
+            query: string
+                The query is a json formatted string.
+
+        Returns:
+            queryset filtered by query
+        '''
+        queriabel_date_fields = ['datetime', 'created', 'updated']
+        json_query = json.loads(query)
```
boecklic, line 213: I would do the conversion (and error handling!) of query string to json in the view and provide the `query` as dict parameter here.

app/stac_api/managers.py
```diff
@@ -153,6 +162,73 @@ def _parse_datetime_query(self, date_time):
...
+        for attribute in json_query:
+            for operator in json_query[attribute]:
+                value = json_query[attribute][operator]  # get the values given by the operator
+                # treat date
+                if attribute in queriabel_date_fields:
+                    if isinstance(value, list):
+                        value = [fromisoformat(i) for i in value]
```
boecklic, line 220: what happens if the value is not in isoformat?

app/stac_api/validators.py
```diff
@@ -216,3 +223,306 @@ def validate_asset_multihash(value):
...
+        queriabel_fields = self.queriabel_date_fields + self.queriabel_str_fields
+
+        # validate json
+        try:
+            json_query = json.loads(query)
+        except json.JSONDecodeError as error:
+            message = f"The application could not decode the query parameter" \
+                      f"Please check the syntax ({error})." \
+                      f"{query}"
+            raise RestValidationError(code='query-invalid', detail=_(message))
```
ltshb, line 304: Logging error is missing here.

app/stac_api/validators.py
```diff
@@ -216,3 +223,306 @@ def validate_asset_multihash(value):
...
+        Args:
+            query: string
+                The query parameter to be validated.
+        '''
+        # summing up the fields based of different types
+        queriabel_fields = self.queriabel_date_fields + self.queriabel_str_fields
+
+        # validate json
+        try:
+            json_query = json.loads(query)
```
boecklic, line 299: being picky: this is rather `query_dict` than a `json_query` (the object is a dict, not a string formatted as json)... so it's rather `query_dict = json.loads(json_query)`...

app/stac_api/validators.py
```diff
@@ -216,3 +223,306 @@ def validate_asset_multihash(value):
             message=_('Invalid multihash value; %(error)s'),
             params={'error': error}
         )
+
+
+class ValidateSearchRequest:
```
boecklic, line 228: Nice! Looks good also from security point of view. (the whole `ValidateSearchRequest` class)

app/stac_api/managers.py
ltshb: OK I've seen that this is catch by the validator, so not needed to catch it here, but I would add the `Raises: ValueError ...` in docstring so we know that this can raise exception with invalid data.

boecklic (approved):
Nice!

app/stac_api/validators.py
```diff
@@ -216,3 +223,306 @@ def validate_asset_multihash(value):
...
+            date_time: string
+                The datetime to get validated
+        '''
+        start, sep, end = date_time.partition('/')
+        message = None
+        try:
+            if start != '..':
+                start = datetime.fromisoformat(start.replace('Z', '+00:00'))
+            if end and end != '..':
+                end = datetime.fromisoformat(end.replace('Z', '+00:00'))
```
ltshb, line 431: The `replace('Z', '+00:00')` is not needed use the `fromisoformat` helper like above instead of using the datetime method `fromisoformat`

app/stac_api/validators.py
```diff
@@ -216,3 +223,306 @@ def validate_asset_multihash(value):
...
+            if (
+                list_bbox_values[0] == list_bbox_values[2] and
+                list_bbox_values[1] == list_bbox_values[3]
+            ):
+                bbox_geometry = Point(float(list_bbox_values[0]), float(list_bbox_values[1]))
+            else:
+                bbox_geometry = Polygon.from_bbox(list_bbox_values)
+            validate_geometry(bbox_geometry)
+
+        except (ValueError, ValidationError, IndexError) as error:
```
ltshb, line 507: What about  `Geometry` exceptions in the `Point` and `Polygon` constructor ?

ltshb: Should not like below also catch `GDALException` ?

app/stac_api/utils.py
```diff
@@ -189,3 +190,21 @@ def create_multihash_string(digest, hash_code):
         multihash string
     '''
     return multihash.to_hex_string(multihash.encode(digest, hash_code))
+
+
+def harmonize_post_get_for_search(request):
```
ltshb, line 195: Short docstring would be good

app/stac_api/validators.py
ltshb, line 228: This class is very good however it should be part of the `validators_serializer.py` module to not messup with the django ValidationError and the rest_framework ValidationError. The validators.py module should be only for validators that can be used by models and/or serializers. For views/serializers validator only we should use validators_serializer.py

app/stac_api/validators.py
```diff
@@ -1,14 +1,21 @@
...
 import multihash
 
+from django.contrib.gis.gdal.error import GDALException
 from django.contrib.gis.geos import GEOSGeometry
+from django.contrib.gis.geos import Point
+from django.contrib.gis.geos import Polygon
 from django.core.exceptions import ValidationError
 from django.utils.translation import gettext_lazy as _
 
+from rest_framework.exceptions import ValidationError as RestValidationError
```
ltshb, line 15: See my comment below but this is not needed anymore if moved into validators_serializer.py

ltshb (commented):
Great jobs this make the code much clean :-) just some minor comments.

app/stac_api/validators.py
```diff
@@ -216,3 +223,306 @@ def validate_asset_multihash(value):
...
+                    isinstance(value, str) and operator in int_operators and
+                    attribute in int_operators
+                ):
+                    message = f"You are not allowed to compare a string/date ({attribute})"\
+                              f" with a number operator." \
+                              f"for string use one of these {str_operators}"
+                    self.errors[f"query-operator-{operator}"] = _(message)
+                if (
+                    isinstance(value, int) and operator in str_operators and
+                    operator in str_operators
```
hansmannj, line 363: `operator in str_operators` is redundant here. Should one of the two perhaps be `attribute in str_operators` here?

app/stac_api/validators.py
hansmannj, line 228: Yes, very nice class :-)

hansmannj (approved):
Very nice, thanks :+1:
Just one question regarding a redundancy in the code (see comment).
And perhaps a further unit test covering the case where a query is done to retrieve items with a certain datetime (as in https://jira.swisstopo.ch/browse/BGDIINF_SB-1597) might make sense.
But we can also add that later and merge now.

app/stac_api/managers.py
rebert: this is handled in the validator before

rebert: a ValidationError will be raised in the validator

app/stac_api/validators.py
rebert: absolutely

app/stac_api/validators.py
rebert: ok

app/stac_api/validators.py
rebert: will have a look

app/stac_api/managers.py
rebert: thanks

app/stac_api/managers.py
```diff
@@ -41,8 +41,8 @@ def filter_by_bbox(self, bbox):
             logger.debug('Query parameter bbox = %s', bbox)
             list_bbox_values = bbox.split(',')
             if (
-                list_bbox_values[0] == list_bbox_values[2] and
-                list_bbox_values[1] == list_bbox_values[3]
+                float(list_bbox_values[0]) == float(list_bbox_values[2]) and
+                float(list_bbox_values[1]) == float(list_bbox_values[3])
```
hansmannj, line 43: Float comparisons are a bit tricky, I think (see: https://stackoverflow.com/questions/26091689/python-floating-point-number-comparison)
We have our `float_in()` in `collection_summaries.py` for similar reasons.

I am not sure, if we would run into problems here, but it might be safer, to use a function for that comparison for equality.
E.g. something like this:
```
def floats_equal(float1, float2):
     if abs(float1 - float2) < whatever_small_number_e_g_1e-10:
         return True
    return False
```

app/stac_api/validators.py
```diff
@@ -496,8 +496,8 @@ def validate_bbox(self, bbox):
         try:
             list_bbox_values = bbox.split(',')
             if (
-                list_bbox_values[0] == list_bbox_values[2] and
-                list_bbox_values[1] == list_bbox_values[3]
+                float(list_bbox_values[0]) == float(list_bbox_values[2]) and
+                float(list_bbox_values[1]) == float(list_bbox_values[3])
```
ltshb, line 500: Same as above, would keep the string comparaison here

app/stac_api/managers.py
hansmannj: Even better :nerd_face:

app/stac_api/validators.py
ltshb: Oops forget about testing string equality, I didn't thought of this use case : `[6.10,47.10,6.1,47.1]` which would not pass the string equality but would pass the float equality !

app/stac_api/managers.py
ltshb: Oops forget about testing string equality, I didn't thought of this use case : `[6.10,47.10,6.1,47.1]` which would not pass the string equality but would pass the float equality !

hansmannj: Haha, same thought at the same time :-)

app/stac_api/managers.py
rebert: the validation has been removed completely from the QuerySet filter

app/stac_api/managers.py
rebert: It is more about the POST, where 
{ [6.10,
      47.10,
  6.1,
       47.1]

brought the comparison into trouble

hansmannj (approved):
Very nice job :smiley: 
I did not re-check all the code but only had a look at the issues I commented on.
Looks good to me :+1:

app/stac_api/managers.py
```diff
@@ -153,6 +161,64 @@ def _parse_datetime_query(self, date_time):
...
+
+    def filter_by_intersects(self, intersects):
+        '''Filter by intersects parameter
+
+        Args:
+            intersects: string
+                Is a geojson formatted string
+
+        Returns:
+            queryset filtered by intersects
```
ltshb, line 196: You could add the `Raises:` in the docstring with the exception that might occur if the intersects is not correct.

app/stac_api/managers.py
```diff
@@ -116,11 +119,16 @@ def _filter_by_datetime_range(self, start_datetime, end_datetime):
...
+        A helper function of filter_by_datetime
+
+        Args:
+            date_time: string
+                Datetime as string (should be in isoformat)
+
+        Returns:
+            queryset filtered by datetime
+
+        Raises: ValidationError
```
ltshb, line 131: Short comment to explain why ValidationError can occurs: e.g: 
`
Raises:
    ValidationError: when the date_time string is not a valid isoformat
`

ltshb (approved):
Great jobs :+1:  
There is two minor docstring improvement possible.

PR #182, opened 2021-01-21
BGDIINF_SB-1546: Improvements of collections update logs

ltshb:
This improves the logging of the collection updates (extent and summaries).

To simplify the logging and improve the filter possibilities, I moved all code related to the collection update in separate modules. This give the possibility for example on ELK to filter all the logs related to the collection update summaries, or to the update spatial extent, ...

In the last commits I refactored the collection update in mixins, this is not really needed and could be reverted if you don't like it ? Note that some logs are then still in the models.py module.

@rebert @hansmannj  @boecklic for the review you might want to review commit by commit, otherwise the diff is just huge due to reorganisation. The last 3 commits is just the refactoring in mixin.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@
   - [Setting up the local database](#setting-up-the-local-database)
   - [Using a local PostGres database instead of a container](#using-a-local-postgres-database-instead-of-a-container)
   - [Starting dev server](#starting-dev-server)
-  - [Running test](#running-test)
+  - [Running tests](#running-tests)
   - [Using Django shell](#using-django-shell)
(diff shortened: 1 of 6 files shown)


boecklic (approved):
Nice! Maybe you could combine all the mixins in one `collection_mixins.py`?

ltshb:
> Nice! Maybe you could combine all the mixins in one `collection_mixins.py`?

We could be then we would loose the ability to filter logs based on the type of collections update in Kibana which can be usefull if we want to debug for example the temporal extent update.

hansmannj (approved):
Nice, looks good to me so far.
But I did not have enough time to dive into the very details of the code (again) ;-)

PR #185, opened 2021-01-25
BGDIINF_SB-1541 index model fields

hansmannj:
* added indices where extensive tests identified them to be useful
* added debug prints for SQL query and output of `EXPLAIN.. ANALYZE`


app/stac_api/models.py
```diff
@@ -483,6 +487,31 @@ class Item(models.Model):
 
     class Meta:
         unique_together = (('collection', 'name'),)
+        indexes = [
+            models.Index(fields=['name'], name='item_name_idx'),
+            models.Index(fields=['properties_datetime'], name='item_datetime_idx'),
+            models.Index(fields=['properties_start_datetime'], name='item_start_datetime_idx'),
+            models.Index(fields=['properties_end_datetime'], name='item_end_datetime_idx'),
+            BrinIndex(fields=['created'], name='item_created_idx'),
+            models.Index(fields=['updated'], name='item_updated_idx'),
```
ltshb, line 496: Are those really required ? For which use case ? And why only have the index on the Items and not on the collections and Assets ?

boecklic (changes requested):
In general I'd be conservative with indices, unless you're sure that you really benefit from having them (as mentioned in a comment, we should probably get an introduction from @ltclm / @rebert into `EXPLAIN ANALYZE`, applied on real data). Every additional index creates an additional timing overhead on write operations, so it should probably be commented why it's necessary and for which use cases (unless it's obvious). Even if service-stac is not that critical in timing for write requests, it could be a good opportunity to learn a bit more on PostgreSQL internals.

app/stac_api/models.py
```diff
@@ -483,6 +487,31 @@ class Item(models.Model):
...
+                fields=['properties_datetime', 'properties_end_datetime'],
+                name='item_datetime_end_dttm_idx'
+            ),
+            models.Index(
+                fields=[
+                    'properties_datetime', 'properties_start_datetime', 'properties_end_datetime'
+                ],
+                name='item_dttme_start_end_dttm_idx'
+            ),
+        ]
```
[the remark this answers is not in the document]
hansmannj: +1 for the intro and I will also take time to check the EXPLAIN ANALYZE PostgeSQL command.
What I did in this first shot is, to check for all `.filter()` or `Q()` and the like and create indices for the queried fields. I tried to avoid redundancies with autogenerated indices for PK and FK, but I might have missed some and still have redundancies.

I think, an intro by @ltclm / @rebert might be very nice here :-)

app/stac_api/models.py
hansmannj: This is due to `queriable_date_fields = ['datetime', 'created', 'updated']` in `views.py`

app/stac_api/models.py
```diff
@@ -654,6 +683,14 @@ class Language(models.TextChoices):
 
     class Meta:
         unique_together = (('item', 'name'),)
+        indexes = [
+            models.Index(fields=['name'], name='asset_name_idx'),
+            models.Index(fields=['item'], name='asset_item_name_idx'),
```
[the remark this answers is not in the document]
hansmannj: Yep, `item` is a FK and redundant here, thanks

app/stac_api/models.py
```diff
@@ -483,6 +487,31 @@ class Item(models.Model):
...
     class Meta:
         unique_together = (('collection', 'name'),)
+        indexes = [
+            models.Index(fields=['name'], name='item_name_idx'),
+            models.Index(fields=['properties_datetime'], name='item_datetime_idx'),
+            models.Index(fields=['properties_start_datetime'], name='item_start_datetime_idx'),
+            models.Index(fields=['properties_end_datetime'], name='item_end_datetime_idx'),
+            BrinIndex(fields=['created'], name='item_created_idx'),
+            models.Index(fields=['updated'], name='item_updated_idx'),
+            models.Index(fields=['properties_title'], name='item_title_idx'),
```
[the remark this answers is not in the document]
hansmannj: According to `queriable_str_fields = ['title'] in `views.py` it should be.
`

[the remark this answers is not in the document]
boecklic: You're right, it is queryable in POST /search requests https://service-stac.dev.bgdi.ch/api/stac/v0.9/static/apitransactional.html#operation/postSearchSTAC

app/stac_api/models.py
```diff
@@ -270,6 +274,51 @@ class Item(models.Model):
...
+            # geometry is used in "filter_by_intersects", views.py:415
+            models.Index(fields=['geometry'], name='item_geometry_idx'),
+            # combination collection name and item name is used in get_item_etag
+            # views.py:58
+            # TODO: still need to figure out, how to create a multi-field index # pylint: disable=fixme
+            # with fields from two different classes..
+            # self.collection.name won't work
+            # models.Index(
+            #     fields=['self.collection.name', 'name'], name='item_coll_name_item_name_idx'
```
hansmannj, line 298: Any idea anyone, how the `self.collection.name` field would be correctly referenced here in a multi-field index with fields from two different classes?

hansmannj: `collection__name` worked neither

boecklic: I'm not even sure this is possible conceptually, from how relational databases work. I could imagine that you need two queries anyhow for this usecase, the first one to get the collection_id for a specific collection name (for this query you would use an index on collection name) and a second query to get the item name, searching only in items with a specific collection id (for this you would use a combined index collection_id, item_name, (which in django would be `fields=['collection','name']).

app/stac_api/models.py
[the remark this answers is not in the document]
ltclm: you can paste the explain analyze output here: https://explain.depesz.com/
it will show you the slow parts of the query and there is lots of background information on the different numbers.
A first step would be to look out for sequential scans, such scans should be avoided whenever possible.

i have no idea about the django / orm limitations concerning index creation in postgres. I would try to analyze the slow queries with postgres / pgadmin, identify the missing indexes and then try to add them to the orm model ( if possible ).

app/stac_api/models.py
ltshb: I had similar issue with MongoDB. To solve it, we could duplicate the collection name in the item table, we will have a drawback in the write but it will improve performance in the query.

hansmannj:
In order to check the influence of model field indices I made 10 queries like:
```bash
curl --request POST \
  --url http://0.0.0.0:8000/api/stac/v0.9/search \
  --header 'Content-Type: application/json' \
  --data '{"query": 
 	{"datetime":
	 	{"lte":"2525-01-02T00:00:00Z"}  
	},
 "limit": 100
}'
```

one without indices, one with indices.
There was not a significant difference, and if any, then the queries w/o the indices were even slightly faster:

`SQLPanel_sql_time`, average of 10 queries:
w/o indices:  118.3ms (201 queries), with indices: 124.6 ms (204 queries)

`TimerPanel_stime` (System CPU time):
w/o indices: 265.5 ms, with indices: 288.2 ms

hansmannj:
However, this query uses the ` item_dttme_start_end_dttm_idx` index:
```bash
curl --request POST \
  --url http://0.0.0.0:8000/api/stac/v0.9/search \
  --header 'Content-Type: application/json' \
  --data '{"query":
		{"datetime":
				{"in":["1960-10-28T13:05:10Z","2525-10-19T00:00:00Z"]}
		}
}'
```

But for some reason, no items are returned (found). Unfortunately I do not have time to debug further today and most likely not during the next week in my holidays. So could continue the week after, or someone can continue here, in case it is urgent.

ltclm:
Salut @hansmannj 

the btree index seems to be the right choice for datetime columns and the operators that we are using for the filter.
I did some tests with 250'000 items:

```sql
CREATE INDEX idx_stac_api_item_created
    ON public.stac_api_item USING btree
    (created ASC NULLS LAST)
    TABLESPACE pg_default; 

select  count(*), min(created), max(created) FROM stac_api_item;
250000	"2020-02-13 14:22:56.246497+00"	"2021-02-12 14:22:16.891535+00"

-- with btree index on created column
EXPLAIN ANALYZE select * FROM stac_api_item WHERE created > '2020-01-01' -- seq scan --> alle daten		-- Execution Time: 61.298 ms
EXPLAIN ANALYZE select * FROM stac_api_item WHERE created > '2020-07-01' -- seq scan --> alle daten		-- Execution Time: 56.135 ms
EXPLAIN ANALYZE select * FROM stac_api_item WHERE created > '2020-12-01' -- index --> nicht alle daten		-- Execution Time: 34.665 ms
EXPLAIN ANALYZE select * FROM stac_api_item WHERE created > '2021-07-01' -- index --> keine daten		-- Execution Time: 0.019 ms
EXPLAIN ANALYZE select * FROM stac_api_item WHERE created > '2021-07-01' -- index --> keine daten		-- Execution Time: 0.021 ms

-- without btree index on created column
EXPLAIN ANALYZE select * FROM stac_api_item WHERE created > '2020-01-01' -- seq scan --> alle daten		-- Execution Time: 63.226 ms
EXPLAIN ANALYZE select * FROM stac_api_item WHERE created > '2020-07-01' -- seq scan --> alle daten		-- Execution Time: 57.670 ms
EXPLAIN ANALYZE select * FROM stac_api_item WHERE created > '2020-12-01' -- seq scan  --> nicht alle daten	-- Execution Time: 45.828 ms
EXPLAIN ANALYZE select * FROM stac_api_item WHERE created > '2021-07-01' -- seq scan  --> keine daten		-- Execution Time: 119.938 ms
EXPLAIN ANALYZE select * FROM stac_api_item WHERE created > '2021-07-01' -- seq scan  --> keine daten		-- Execution Time: 112.852 ms


```

Whether the index is used or not depends on the datetime filter you are using, if the filter is covering the whole dataset. a sequential scan will be done in any case and the index is not used. see first two examples. In any other cases an index scan will be done.

This is the impact of this index on the write performance (250'000 items with sql):

```sql
-- without the index
INSERT 0 250000
Query returned successfully in 11 secs 261 msec.

-- with index
INSERT 0 250000
Query returned successfully in 13 secs 550 msec.
```

hansmannj:
> Salut @hansmannj
> 
> the btree index seems to be the right choice for datetime columns and the operators that we are using for the filter.
> I did some tests with 250'000 items:
> 

Cool, thanks for testing @ltclm!
In the special case of `created` I also thought about using BRIN, as the data for that column should be sorted and in theory also never change, once created. But after reading this: https://info.crunchydata.com/blog/avoiding-the-pitfalls-of-brin-indexes-in-postgres I am not so sure about using BRIN any more..

app/stac_api/models.py
hansmannj: I'll drop this specific combined index for now. Currently the potential use case for this index uses a combination of the indices `item_name_idx` and `stac_api_collection_pkey` instead. That might do as well and prevents this probably too expensive and not really needed index in this case.

app/stac_api/models.py
```diff
@@ -277,6 +280,31 @@ class Item(models.Model):
...
+            # combination of datetime and start_ and end_datetimes are used in
+            # managers.py:110 and following
+            models.Index(
+                fields=[
+                    'properties_datetime', 'properties_start_datetime', 'properties_end_datetime'
+                ],
+                name='item_dttme_start_end_dttm_idx'
+            ),
+        ]
+
```
ltshb, line 307: I would leave this block either after all class attribute definition or before them, but not in the middle. Because Collection and Asset already have the inner `Class Meta` at the top I would put it at the top.

app/stac_api/views.py
```diff
@@ -44,6 +44,13 @@ def get_collection_etag(request, *args, **kwargs):
     The ETag is an UUID4 computed on each object changes (including relations; provider and links)
     '''
     tag = get_etag(Collection.objects.filter(name=kwargs['collection_name']))
+    # uncomment the following lines to see the SQL query and the EXPLAIN.. ANALYZE
+    # output:
+    # logger.debug(
+    #     Collection.objects.filter(name=kwargs['collection_name']
+    #                              ).explain(verbose=True, analyze=True)
+    # )
+    # logger.debug(Collection.objects.filter(name=kwargs['collection_name']).query)
```
ltshb, line 53: I just had idea, instead of commenting these lines out, we could simply add a settings to enable them, e.g. 
```python
if settings.DEBUG_ENABLE_DB_EXPLAIN_ANALYZE:
   logger.debug ....
```
The settings would be `False` by default and could be set by environment variable. This way we could always enable this logs even in already deployed version if needed. @boecklic what do you think ?

ltshb (commented):
See inline comments.

app/stac_api/models.py
```diff
@@ -277,6 +280,31 @@ class Item(models.Model):
...
+            models.Index(fields=['properties_datetime'], name='item_datetime_idx'),
+            models.Index(fields=['properties_start_datetime'], name='item_start_datetime_idx'),
+            models.Index(fields=['properties_end_datetime'], name='item_end_datetime_idx'),
+            # created, updated, and title are "queriable" in the search endpoint
+            # see: views.py:322 and 323
+            models.Index(fields=['created'], name='item_created_idx'),
+            models.Index(fields=['updated'], name='item_updated_idx'),
+            models.Index(fields=['properties_title'], name='item_title_idx'),
+            # geometry is used in "filter_by_intersects", views.py:415 and filtering for bbox
+            models.Index(fields=['geometry'], name='item_geometry_idx'),
```
ltclm, line 297: not really sure about that but i think geometry index is created by default by django. i have found one in my local django installation:
```sql
CREATE INDEX stac_api_item_geometry_id
    ON public.stac_api_item USING gist
    (geometry)
    TABLESPACE pg_default;
```

https://docs.djangoproject.com/en/3.1/ref/contrib/gis/model-api/#spatial-index

ltclm (approved):
can't really judge the code part, but the db index part looks good to me!

app/stac_api/models.py
hansmannj: That was my intention when I shifted it from the bottom, where it currently is, to the top. Somehow pasted it at the wrong place though ;-)

app/stac_api/views.py
hansmannj: Implemented

app/stac_api/models.py
hansmannj: Good point @ltclm, thanks for the hint.
I don't really know much about the geometry indexing so far, most likely @rebert knows more about GeoDjango and geometric indexing.
To test it, I disabled the manually created geometry index and migrated backwards. The requests that formerly used the manually created geometry index now use an auto-generated index `stac_api_item_geometry_id`. So I guess that is fine and we can also drop the manually created index on geometry, as the auto-generated one will be fine.

boecklic: Geometry fields are automatically indexed, yes: https://docs.djangoproject.com/en/3.1/ref/contrib/gis/model-api/#spatial-index `spatial_index` is an option to every `GeometryField` which is by default `True`.

app/stac_api/views.py
boecklic: We can have it as additional option, yes. What I find more intuitive in general (other than many different flags to turn certain debug msg on and off) is just having more than one (e.g. 2 or 3) debug levels. DEBUG1 is 'just enough to retrace what's happening' on prod, DEBUG2 is 'some more stuff helpful for development', DEBUG3 is 'really everything'... these messages could then go to level 3

boecklic (approved):
Thanks for the thorough and extensive work on this! It's not reflected in the number of lines of code, but the effort was worth it!

app/stac_api/views.py
hansmannj: That would make sense @boecklic. I'd suggest to do this kind of "debugging triage" in a separate PR, as probably debug prints other than only from this PR might be affected. Probably a dedicated ticket for this cleaning up and triage might make sense. So for the moment, we could merge this PR as is and re-think our debugging levels in general in a next step.
What do you think?

boecklic: Sure, no hurry on that, if agreed we can preview this for post ogd cleanup

PR #186, opened 2021-01-25
BGDIINF_SB-1571: Do not use the cache middleware for static files

ltshb:
Static files already have the cache header therefore do not overwrite
them.

### DIFF
--- a/app/middleware/cache_headers.py
+++ b/app/middleware/cache_headers.py
@@ -29,7 +29,10 @@ class CacheHeadersMiddleware:
         if request.path in ['/checker', '/get-token']:
             # never cache the /checker and /get-token endpoints.
             add_never_cache_headers(response)
-        elif request.method in ('GET', 'HEAD'):
+        elif (
+            request.method in ('GET', 'HEAD') and
+            not request.path.startswith(f'/api/stac/{settings.STAC_VERSION}/static/')
+        ):
             logger.debug(
                 "Patching cache headers for request %s %s",
                 request.method,
--- a/app/middleware/cache_headers.py
+++ b/app/middleware/cache_headers.py
@@ -1,4 +1,5 @@
 import logging
+from urllib.parse import urlparse
 
 from django.conf import settings
 from django.utils.cache import add_never_cache_headers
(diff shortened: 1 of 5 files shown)


app/middleware/cache_headers.py
```diff
@@ -29,7 +29,10 @@ def __call__(self, request):
         if request.path in ['/checker', '/get-token']:
             # never cache the /checker and /get-token endpoints.
             add_never_cache_headers(response)
-        elif request.method in ('GET', 'HEAD'):
+        elif (
+            request.method in ('GET', 'HEAD') and
+            not request.path.startswith(f'/api/stac/{settings.STAC_VERSION}/static/')
```
boecklic, line 34: you might rather want to use `STATIC_URL` from settings.py

ltkum (approved):
Seems good to me. 

I find it cleaner to have STAC_VERSION in the config, and after a quick tour of the project, I think you made sure to change all calls to serializers.STAC_VERSION to settings.STAC_VERSION.

app/middleware/cache_headers.py
boecklic: ...and what just pops into my mind: shouldn't we rather use `/api/stac/static` and omit the version info here? If we ever have to support v1.0.0 it will only affect the REST part, not the admin / static etc. (/admin is already accessible under /api/stac/static). Only issue is with spec, for which I don't have a good solution right now... (maybe /api/stac/static/v0.9.0/api.html).

boecklic (changes requested):

app/middleware/cache_headers.py
ltshb: That's a good point, but I suggest to create a new jira ticket to address this. I'll use the `STATIC_URL` as mentioned above for this PR though. But leave the change of moving the `STAC_VERSION` in settings if you don't mind @boecklic ?

boecklic (approved):

app/middleware/cache_headers.py
ltshb: Ticket for the static path: https://jira.swisstopo.ch/browse/BGDIINF_SB-1610

PR #187, opened 2021-01-27
BGDIINF_SB-1612 default prometheus-consumable app metrics

boecklic:
This PR adds django-prometheus and exposes the default app metrics defined in that package. This is intended as a first iteration to test integration of service and prometheus, get to know prometheus and it's possibilities. These findings will help to tailor exposed app metrics to our needs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -128,10 +128,13 @@ format:
 	$(ISORT) $(PYTHON_FILES)
 
 
-# make sure that the code conforms to the style guide
-# The DJANGO_SETTINGS module must be made available to pylint
-# to support e.g. string model referencec (see
-# https://github.com/PyCQA/pylint-django#usage)
+# make sure that the code conforms to the style guide. Note that
+# - the DJANGO_SETTINGS module must be made available to pylint
+#   to support e.g. string model referencec (see
+#   https://github.com/PyCQA/pylint-django#usage)
+# - export of migrations for prometheus stats must be disabled,
+#   otherwise it's attempted to connect to the db during linting
+#   (which is not available)
 .PHONY: lint
 lint:
 	@echo "Run pylint..."
--- a/app/config/settings_prod.py
+++ b/app/config/settings_prod.py
@@ -70,12 +70,17 @@ INSTALLED_APPS = [
     'solo.apps.SoloAppConfig',
     'storages',
     'whitenoise.runserver_nostatic',
+    'django_prometheus',
     'config.apps.StacAdminConfig',
     'stac_api.apps.StacApiConfig',
 ]
 
+# Middlewares are executed in order, once for the incoming
+# request top-down, once for the outgoing response bottom up
+# Note: The prometheus middlewares should always be first and
+# last, put everything else inbetween
 MIDDLEWARE = [
-    'middleware.cache_headers.CacheHeadersMiddleware',
+    'django_prometheus.middleware.PrometheusBeforeMiddleware',
     'middleware.logging.RequestResponseLoggingMiddleware',
     'django.middleware.security.SecurityMiddleware',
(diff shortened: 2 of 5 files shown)


app/config/urls.py
```diff
@@ -30,4 +30,5 @@
     urlpatterns = [
         path('__debug__/', include(debug_toolbar.urls)),
         path('tests/test_http_500', TestHttp500.as_view()),
+        path('prometheus/', include('django_prometheus.urls')),
```
ltshb, line 33: Is it on purpose that you add the prometheus endpoint only for debug ? Just as a side note, I've found out that our docker image build with --target debug suffer  from  a memory leak ! The --target production is not touched by this memory leak.

app/config/settings_prod.py
```diff
@@ -70,11 +70,17 @@
...
     'stac_api.apps.StacApiConfig',
 ]
 
+# Middlewares are executed in order, once for the incoming
+# request top-down, once for the outgoing response bottom up
+# Note: The prometheus middlewares should always be first and
+# last, put everything else inbetween
 MIDDLEWARE = [
+    'django_prometheus.middleware.PrometheusBeforeMiddleware',
     'middleware.cache_headers.CacheHeadersMiddleware',
```
boecklic, line 84: @ltshb just noted that `CacheHeadersMiddleware` is on top of the middleware list. Is this required? If not I would move it all the way down to between `ExceptionLoggingMiddleware` and `PrometheusAfterMiddleware`

ltshb (commented):
See comment below

app/config/settings_prod.py
ltshb: I don't think this is required, you can try to move it down.

rebert (approved):
LGTM

app/config/settings_prod.py
boecklic: done

app/config/urls.py
boecklic: No, of course it should not only be in debug, thanks.

app/config/settings_prod.py
```diff
@@ -262,3 +269,8 @@ def get_logging_config():
 # Timeout in seconds for call to external services, e.g. HTTP HEAD request to
 # data.geo.admin.ch/collection/item/asset to check if asset exists.
 EXTERNAL_SERVICE_TIMEOUT = 3
+
+# By default django_promtheus tracks the number of migrations
+# This causes troubles during linting, so we need to be able to
+# disable this during linting.
+PROMETHEUS_EXPORT_MIGRATIONS = os.getenv('PROMETHEUS_EXPORT_MIGRATIONS', True)
```
ltshb, line 276: Here we need to convert string to bool:

`PROMETHEUS_EXPORT_MIGRATIONS = strtobool(os.getenv('PROMETHEUS_EXPORT_MIGRATIONS', 'True'))`

Makefile
```diff
@@ -128,14 +128,17 @@ format:
...
+#   to support e.g. string model referencec (see
+#   https://github.com/PyCQA/pylint-django#usage)
+# - export of migrations for prometheus stats must be disabled,
+#   otherwise it's attempted to connect to the db during linting
+#   (which is not available)
 .PHONY: lint
 lint:
 	@echo "Run pylint..."
-	DJANGO_SETTINGS_MODULE=config.settings $(PYLINT) $(PYTHON_FILES)
+	DJANGO_SETTINGS_MODULE=config.settings PROMETHEUS_EXPORT_MIGRATIONS=false $(PYLINT) $(PYTHON_FILES)
```
ltshb, line 141: It should have any impact but simply as styling, I would use `False` instead of `false`  or `0`

PR #189, opened 2021-01-27
BGDIINF_SB-1546: Added debug log when saving/deleting models

ltshb:
Added more logs in models and improve error message in asset validation.


app/stac_api/apps.py
```diff
@@ -74,11 +74,22 @@ def get_page_size(self, request):
             )
 
         if page_size <= 0:
+            logger.error(
```
rebert, line 77: of course :-)

rebert (approved):
thanks @ltshb

PR #191, opened 2021-01-28
BGDIINF_SB-1605: Change the domain of geoadmin:variant in model

rebert:
- Space allowed
- 25 Characters instead of 15


hansmannj (approved):
:+1:
Nice, thanks for the fast implementation :smile: 
Perhaps it would make sense to adapt an existing unit test so that the geoadmin:variant is 25 characters long and contains spaces? :grimacing: :wink:

ltshb (approved):
As @hansmannj  mentioned a new unittest test case with a name with 25 character and space could be added.

ltshb (approved):

PR #192, opened 2021-01-28
BGDIINF_SB-1610: Changed static file path

ltshb:
Change the static file path from /api/stac/v0.9/static to
/api/stac/static/.

The spec goes then in /api/stac/static/spec/v0.9/.

Also harmonized the uses of os path in settings, by using the Path
library and removing the harcoded /. This would also allow using the
code on Windows server.


boecklic (approved):

rebert (approved):
A good idea to map the spec to a version

hansmannj (approved):

PR #193, opened 2021-01-28
BGDIINF_SB-1621: Set the limit of the search endpoint to max 100

rebert:


hansmannj (approved):
Thanks for quickly fixing this!

PR #194, opened 2021-01-28
BGDIINF_SB-1626: Fixed asset file validation cache issue

ltshb:
USing the HTTP HEAD request on the asset file on S3 can be dangerous on
the deployment because this request is suggest to caching. By default
the caching has an expiry time of 2 hours ! Which means that the answer
of the get HTTP HEAD might not be correct due to cache. This is quite
dangerous for validation.

Therefore now we use boto3 to check the asset file existance and to
check its checksum.

Added also a management command to upload dummy assets on S3 for testing
purpose.


ltshb:
Ok for this one I need to addapt the unittest mocking, but its now too late, I'll do it on Monday ;-)

boecklic (commented):
Nice!

`dummy_data` and `dummy_asset` could be combined by extending the `dummy_data` with a flag to upload assets as well, no?

ltshb:
> Nice!
> 
> `dummy_data` and `dummy_asset` could be combined by extending the `dummy_data` with a flag to upload assets as well, no?

@boecklic I thought about this, but we need to have it separate for the use case that I had. I wanted to upload asset files in order to test the POST request, in this case you don't want the asset metadata.
And if I'm not mistaken, the dummy_data already upload the asset file due to `SimpleUploadedFile`.

boecklic (approved):

PR #196, opened 2021-01-28
BGDIINF_SB-1624: Comment ou temporarly asset management

ltshb:
In order to avoid any bad surprise short for the go live on march and
due the performance issue when changing item geometry the asset
management on S3 is temporarly commented out. This needs to be somehow
reintroduce after the go live within BGDIINF_SB-1625

### DIFF
--- a/app/stac_api/models.py
+++ b/app/stac_api/models.py
@@ -4,7 +4,7 @@ import os
 import time
 from uuid import uuid4
 
-import botocore.exceptions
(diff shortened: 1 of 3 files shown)


hansmannj (approved):
Looks good to me, thanks!

boecklic (approved):

boecklic:
:+1: thanx for the night shift!

rebert (approved):

PR #197, opened 2021-02-01
BGDIINF_SB-1617: adding exact datetime search queries tests

ltkum:
Issue : Tests on the search endpoint lacked a testcase to check if exact
datetimes queries worked.

Fix : Created Two tests
 One checks if an exact datetime query works as intended
 One checks if an exact datetime query with a non-date string returns a
400 error as should be the case.


rebert (approved):
thanks

PR #199, opened 2021-02-01
BGDIINF_SB-1391: adding favorite icon support

ltkum:
Issue : When looking at pages from service-stac through a browser, there
was no favorite icons.

Fix : Uploaded a favorite Icon in the project in app/stac_api/static/
 In settings, told django that the /static/ directory is where to find
static content
 In urls, added an url redirection for the favicon (since it would be at
/static/favicon.ico rather than /favicon.ico as most browsers think it
is.)
 In templates, specified that the favicon is in the static directory,
for the browsers that would not follow redirections for favicons

Issue : 
  Endpoints that do not rely on templates (for example : json responses meant to be called via api_calls, such as /checker for the most abvious example) will not display a favicon. Since these are not meant to be called via a browser, I do not think it matters much but I think it's faire to mention it.


app/config/settings_prod.py
```diff
@@ -274,3 +274,7 @@ def get_logging_config():
 # By default django_promtheus tracks the number of migrations
 # This causes troubles in various places so we disable it
 PROMETHEUS_EXPORT_MIGRATIONS = False
+
+
+# Where the application goes to find its static files.
+STATIC_URL = "/static/"
```
ltshb, line 280: The `STATIC_URL` is already defined in line 178 and has been modified by  PR #192

app/stac_api/urls.py
```diff
@@ -45,4 +47,8 @@
         AssetDetail.as_view(),
         name='asset-detail'
     ),
+    path(
+        "favicon.ico",
+        RedirectView.as_view(url=staticfiles_storage.url("favicon.ico")),
+    ),
```
ltshb, line 53: This should not be needed and the static file should be served by Whitenoise.

ltshb (changes requested):
I would put the favicon in `app/stac_api/templates` as this path is already part of the static asset served by Whitenoise.

app/templates/admin/base.html
```diff
@@ -1,4 +1,7 @@
 {% extends "admin/base.html" %}
+{% load static %}
+
+<link rel="icon"href="{% static 'favicon.ico' %}">
```
boecklic, line 4: In Django it's usually a good idea to place everything inside a block/extend a block. For the favicon you can use the block `extrahead` (see https://stackoverflow.com/a/64368855/9896222)
(plus, there's a missing space between 'icon' and href)

app/templates/rest_framework/base.html
```diff
@@ -1,4 +1,7 @@
 {% extends "rest_framework/base.html" %}
+{% load static %}
+
+<link rel="icon"href="{% static 'favicon.ico' %}">
```
boecklic, line 4: dito

app/stac_api/templates/admin/ol_swisstopo.html
```diff
@@ -1,3 +1,7 @@
+{% load static %}
+
+<link rel="icon"href="{% static 'favicon.ico' %}">
```
boecklic, line 3: dito

boecklic (changes requested):

app/config/settings_prod.py
ltkum: and now we've removed it.

app/stac_api/urls.py
ltkum: I confirm this is not needed with Whitenoise.

app/templates/admin/base.html
ltkum: It's been corrected in all files.

ltshb (changes requested):
Favicon on the landing page is not working, it only works for django admin page.

ltkum:
image

@ltshb  It seems to be working for me locally. Is there anything different between your environment and mine ?

ltkum:
From what I could see, it appears in firefox but not in chromium. 
Chromium doesn't even try to get it, while firefox tries to get it (twice : once at /favicon.ico, once at /api/stac/static/favicon.ico)
The head link to favicon is indeed not there. I'll look into it.

ltkum:
From what I understand : 
I'll have to check how to add a <head> element to class based views with no templates.

ltshb:
It looks like it search for the favicon at the wrong place, the browser does a GET	http://127.0.0.1:8000/favicon.ico ! I used firefox  in private mode, but I don't think this is related to firefox.

ltshb (approved):
:+1:

boecklic (approved):
:+1:

rebert (commented):
@ltkum nice. I wanted to have an eye on the icon, but did not find it. Where is the famous icon located?

rebert (approved):

PR #201, opened 2021-02-01
BGDIINF_SB-1630: Faster update of aggregated collection bbox

rebert:
Now the update of the bbox of the collections is also practical

when having 10K of items.

### DIFF
--- a/app/stac_api/collection_spatial_extent.py
+++ b/app/stac_api/collection_spatial_extent.py
@@ -64,9 +64,6 @@ class CollectionSpatialExtentMixin():
             if trigger == 'update' and geometry != original_geometry:
                 # is the new bbox larger than (and covering) the existing
                 if Polygon.from_bbox(GEOSGeometry(geometry).extent).covers(self.extent_geometry):
-                    # pylint: disable=fixme
-                    # TODO: cover this code by a unittest, remove this comment when BGDIINF_SB-1595
-                    # is implemented
                     logger.info(
                         'Updating collections extent_geometry with item geometry changed '
                         'from %s to %s, (larger and covering bbox)',
--- a/app/stac_api/collection_spatial_extent.py
+++ b/app/stac_api/collection_spatial_extent.py
@@ -91,12 +91,18 @@ class CollectionSpatialExtentMixin():
                         },
                     )
                     start = time.time()
-                    qs = type(item).objects.filter(collection_id=self.pk
-                                                  ).exclude(id=item.pk
-                                                           ).aggregate(Extent('geometry'))
+                    raw_sql = '''
+                    SELECT
+                    1 AS id,
+                    st_AsText(st_extent(geometry)) AS geometry__extent
+                    FROM stac_api_item
+                    WHERE collection_id = %d
+                    AND NOT id = %d
+                    ''' % (self.pk, item.pk)
+                    qs = type(item).objects.raw(raw_sql)
                     union_geometry = GEOSGeometry(geometry)
                     self.extent_geometry = union_geometry.union(
-                        Polygon.from_bbox(qs['geometry__extent'])
+                        GEOSGeometry(qs[0].geometry__extent)
                     )
                     logger.info(
                         'Collection extent_geometry updated to %s in %ss, after item update',
@@ -119,10 +125,17 @@ class CollectionSpatialExtentMixin():
                     },
                 )
                 start = time.time()
-                qs = type(item).objects.filter(collection_id=self.pk
-                                              ).exclude(id=item.pk).aggregate(Extent('geometry'))
-                if bool(qs['geometry__extent']):
-                    self.extent_geometry = Polygon.from_bbox(qs['geometry__extent'])
+                raw_sql = '''
+                    SELECT
+                    1 AS id,
+                    st_AsText(st_extent(geometry)) AS geometry__extent
+                    FROM stac_api_item
+                    WHERE collection_id = %d
+                    AND NOT id = %d
+                    ''' % (self.pk, item.pk)
+                qs = type(item).objects.raw(raw_sql)
+                if bool(qs[0].geometry__extent):
+                    self.extent_geometry = GEOSGeometry(qs[0].geometry__extent)
                 else:
                     self.extent_geometry = None
                 logger.info(
--- a/app/stac_api/collection_spatial_extent.py
+++ b/app/stac_api/collection_spatial_extent.py
@@ -1,7 +1,6 @@
 import logging
 import time
 
-from django.contrib.gis.db.models import Extent
 from django.contrib.gis.geos import GEOSGeometry
 from django.contrib.gis.geos import Polygon
 from django.contrib.gis.geos.error import GEOSException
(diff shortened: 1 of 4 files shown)


ltshb (approved):
:1st_place_medal:

ltclm:
@rebert is django using st_extent now?

```sql
stopo_master
select count(*) FROM grid_1km; 					-- 144826 objekte
EXPLAIN ANALYZE select st_union(the_geom) FROM grid_1km;	-- Execution Time: 35399.854 ms
EXPLAIN ANALYZE select st_extent(the_geom) FROM grid_1km; 	-- Execution Time: 24.653 ms
```

rebert:
@ltclm when I grep the sql logs I don't catch st_extent... But I guess with this change the calculation for the aggregated extent is done by the database using its potentail as spatial index... what is already far better than doing the union in a python loop... But I guess that there is even more optimisation potential...

ltclm:
i was able to isolate the postgres queries with docker-compose logs --follow
i have populated 40'000 dummy extents and then tried to edit the perimeter of an item at the edge of the collection.

When i change the geometry of an item at the edge of the collection, the new extent of the collection is wrong:
```sql
-- save item at the edge http://localhost:8000/api/stac/admin/stac_api/item/46127/change/
SELECT ST_Extent("stac_api_item"."geometry") AS "geometry__extent" FROM "stac_api_item" WHERE ("stac_api_item"."collection_id" = 34 AND NOT ("stac_api_item"."id" = 46127))
-- update extent of collection
UPDATE "stac_api_collection" 
SET 
	"name" = 'perftest-collection-0'
	, "created" = '2021-02-01T21:16:34.476031+00:00'::timestamptz
	, "updated" = '2021-02-02T09:42:59.004413+00:00'::timestamptz
	, "description" = 'This is a description'
	, "extent_geometry" = ST_GeomFromEWKB('\x0103000020e61000000100000008000000976b9643dc2a1c403020f6fd8b6e4740d65a1dbe83461940d5b29cd8786f4740976b9643dc2a1c405b69cd33b16f4740976b9643dc2a1c406da007bcbdb6474027add1c66bc522406da007bcbdb6474027add1c66bc52240b79562553e274740976b9643dc2a1c40b79562553e274740976b9643dc2a1c403020f6fd8b6e4740'::bytea)
	, "extent_start_datetime" = '1970-01-01T03:22:10+00:00'::timestamptz
	, "extent_end_datetime" = '2020-12-31T23:01:10+00:00'::timestamptz
	, "license" = 'test'
	, "summaries" = '{"eo:gsd": [2, 2.5, 5, 10], "proj:epsg": [2056, 4326, 21781], "geoadmin:variant": ["var1", "var2", "var3"]}'
	, "title" = 'Test title'
	, "etag" = '8d82a14c-c985-4206-aacc-6a4c4c939b83'
WHERE "stac_api_collection"."id" = 34
-- update geometry of item
UPDATE "stac_api_item" SET "name" = 'perftest-item-37120', "collection_id" = 34, "geometry" = ST_GeomFromEWKB('\x0103000020e6100000010000000600000075dd836cc52e1c4053afaabd8a6e4740d65a1dbe83461940d5b29cd8786f474007f2f925b42e1c400b76b27eb16f4740c06cf3a2223c1c40b4b68ef3b26f4740edc34056333c1c40dd1677328c6e474075dd836cc52e1c4053afaabd8a6e4740'::bytea), "created" = '2021-02-01T21:55:58.058379+00:00'::timestamptz, "updated" = '2021-02-02T09:42:59.055450+00:00'::timestamptz, "properties_datetime" = '1978-10-29T02:51:32+00:00'::timestamptz, "properties_start_datetime" = NULL, "properties_end_datetime" = NULL, "properties_title" = 'My Title', "etag" = 'cf1849b4-83d7-4e39-be06-85bdb2e30ec0' WHERE "stac_api_item"."id" = 46127

```

This is the new geometry of the item with id ``46127``
```sql
SELECT * from stac_api_item WHERE collection_id = 34 AND id = 46127
```
image

And this is the new extent of the collection
```sql
SELECT * from stac_api_item WHERE collection_id = 34 AND id = 46127
```

image

As you can see django is not writing an extent / bbox back to the collection table. He's making a union of the old extent with the new geometry and writing this as geometry back to the table.

im not familier with the model / specification of the collection but i think there we should only have bboxes instead of geometries.

ltclm:
@rebert  just tested the new commit, the update is faster but the resulting collection geometry is still wrong. should be converted to bbox / envelope.

rebert:
@ltclm I will:
- add some more unitests
- make a refactoring jjira ticket to use aggregated bbox on the fly from db

ltclm:
the collections geometry is now a bbox, thank you!

app/tests/test_item_to_collection_endpoint.py
```diff
@@ -0,0 +1,197 @@
...
+            self.collection, sample='item-paris', db_create=True
+        ).model
+        response_collection = self.client.get(f"/{API_BASE}/collections/{collection_name}")
+        bbox_collection_paris = response_collection.json()['extent']['spatial']['bbox'][0]
+
+        self.assertNotEqual(
+            self._round_list(bbox_collection_ch), self._round_list(bbox_collection_paris)
+        )
+
+        polygon_ch = Polygon.from_bbox(bbox_collection_ch)
```
ltshb, line 182: The order of the test is defined by alphabetic order, the DB is not necessary reseted between call because the DB is set with the setUpTestData() which is called only once before starting the test of the class. For such test case we should use setUp() method to create the data in DB. This way the DB is reseted before each test.

ltshb (commented):
See inline comment

app/stac_api/collection_spatial_extent.py
```diff
@@ -90,14 +87,22 @@ def update_bbox_extent(self, trigger, geometry, original_geometry, item):
...
                     start = time.time()
-                    qs = type(item).objects.filter(collection_id=self.pk).exclude(id=item.pk)
+                    raw_sql = '''
+                    SELECT
+                    1 AS id,
+                    st_AsText(st_extent(geometry)) AS geometry__extent
+                    FROM stac_api_item
+                    WHERE collection_id = %d
+                    AND NOT id = %d
+                    ''' % (self.pk, item.pk)
```
boecklic, line 97: Nice solution! You think you can manage to get rid of the raw sql by making use of https://docs.djangoproject.com/en/3.1/ref/models/expressions/#django.db.models.Func? Would be great if we could avoid raw_sql as much as possible

rebert:
@boecklic yes, I guess. @ltclm shared a solution with me yesterday. Something like:
`q = Collection.objects.annotate(num_item=Extent('item__geometry'))
q[2]`

ltclm:
@boecklic @rebert 
yes you can aggregate the geometries without raw sql and really good performance of linked tables:

```python
    def extent_geometry_on_the_fly(self):                                       
        from django.contrib.gis.db.models import Extent                         
        from django.contrib.gis.geos import GEOSGeometry, Polygon               
                                                                                
        "Returns the bbox on-the-fly"                                           
        bbox = Collection.objects.filter(id=self.id).aggregate(Extent('item__geometry')
                                                              )['item__geometry__extent']
                                                                                
        return GEOSGeometry(Polygon.from_bbox(bbox), srid=4326) 
```

which is triggering the following sql command:

```SQL
SELECT 
"stac_api_collection"."id"
, "stac_api_collection"."name"
, "stac_api_collection"."created"
, "stac_api_collection"."updated"
, "stac_api_collection"."description"
, "stac_api_collection"."extent_geometry"
, "stac_api_collection"."extent_start_datetime"
, "stac_api_collection"."extent_end_datetime"
, "stac_api_collection"."license"
, "stac_api_collection"."summaries"
, "stac_api_collection"."title"
, "stac_api_collection"."etag"
, st_setsrid(ST_Extent("stac_api_item"."geometry")::geometry,4326) AS "num_item" 
FROM "stac_api_collection" LEFT OUTER JOIN "stac_api_item" 
ON ("stac_api_collection"."id" = "stac_api_item"."collection_id") 
GROUP BY "stac_api_collection"."id" LIMIT 1 OFFSET 2

-- 35ms with 40000 features
```
I would recommend you to completely get rid of the statification of the bbox in the collection table and avoid unnecessarily complex code as much as  possible. **That's exactly what spatial indexes are for**! postgis is able to calculate the bbox of >100k objects in less than 50ms.

ltshb:
I like the solution of @ltclm I proposed to do it in a new PR which reference this one :-)

PR #203, opened 2021-02-01
BGDIINF_SB-1626: make sure that the required read only fields are checked in unittest

ltshb:
Unittest improvements:

- Make sure that in case of an expected HTTP 400, the error message is the one we expect. HTTP 400 have many different reason that's why we should test the reason as well in unittest
- Make sure also that STAC required read only fields are present in unittest.

### DIFF
--- a/app/tests/base_test.py
+++ b/app/tests/base_test.py
@@ -114,6 +114,7 @@ class StacBaseTestCase(TestCase):
         if ignore is None:
             ignore = []
         self._check_stac_dictsubset('collection', expected, current, ignore)
+        # check required fields
         for key, value in [
             ('stac_version', '0.9.0'),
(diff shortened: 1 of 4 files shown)


app/stac_api/management/commands/dummy_data.py
```diff
@@ -112,7 +112,7 @@ def handle(self, *args, **options):
                         media_type = random.choice(MEDIA_TYPES)
                         asset, _ = Asset.objects.get_or_create(
                             item=item,
-                            name=f'{asset_id}{media_type[2]}',
+                            name=f'{asset_id}{random.choice(media_type[2])}',
```
rebert, line 115: nice

rebert (approved):
Thanks

PR #204, opened 2021-02-01
adding local / Pod IP to ALLOWED_HOSTS

boecklic:
The localhost IP / IP of the Pod must be in `ALLOWED_HOSTS` to allow request from local without host header. This is the case e.g. for prometheus (in principle it's the same issue with the health check)

### DIFF
--- a/app/config/settings_prod.py
+++ b/app/config/settings_prod.py
@@ -45,7 +45,10 @@ DEBUG = False
 # https://docs.djangoproject.com/en/dev/ref/settings/#secure-proxy-ssl-header
 SECURE_PROXY_SSL_HEADER = ('HTTP_CLOUDFRONT_FORWARDED_PROTO', 'https')
 
-ALLOWED_HOSTS = []
+# We need to have the IP of the Pod/localhost in ALLOWED_HOSTS
+# as well to be able to scrape prometheus /metrics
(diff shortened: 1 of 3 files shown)


app/config/settings_prod.py
```diff
@@ -45,7 +45,9 @@
 # https://docs.djangoproject.com/en/dev/ref/settings/#secure-proxy-ssl-header
 SECURE_PROXY_SSL_HEADER = ('HTTP_CLOUDFRONT_FORWARDED_PROTO', 'https')
 
-ALLOWED_HOSTS = []
+# We need to have the IP of the Pod/localhost in ALLOWED_HOSTS
+# as well to be able to scrape prometheus /metrics
+ALLOWED_HOSTS = [os.getenv('THIS_POD_IP', '127.0.0.1')]
```
ltshb, line 50: Why not adding this ip directly in the ALLOWED_HOSTS environment variable ?

ltshb (commented):
See comment ? If you still decide to keep the new environment variable, I would document it in the readme.md like the others.

app/config/settings_prod.py
boecklic: added it to the readme

ltshb (approved):

PR #205, opened 2021-02-02
Improvement of management command and logging.

ltshb:
Now the dummy_data command supports to run in parallel to speed up
things and also support to use a collection names list as parameter in
order to create different collections with different numbers of items.

The general command output has been improved.

Now the logging of management command is redirected automatically to log file like for test.


app/stac_api/management/commands/dummy_data.py
```diff
@@ -73,59 +213,24 @@ def add_arguments(self, parser):
...
         )
 
+        parser.add_argument(
+            '--parallel-collections',
+            type=int,
+            default=1,
+            help="Number of collection created in parallel (default 1)"
+        )
+
+        parser.add_argument(
```
rebert, line 223: nice. but when I run it locally with the default settings, I get:
`postgres@service_stac_local FATAL:  sorry, too many clients already`

```
service_stac_local=# SHOW max_connections;
 max_connections 
-----------------
 100
(1 row)

service_stac_local=# select count(*) from pg_stat_activity;
 count 
-------
   105
(1 row)
```
As a **default** we either have to augment the number of max_connections or to run less in parallel

rebert (approved):
thanks for the improvements - just yesterday I realized, that it takes some time to populate dummy data. Just a comment about reducing the default of the parallelization

app/stac_api/management/commands/dummy_data.py
ltshb: Strange, the defaults are 1 collection and 10 items in parallel which means 1 * 10 connections. This should work. Are you that you don't have left over connection trying with a bigger number. I had the case where the connection where not properly terminated when using big number in parallel options. NOTE that the --collection-parallel and --item-parallel needs to be multiplied to get the number of DB connections (10 collection and 10 items in parallel => 100 connection !)

rebert: to get the numbers, I stopped docker-compose, started it again and made app/manage.py dummy_data populate ... now I have 6 "maintenance" connections open ...

ltshb:
> thanks for the improvements - just yesterday I realized, that it takes some time to populate dummy data. Just a comment about reducing the default of the parallelization

What do you suggest for the default parallel value ?

boecklic (approved):
:+1:

PR #206, opened 2021-02-04
BGDIINF_SB-1630: Avoid raw sql in bbox from item to collection

rebert:


app/stac_api/collection_spatial_extent.py
```diff
@@ -87,19 +88,21 @@ def update_bbox_extent(self, trigger, geometry, original_geometry, item):
...
-                    qs = type(item).objects.raw(raw_sql)
-                    union_geometry = GEOSGeometry(geometry)
-                    self.extent_geometry = Polygon.from_bbox(
-                        union_geometry.union(GEOSGeometry(qs[0].geometry__extent)).extent
-                    )
+                    qs_bbox_other_items = type(item).objects.filter(collection_id=self.pk
+                                                                   ).exclude(id=item.pk).aggregate(
+                                                                       Extent('geometry')
+                                                                   )
+                    geometry_updated_item = GEOSGeometry(geometry)
```
ltshb, line 95: Make sure to run yapf. It doesn't seem that yapf would leave such formatting but I might be wrong.

ltshb (approved):
Just make sure to run yapf before merging, see my inline comment

PR #207, opened 2021-02-04
BGDIINF_SB-1638: Profiling tool

ltshb:
Added 2 django command line for profiling `ItemSerializer.data` and the the `paginator.paginate_queryset(qs, request)` method. Also added a decorator to profile any method. Profiling should only be used for debugging/testing as it adds a lot of overhead.


boecklic (approved):

hansmannj (approved):
LGTM

PR #208, opened 2021-02-04
BGDIINF_SB-1589: Correction in Spec POST / Search

rebert:
Added bbox in the examples.

### DIFF
--- a/spec/components/schemas.yml
+++ b/spec/components/schemas.yml
@@ -110,10 +110,51 @@ components:
       minItems: 4
       type: array
       readOnly: true
+    bboxfilter:
+      description: >-
+        Only features that have a geometry that intersects the bounding box are selected.
+        The bounding box is provided as four numbers:
+
+
+        * Lower left corner, coordinate axis 1
+
+        * Lower left corner, coordinate axis 2
+
+        * Upper right corner, coordinate axis 1
+
+        * Upper right corner, coordinate axis 2
+
+
+        The coordinate reference system of the values is WGS84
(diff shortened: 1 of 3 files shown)


spec/components/schemas.yml
```diff
@@ -113,7 +113,6 @@ components:
       maxItems: 4
       minItems: 4
       type: array
-      readOnly: true
```
boecklic, line 116: hmm, shouldn't this stay readonly, since e.g. in https://service-stac.dev.bgdi.ch/api/stac/static/spec/v0.9/apitransactional.html#operation/getFeature it's readonly.
shouldn't `/search` path rather reference `bboxFilter` schema?

boecklic (changes requested):

spec/components/schemas.yml
hansmannj: Yes, the point in that ticket is, that here: 
https://service-stac.dev.bgdi.ch/api/stac/static/spec/v0.9/apitransactional.html#operation/postSearchSTAC
there is no `bbox` listed in the `REQUEST BODY SCHEMA` table on the left, but on the right, in the `Request samples` there's `bbox` in the sample payload.

hansmannj: Sorry for my earlier comment here, it was too early obviously. These current changes seem to work.
Currently there's already a reference to `"#/components/schemas/searchBody"` in the `requestBody` of the `/search` `post` part. And the `searchBody` already references the `bboxFilter`. Not sure, why it does not appear, although it should..

boecklic (approved):

hansmannj (approved):
Nice :+1:

PR #212, opened 2021-02-05
ISSUE BGDIINF_SB_1631: wrong status on items and assets queries

ltkum:
Issue : When trying to access the items from a non existing collection,
or when trying to access the assets from one non existing item, the
application would return a 200 status with an empty list of items or
assets.

Fix : using the get_object_or_404() method allows consistent comportment
with other endpoints which returned a correct 404 not found status.

### DIFF
--- a/app/stac_api/views.py
+++ b/app/stac_api/views.py
@@ -180,9 +180,10 @@ class ItemsList(generics.GenericAPIView, views_mixins.CreateModelMixin):
         return queryset
 
     def list(self, request, *args, **kwargs):
-        queryset = self.filter_queryset(self.get_queryset())
         if not Collection.objects.filter(name=self.kwargs['collection_name']).exists():
+            logger.error("The collection %s does not exist", self.kwargs['collection_name'])
             raise Http404("This collection does not exists.")
+        queryset = self.filter_queryset(self.get_queryset())
         page = self.paginate_queryset(queryset)
         if page is not None:
             serializer = self.get_serializer(page, many=True)
@@ -363,10 +364,13 @@ class AssetsList(generics.GenericAPIView, views_mixins.CreateModelMixin):
         )
 
     def get(self, request, *args, **kwargs):
-        queryset = self.filter_queryset(self.get_queryset())
-        if not Collection.objects.filter(name=self.kwargs['collection_name']).exists() or \
-                not Item.objects.filter(name=self.kwargs['item_name']).exists():
+        if (not Collection.objects.filter(name=self.kwargs['collection_name']).exists() or
+                not Item.objects.filter(name=self.kwargs['item_name']).exists()):
+            logger.error("Either the collection (%s) or the item (%s) do not exist",
+                         self.kwargs['collection_name'], self.kwargs['item_name'])
             raise Http404("This collection does not exists.")
+
+        queryset = self.filter_queryset(self.get_queryset())
         page = self.paginate_queryset(queryset)
         if page is not None:
             serializer = self.get_serializer(page, many=True)
--- a/app/stac_api/views.py
+++ b/app/stac_api/views.py
@@ -4,9 +4,9 @@ from collections import OrderedDict
 from datetime import datetime
 
 from django.conf import settings
+from django.http import Http404
 from django.http import JsonResponse
 from django.shortcuts import get_object_or_404
-from django.http import Http404
 
 from rest_framework import generics
 from rest_framework import mixins
@@ -181,8 +181,8 @@ class ItemsList(generics.GenericAPIView, views_mixins.CreateModelMixin):
 
     def list(self, request, *args, **kwargs):
         if not Collection.objects.filter(name=self.kwargs['collection_name']).exists():
-            logger.error("The collection %s does not exist", self.kwargs['collection_name'])
-            raise Http404("This collection does not exists.")
+            logger.error("The collection, %s, does not exist", self.kwargs['collection_name'])
+            raise Http404(f"This collection, {self.kwargs['collection_name']}, does not exists.")
         queryset = self.filter_queryset(self.get_queryset())
         page = self.paginate_queryset(queryset)
         if page is not None:
@@ -364,11 +364,19 @@ class AssetsList(generics.GenericAPIView, views_mixins.CreateModelMixin):
         )
 
     def get(self, request, *args, **kwargs):
-        if (not Collection.objects.filter(name=self.kwargs['collection_name']).exists() or
-                not Item.objects.filter(name=self.kwargs['item_name']).exists()):
-            logger.error("Either the collection (%s) or the item (%s) do not exist",
-                         self.kwargs['collection_name'], self.kwargs['item_name'])
+        if not Collection.objects.filter(name=self.kwargs['collection_name']).exists():
+            logger.error("The collection, %s, does not exist", self.kwargs['collection_name'])
             raise Http404("This collection does not exists.")
+        if not Item.objects.filter(name=self.kwargs['item_name']).exists():
+            logger.error(
+                "The item, %s, is not part of the collection, %s",
+                self.kwargs['item_name'],
+                self.kwargs['collection_name']
+            )
+            raise Http404(
+                f"The item {self.kwargs['item_name']} is not part of the collection "
+                f"{self.kwargs['collection_name']}"
+            )
 
         queryset = self.filter_queryset(self.get_queryset())
         page = self.paginate_queryset(queryset)
--- a/app/stac_api/views.py
+++ b/app/stac_api/views.py
@@ -6,7 +6,6 @@ from datetime import datetime
 from django.conf import settings
 from django.http import JsonResponse
 from django.shortcuts import get_object_or_404
-
 from rest_framework import generics
 from rest_framework import mixins
 from rest_framework.permissions import AllowAny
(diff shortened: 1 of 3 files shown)


ltkum:
I have a problem : 

To solve this issue, I simply used the same function which is used in other endpoints : get_object_or_404() (from django shortcuts)

My problem is, now tests are not passing, stating the following assetion error : 
(this is an example from `ERROR: test_items_endpoint (tests.test_items_endpoint.ItemsReadEndpointTestCase)`
`stac_api.models.Item.MultipleObjectsReturned: get() returned more than one Item -- it returned 2! `(sometimes, the number change)

My problem is the following : 
I thought, maybe the 404 method adds something to the query set, so I tried to capture the resulting queryset of `http://localhost:8000/api/stac/v0.9/collections/ch.swisstopo.swisstlmregio/items` with and without the function call. The only difference I found is the timestamp (which is expected, since both queries are not simultaneous). I do not know why this happens. Does anyone have a clue ? 

@boecklic , @ltshb  ?

ltkum:
```
[2021-02-05T17:13:59.512115Z] ERROR    - stac_api.validators_serializer : Failed to retrieve S3 object collection-1/item-1/asset-1 metadata: An error occurred (404) when calling the HeadObject operation: Not Found
[2021-02-05T17:13:59.512246Z] ERROR    - stac_api.validators_serializer : Asset at href http://testserver/collection-1/item-1/asset-1 doesn't exists
```
I found some clues.

ltshb:
You cannot use the `get_object_or_404()` method this one retrieve and return one object from the DB ! Here we need to retrieve a list of objects, so `get_list_or_404()` would be more correct but we can neither use it due to pagination, if we use it before pagination then it might return a huge list (list with several hundreds of thousands objects) and be extremely inefficient. If we use it after the pagination it will neither works because pagination already returns the limited list of objects. We can't neither check for empty `page` after pagination because an empty items list in a collection is allowed. So before pagination, meaning before evaluating the queryset which takes time, we need to check if the collection exists without retrieving it. For this we need to check in django what's the most efficient way of checking existence without returning any data, this check should be then added in the top of the `get_queryset()` method in order to stop the processing of the request as soon as possible.

boecklic:
You can use `.exists()` on a queryset, so here it would be `Collection.objects.filter(some_condition).exists()`. I'm not sure if doing it in the queryset is a good idea, is this check really necessary every time the `get_queryset` is performed? Is it a good idea to 'hide' this logic instead of explicitly check it in the views (risking some copy-paste, but logic is explicit)

ltshb:
> You can use `.exists()` on a queryset, so here it would be `Collection.objects.filter(some_condition).exists()`. I'm not sure if doing it in the queryset is a good idea, is this check really necessary every time the `get_queryset` is performed? Is it a good idea to 'hide' this logic instead of explicitly check it in the views (risking some copy-paste, but logic is explicit)

You're right it might be a better idea to do the check at the beginning of the view method.

ltkum:
> You can use `.exists()` on a queryset, so here it would be `Collection.objects.filter(some_condition).exists()`. I'm not sure if doing it in the queryset is a good idea, is this check really necessary every time the `get_queryset` is performed? Is it a good idea to 'hide' this logic instead of explicitly check it in the views (risking some copy-paste, but logic is explicit)

I did not know we could use a exists() method on querysets, this did the trick and I put this where I initially put my get object or 404 method (right after getting the queryset in the view). It seemed to work and tests are happy. Thanks for the information.

ltshb:
> > You can use `.exists()` on a queryset, so here it would be `Collection.objects.filter(some_condition).exists()`. I'm not sure if doing it in the queryset is a good idea, is this check really necessary every time the `get_queryset` is performed? Is it a good idea to 'hide' this logic instead of explicitly check it in the views (risking some copy-paste, but logic is explicit)
> 
> I did not know we could use a exists() method on querysets, this did the trick and I put this where I initially put my get object or 404 method (right after getting the queryset in the view). It seemed to work and tests are happy. Thanks for the information.

Be aware that it is not the queryset that must exists ! For example for the `collections/COLLECTION_NAME/items` endpoint the items queryset might not exists, ndlr a collection without any items. But before guetting the queryset, you must make sure that the collection exists, `Collection.objects.filter(name=COLLECTION_NAME).exists()`. For the assets endpoint there we need to make sure that the item exists.

ltshb:
I also suggest that you start by adding unittest for those case; 
- GET api/stac/v0.9/collections/non-existing-collection/items
- GET api/stac/v0.9/collections/existing-collection/items/non-existing-item/assets
- GET api/stac/v0.9/collections/non-existing-collection/items/non-existing-item/assets

app/stac_api/views.py
```diff
@@ -181,7 +181,8 @@ def get_queryset(self):
 
     def list(self, request, *args, **kwargs):
         queryset = self.filter_queryset(self.get_queryset())
-
+        if not Collection.objects.filter(name=self.kwargs['collection_name']).exists():
+            raise Http404("This collection does not exists.")
```
ltshb, line 185: This should be moved on top of the method to avoid doing unnecessary queryset. Also add logger.error() message and I would also add the collection name to both the logger message and Http404 message.

app/stac_api/views.py
```diff
@@ -363,7 +364,9 @@ def get_queryset(self):
 
     def get(self, request, *args, **kwargs):
         queryset = self.filter_queryset(self.get_queryset())
-
+        if not Collection.objects.filter(name=self.kwargs['collection_name']).exists() or \
+                not Item.objects.filter(name=self.kwargs['item_name']).exists():
+            raise Http404("This collection does not exists.")
```
ltshb, line 369: Same as above. Also from code style using `()` instead of `\` make the code cleaner in my opinion.

ltshb (changes requested):
See comments inline

app/stac_api/views.py
```diff
@@ -180,9 +180,10 @@ def get_queryset(self):
         return queryset
 
     def list(self, request, *args, **kwargs):
-        queryset = self.filter_queryset(self.get_queryset())
         if not Collection.objects.filter(name=self.kwargs['collection_name']).exists():
+            logger.error("The collection %s does not exist", self.kwargs['collection_name'])
             raise Http404("This collection does not exists.")
```
ltshb, line 185: Name of the collection that don't exist is missing in message.

app/stac_api/views.py
```diff
@@ -363,10 +364,13 @@ def get_queryset(self):
...
 
     def get(self, request, *args, **kwargs):
-        queryset = self.filter_queryset(self.get_queryset())
-        if not Collection.objects.filter(name=self.kwargs['collection_name']).exists() or \
-                not Item.objects.filter(name=self.kwargs['item_name']).exists():
+        if (not Collection.objects.filter(name=self.kwargs['collection_name']).exists() or
+                not Item.objects.filter(name=self.kwargs['item_name']).exists()):
+            logger.error("Either the collection (%s) or the item (%s) do not exist",
+                         self.kwargs['collection_name'], self.kwargs['item_name'])
             raise Http404("This collection does not exists.")
```
ltshb: Formatting issue and the message of the Http404 is not always correct, I mean it is either the collection and/or the item that doesn't exist. Actually here it would be better to have two separate if with each one the correct error message. Don't forget these error messages are sent to customer with the 404.

ltshb (commented):
See comment inlines

app/stac_api/views.py
```diff
@@ -364,11 +364,19 @@ def get_queryset(self):
         )
 
     def get(self, request, *args, **kwargs):
-        if (not Collection.objects.filter(name=self.kwargs['collection_name']).exists() or
-                not Item.objects.filter(name=self.kwargs['item_name']).exists()):
-            logger.error("Either the collection (%s) or the item (%s) do not exist",
-                         self.kwargs['collection_name'], self.kwargs['item_name'])
+        if not Collection.objects.filter(name=self.kwargs['collection_name']).exists():
+            logger.error("The collection, %s, does not exist", self.kwargs['collection_name'])
             raise Http404("This collection does not exists.")
```
ltshb, line 369: Sorry to be very picky but as these error messages go to customer, here I would also add the collection name as above. This improves the user experience, if the user does a typo, he can directly see it in the error message. So I suggest to use something like: `The collection {self.kwargs['collection_name']} does not exists.`

app/stac_api/views.py
```diff
@@ -181,8 +181,8 @@ def get_queryset(self):
 
     def list(self, request, *args, **kwargs):
         if not Collection.objects.filter(name=self.kwargs['collection_name']).exists():
-            logger.error("The collection %s does not exist", self.kwargs['collection_name'])
-            raise Http404("This collection does not exists.")
+            logger.error("The collection, %s, does not exist", self.kwargs['collection_name'])
+            raise Http404(f"This collection, {self.kwargs['collection_name']}, does not exists.")
```
ltshb, line 185: `The collection {self.kwargs['collection_name']} does not exists.` seems to me more appropriate.

ltshb (commented):
Sorry one last remarks see inline comments

app/stac_api/views.py
ltkum: No problem. Next correction incoming.

ltshb (approved):
:+1:  perfect

PR #213, opened 2021-02-11
BGDIINF_SB-1659 bugfix in search endpoint (POST and GET)

hansmannj:
BGDIINF_SB-1659:
Filtering for `updated` or `created` dates did not work both in GET and POST requests on the search endpoint.
It yielded an error message:
FieldError "Cannot resolve keyword 'properties_updated' into field" 

Internally the query appended the text 'properties_' before the 'updated' or 'created'. Hence `filter_by_query()` in `managers.py` was adapted.
Also new unit tests were added that test POST and GET requests filtering for 'updated' and 'created'.

BGDIINF_SB-1657:
Incorrectly spelled or arbitrary query parameters were accepted and did not yield error messages currently.
So the allowed query parameters were limited to `bbox, collections, cursor, datetime, ids, intersects, limit and query` and validated. Any other query parameter or incorrectly spelled query parameters will yield a 400.
Unit tests covering cases with (multiple) wrong query parameters were added.

### DIFF
--- a/app/stac_api/managers.py
+++ b/app/stac_api/managers.py
@@ -215,12 +215,16 @@ class ItemQuerySet(models.QuerySet):
             for operator in query[attribute]:
                 value = query[attribute][operator]  # get the values given by the operator
 
+                if attribute in ["updated", "created"]:
+                    prefix = ""
+                else:
+                    prefix = "properties_"
+
                 # __eq does not exist, but = does it as well
                 if operator == 'eq':
-                    query_filter = f"properties_{attribute}"
+                    query_filter = f"{prefix}{attribute}"
                 else:
-                    query_filter = f"properties_{attribute}__{operator.lower()}"
-
+                    query_filter = f"{prefix}{attribute}__{operator.lower()}"
                 return self.filter(**{query_filter: value})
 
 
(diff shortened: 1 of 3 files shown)


app/stac_api/managers.py
```diff
@@ -215,12 +215,16 @@ def filter_by_query(self, query):
...
+
                 # __eq does not exist, but = does it as well
                 if operator == 'eq':
-                    query_filter = f"properties_{attribute}"
+                    query_filter = f"{prefix}{attribute}"
                 else:
-                    query_filter = f"properties_{attribute}__{operator.lower()}"
-
+                    query_filter = f"{prefix}{attribute}__{operator.lower()}"
                 return self.filter(**{query_filter: value})
```
boecklic, line 228: just an idea: would it make code simpler by removing the `properties_` prefix in the model from the other parameters? then no special treatment would be necessary here

hansmannj: Probably it would, yes. A global search for "properties" in the code yields more than 600 hits, so it might take some time to figure out, where it needs to be replaced and what influences that has, though.

boecklic: I see ...

boecklic (approved):

PR #214, opened 2021-02-11
BGDIINF_SB-1660 remove datetime from queriable fields

boecklic:
This PR contains the following changes:

- remove `datetime` from queryable fields in GET and POST /search (it's redundant with the `datetime` query parameter)
- spec: remove `query` parameter in GET /search
- spec: cleanup in GET/POST /search


hansmannj (approved):
:+1:
Just one minor thing: when checking the new specs locally, the specification of the queriable fields for POST on /search endpoint using `query` parameter is not yet defined.
(`created`, `udated` and `title`).

PR #215, opened 2021-02-12
BGDIINF_SB-1318 : apply official CSS to service Stac API spec pages

ltkum:
Issue :   It was asked to use the official confederation CSS in the API spec
documentation.
  Using the swiss confederation logo, as well as links to the geoportal
and the legal department of swisstopo was also within the tasks.

Fix  : we imported only the 'build' directory from the official
swiss styling github : https://github.com/swiss/styleguide. We also
added a link segment which gives the location of the css. The location
of the favicon was also changed to use the one from the styleguide.
We overloaded some styles directly in the files to allow the menus to work.


hansmannj:
Hi Martin, did you already come along this one: https://github.com/Redocly/redoc/issues/241

ltkum:
Hi. 
I went around the problem and have set up markups around Redoc rather than trying to change the css that would be specific to Redoc. I'll look into it if we need to change Redoc's style. Thanks for the link.

ltkum:
Testlink : https://service-stac.dev.bgdi.ch/api/stac/static/spec/v0.9/api.html 

I think we need a small margin on the left to show the menu / search bar entirely. But you're welcome to check if something else seems out of place.

p1d1d1:
@ltkum the left menu doesn't work, e.g. the entries don't open/close

ltkum:
@p1d1d1 The menu is self generated by Redoc (understand : I kinda have no control over it), and does lead you to the differents sections, it's not supposed to open or close. I know the arrows make it confusing, but the menu does what it's supposed to do.

(again, tested it on both firefox and chromium)

ltkum:
@p1d1d1 

From what I could test : 

On chromium, the menu behaves strangely : 
When you click on an item the first time, it doesn't react, the second time it will lead you to the corresponding section but will shift to the left. Clicking on the menu while it's shifted to the left will shift it to the right, then you can click on an item to go to that section. It's either a Redoc bug (which means it should get its own ticket) or a very bad interaction between Redoc and the bund's css.

On Firefox, the menu simply shifts to the left the first time it's clicked but always lead you to the desired section. Seems like the bugs it's a little bit here, but less severe.

And on int, it expands the menu instead. 

Definitively something is going wrong with the css here. 

If I summarize : 

Expected comportment 
Click --> Menu opens, click again on a section to go there

Current comportment in dev
Click --> Menu shifts, and either takes you to the beginning of the main section, or wait for another click to lead you there.

First Conclusions : the bund CSS seems to be messing with Redoc's menus.

Thanks for the review

p1d1d1:
@ltkum and furthermore on "int" it remains visible when the page scrolls

ltkum:
> @ltkum and furthermore on "int" it remains visible when the page scrolls

Yup, I just saw that. When I remove the Bund CSS, it works as intended. 

I'll see if I can make this work, if not I'll have to make a small css file manually to cover the header and the footer styles.

ltkum:
@p1d1d1 a new version has been deployed but right now, cloudfront is serving the old one. I'll keep you updated once we've emptied the cache.

ltkum:
@p1d1d1 It should be corrected, and the cache is now cleared. Feel free to test and tell me if something feels off to you.

p1d1d1 (commented):
lgtm, but please additionally ask @davidoesch to review since he is PO

spec/static/spec/v0.9/api.html
```diff
@@ -4,20 +4,62 @@
...
-        }, document.getElementById('content'));
-    </script>
+
+        <!-- <div class="container container-main nav-open"> -->
+            <header style="border-bottom:4px solid #dc0018;">
+                <a href="#" class="brand hidden-xs" title="Back to home">
+                    <img src="/api/stac/static/style/img/logo-CH.svg" onerror="this.onerror=null; this.src='/api/stac/static/style/img/logo-CH.png'" alt="Back to home" />
+                    <h1>data.geo.admin.ch STAC API</h1>
+                </a>
+            </header>
```
ltshb, line 29: Wrong indentation

ltshb (commented):
Are really all the styles files required ? From the api.html it seems to me that they don't, only the logo and admin.css are required.
Other than that, by testing I've noticed that the selection of the search box and of the request URL (e.g. GET /conformance) looks not so nice like it used to be. Also in the samples the outer {} curly brace are on white background instead of black.

spec/static/spec/v0.9/api.html
ltkum: Apart from the indentation, I corrected everything we spoke of this morning. It's now deployed on dev.

davidoesch:
LGTM

ltshb (approved):
:+1:  As everyone seems to agreed on this :-)

ltshb (changes requested):
Please don't use the `update` button in the PR as it does a merge and not a rebase to develop. Could you revert the merge and do a rebase before merging ?

ltkum:
> Please don't use the `update` button in the PR as it does a merge and not a rebase to develop. Could you revert the merge and do a rebase before merging ?

Sure thing.

ltshb (approved):

PR #217, opened 2021-02-15
BGDIINF_SB-1664: Fixes and improvement of the admin Asset page

ltshb:
The asset admin page for create/update an asset could not be displayed
when the DB contained a lot of Items (several hundred of thousand). This
was due to the fact that the view was trying to load all items for the
select box.

Now this issue has been solved by using an auto_complete_fields.

Other than that some other improvements have been done:

- Moved the AssetAdmin at the end of the file to have a consistent class
order
- Added collection and item in the list view
- Allow the search of wholeword using ""
- Organized the asset fields in fieldset
- Displayed the collection name in the asset view (Currently not dynamically updated when the item is changed, for this a save and refresh is unfortunately needed)

Here below some printscreen of the improvement:

admin-asset-change
admin-asset-list

### DIFF
--- a/app/stac_api/models.py
+++ b/app/stac_api/models.py
@@ -328,7 +328,8 @@ class Item(models.Model):
         return instance
 
     def __str__(self):
-        return self.name
+        # This is used in the admin page in the autocomplete_fields of the Asset page
+        return f"{self.collection.name}/{self.name}"
 
     def update_etag(self):
         '''Update the ETag with a new UUID
--- a/app/stac_api/admin.py
+++ b/app/stac_api/admin.py
@@ -18,29 +18,6 @@ from stac_api.models import Provider
 from stac_api.utils import build_asset_href
 
 
-@admin.register(Asset)
-class AssetAdmin(admin.ModelAdmin):
-    readonly_fields = ['href', 'checksum_multihash']
-
-    def save_model(self, request, obj, form, change):
-        if obj.description == '':
-            # The admin interface with TextArea uses empty string instead
-            # of None. We use None for empty value, None value are stripped
-            # then in the output will empty string not.
-            obj.description = None
-        super().save_model(request, obj, form, change)
-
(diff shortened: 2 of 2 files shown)


rebert (approved):
nice. tested on functionality

boecklic (approved):
Cool! 
One thing that could help: if we change the `__str__` method of the item to the following:
```
    ...
    def __str__(self):
        return f"{self.collection.name}(_|-|.){self.name}"
```
this would clearly identify the item when searching for it...

hansmannj (approved):
Nice :+1:

ltshb:
> Cool!
> One thing that could help: if we change the `__str__` method of the item to the following:
> 
> ```
>     ...
>     def __str__(self):
>         return f"{self.collection.name}(_|-|.){self.name}"
> ```
> 
> this would clearly identify the item when searching for it...

Sounds like a good idea, I'll try it and if it works add it to this PR

ltshb:
asset-admin-page-autocomplete

boecklic (approved):

app/stac_api/models.py
```diff
@@ -328,7 +328,8 @@ def from_db(cls, db, field_names, values):
         return instance
 
     def __str__(self):
-        return self.name
+        # This is used in the admin page in the autocomplete_fields of the Asset page
+        return f"{self.collection.name}/{self.name}"
```
hansmannj, line 332: Wouldn't it make sense to also adapt the asset's `__str__(self)`, too?

hansmannj (approved):
:+1: looks good to me, just one question

app/stac_api/models.py
ltshb: For consistency could be, but otherwise I don't see any use case where it would be useful. Other than for item I had to add a `item_name` method in the `AssetAdmin` in order to have only the item name without the collection name due to this `Item.__str__(self)`.

PR #219, opened 2021-02-16
don't start threads if parallel is set to 1

boecklic:
Threading in generating dummy data is disabled if the two `--parallel-*` options are 1


ltshb (approved):

hansmannj (approved):
👍

PR #221, opened 2021-02-17
BGDIINF_SB-1669: Removed un-necessary validation for asset href

ltshb:
Asset href has been already set as read-only, but there were still some
code for href validation that was not anymore in use. This code as been
removed to avoid missunderstanding of the use of AWS_S3_CUSTOM_DOMAIN

Also improved an error message for POST/PUT Asset

### DIFF
--- a/app/tests/test_assets_endpoint.py
+++ b/app/tests/test_assets_endpoint.py
@@ -308,8 +308,8 @@ class AssetsWriteEndpointTestCase(StacBaseTestCase):
         self.assertEqual(
             {
                 'geoadmin:variant': [
-                    'Invalid geoadmin:variant, special characters beside one '
-                    'space are not allowed',
+                    'Invalid geoadmin:variant "more than twenty-five characters with s", '
+                    'special characters beside one space are not allowed',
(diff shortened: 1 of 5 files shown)


app/tests/test_assets_endpoint.py
```diff
@@ -308,8 +308,8 @@ def test_asset_endpoint_post_characters_geoadmin_variant(self):
         self.assertEqual(
             {
                 'geoadmin:variant': [
-                    'Invalid geoadmin:variant, special characters beside one '
-                    'space are not allowed',
+                    'Invalid geoadmin:variant "more than twenty-five characters with s", '
```
rebert, line 311: what does with s" mean?

rebert (approved):
Thanks @ltshb

app/tests/test_assets_endpoint.py
ltshb: between the "" is the value of the geoadmin:variant, so `more than twenty-five characters with s` is the value in the samples why it has an s at the end its a mistery :-)

PR #222, opened 2021-02-17
BGDIINF_SB-1667: added unittest to a case where geoadmin_variant with…

rebert:
… None will be updated

### DIFF
--- a/app/tests/test_collections_summaries.py
+++ b/app/tests/test_collections_summaries.py
@@ -179,6 +179,8 @@ class CollectionsSummariesTestCase(TestCase):
         collection = self.data_factory.create_collection_sample().model
         item = self.data_factory.create_item_sample(collection=collection, name='base-bbox').model
         asset = self.add_asset(item, 'har', None, None, None)
+        self.assertEqual(collection.summaries["geoadmin:variant"], [])
         asset.geoadmin_variant = "krel"
         asset.full_clean()
         asset.save()
+        self.assertEqual(collection.summaries["geoadmin:variant"], ["krel"])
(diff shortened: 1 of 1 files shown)


app/tests/test_collections_summaries.py
```diff
@@ -173,3 +173,12 @@ def test_update_collection_summaries_asset_update(self):
             "Collection's summaries[proj:epsg] has not been correctly "
             "updated after asset has been inserted."
         )
+
+    def test_update_collection_summaries_none_variant(self):
+        # update a variant, that as been None as a start value
+        collection = self.data_factory.create_collection_sample().model
+        item = self.data_factory.create_item_sample(collection=collection, name='base-bbox').model
+        asset = self.add_asset(item, 'har', None, None, None)
```
ltshb, line 181: An asserttion before that check the collection summaries would be good. Also then an assertion after the add asset that the collection summaries are still empty would be good.

app/tests/test_collections_summaries.py
```diff
@@ -173,3 +173,12 @@ def test_update_collection_summaries_asset_update(self):
...
+        asset.geoadmin_variant = "krel"
+        asset.full_clean()
+        asset.save()
```
ltshb, line 184: Finally an assertion at the end that check the collection summaries with the "krel" variant would be good

ltshb (changes requested):

app/tests/test_collections_summaries.py
rebert: ha - I forgot assertions :-) But it works without 500er ...

ltshb (approved):

IsabelleBzr (approved):

hansmannj (approved):
:+1:

PR #223, opened 2021-02-17
BGDIINF_SB-1662: healtheck endpoint for pingdom

rebert:
The path to the healthcheck is stored in the environment


app/config/settings_prod.py
```diff
@@ -216,6 +216,8 @@
     # permissions to change ACL, then this setting must be None
     AWS_DEFAULT_ACL = None
     AWS_S3_SIGNATURE_VERSION = "s3v4"
+    # The path for the pingdom healthcheck
+    HEALTHCHECK_ENDPOINT = os.environ.get('HEALTHCHECK_ENDPOINT', None)
```
ltshb, line 220: This should not be part of the AWS settings but outside this try except block. Then I would raise an error with a correct message if no HEALTHCHECK_ENDPOINT is set.

app/tests/test_healthcheck.py
```diff
@@ -0,0 +1,16 @@
...
+
+class HealthCheckTestCase(TestCase):
+
+    def setUp(self):
+        self.client = Client()
+
+    def test_healthcheck(self):
+        response = self.client.get(f"/{API_BASE}/{HEALTHCHECK_ENDPOINT}")
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response['Content-Type'], 'application/json')
```
ltshb, line 16: Test also for header `Cache-control: no-cache`

ltshb (changes requested):
The cache middleware should also be updated to add the correct cache-control header in order to bypass the CF caching.

app/stac_api/urls.py
```diff
@@ -45,4 +46,5 @@
         AssetDetail.as_view(),
         name='asset-detail'
     ),
+    path(f"{API_BASE}/{HEALTHCHECK_ENDPOINT}", CollectionList.as_view(), name='collection-list')
```
ltshb, line 49: I would put this just below the landing page and change the name as `health-check` instead of 'collection-list'. In order to keep a meaningful order of the endpoint in code.

app/stac_api/utils.py
```diff
@@ -50,6 +52,25 @@ def get_link(links, rel, raise_exception=False):
...
+    try:
+        # decode base64
+        base64_bytes = base64_path.encode('ascii')
+        message_bytes = base64.b64decode(base64_bytes)
+        path = message_bytes.decode('ascii')
+    except (TypeError, binascii.Error) as error:
+        path = base64_path
+    return path
+
+
```
ltshb, line 73: No need of this the encoding should be done outside the service and no need to decode.

app/config/settings_prod.py
```diff
@@ -184,6 +185,9 @@
 STATIC_ROOT = BASE_DIR / 'var' / 'www' / 'stac_api' / 'static_files'
 STATICFILES_DIRS = [BASE_DIR / "spec" / "static", BASE_DIR / "app" / "stac_api" / "templates"]
 STATICFILES_STORAGE = 'whitenoise.storage.CompressedManifestStaticFilesStorage'
+# The path for the pingdom healthcheck
+HEALTHCHECK_ENDPOINT = decode_path_if_base64(os.environ.get('HEALTHCHECK_ENDPOINT', 'healthcheck'))
```
ltshb, line 189: No decoding here just take whats comes from env variable or default to 'healthcheck'

app/middleware/cache_headers.py
```diff
@@ -26,9 +26,8 @@ def __call__(self, request):
 
         # Code to be executed for each request/response after
         # the view is called.
-
-        if request.path in ['/checker', '/get-token']:
-            # never cache the /checker and /get-token endpoints.
+        if request.path.split('/')[-1] in ['checker', 'get-token', settings.HEALTHCHECK_ENDPOINT]:
```
ltshb, line 29: To avoid forgetting this check if one of those path would change in the future, I would use the exact same string as in url.py, this way they can be found with a search and replace; `['checker', f'{API_BASE}/get-tocken', f'{API_BASE}/{HEALTH_CHECK_ENDPPOINT}']`

ltshb (commented):
After discussion with @boecklic, we decided to leave the base64 aside for the moment. The reason is that it don't really brings anything to hide this endpoint, we already have a public endpoint that is not cache at all; POST /search.
Having the endpoint configurable via environ variable, still let us in the future change the endpoint with a base64 value without changing the code though.

app/middleware/cache_headers.py
```diff
@@ -26,9 +26,12 @@ def __call__(self, request):
...
         # the view is called.
-
-        if request.path in ['/checker', '/get-token']:
-            # never cache the /checker and /get-token endpoints.
+        # comparing the last separation by / in both strings
+        if request.path.split('/')[-1] in [
+            i.split('/')[-1] for i in
+            ['/checker', f'{settings.API_BASE}/get-token', f'/{settings.HEALTHCHECK_ENDPOINT}']
+        ]:
+            # never cache the /checker, /get-token and healthcheck endpoints.
```
boecklic, line 34: We could simplify this by choosing patterns:
`/XYZ`: is used cluster-internally (not publicly visible), currently `/checker` and `/metrics`, uncached
`/api/XYZ` and `/api/stac/XYZ`: is used/visible externally, uncached, e.g. `/api/healthcheck`, `/api/stac/get-token`, ...
`/api/stac/static`: special handling of cache settings by whitenoise
`/api/stac/v0.9*: default cache settings

This would require some changes in `app/config/urls.py` and `app/stac_api/urls.py` (essentially moving the `/api/stac/` prefix part from `app/stac_api/urls.py` to `app/config/urls.py`...

app/middleware/cache_headers.py
```diff
@@ -26,9 +26,12 @@ def __call__(self, request):
...
         # Code to be executed for each request/response after
         # the view is called.
-
-        if request.path in ['/checker', '/get-token']:
-            # never cache the /checker and /get-token endpoints.
+        # comparing the last separation by / in both strings
+        if request.path.split('/')[-1] in [
+            i.split('/')[-1] for i in
+            ['/checker', f'{settings.API_BASE}/get-token', f'/{settings.HEALTHCHECK_ENDPOINT}']
+        ]:
```
ltshb, line 33: Doesn't seems to me correct here, the split is not needed if we use the full path in the list. Also the path for healthcheck is missing the API_BASE.
I've also just found out that the `/metrics` endpoint in this list is missing, it could be corrected part of this PR.

app/stac_api/urls.py
```diff
@@ -45,4 +46,5 @@
         AssetDetail.as_view(),
         name='asset-detail'
     ),
+    path(f'{HEALTHCHECK_ENDPOINT}', CollectionList.as_view(), name='health-check')
```
boecklic, line 49: nice!

ltshb, line 49: This won't work because only path behing `/api` are accessible from outside so healthcheck should be `{API_BASE}/{HEALTHCHECK_ENDPOINT}`

boecklic (commented):
nice. made a comment regarding url prefixes / url organisation

ltshb (commented):
Still some issues ;-)

app/middleware/cache_headers.py
ltshb: Good idea to resume we would have the following patterns:

- '/\w+' => cluster-internals, no cache; /checker, /metrics
- '/api/stac/\w+' => visible external, no cache; /api/stac/healthcheck /api/stac/get-tocken
- '/api/stac/admin/.*' => cache handled by django admin site
- '/api/stac/static/.*' => cache handled by Whitenoise
- '/api/stac/v0.9/.*' => add cache headers

app/stac_api/urls.py
rebert: Thanks - I didn't know that

app/config/urls.py
```diff
@@ -15,13 +15,24 @@
...
+
+
 urlpatterns = [
-    path('', include('stac_api.urls')),
-    path('api/stac/admin/', admin.site.urls),
-    path('', include('django_prometheus.urls')),
+    path('checker', checker, name='checker'),
+    path(f'{STAC_BASE}/', include('stac_api.urls')),
+    path(f'{STAC_BASE}/admin/', admin.site.urls),
+    path(STAC_BASE, include('django_prometheus.urls')),
```
boecklic, line 35: this should actually remain as it was (`/metrics` url must be at the root), so `path('', include('django_prometheus.urls')),`

app/middleware/cache_headers.py
```diff
@@ -8,11 +9,14 @@
...
 
+STAC_BASE = settings.STAC_BASE
+STAC_BASE_V = settings.STAC_BASE_V
+
 
 class CacheHeadersMiddleware:
     '''Middleware that adds appropriate cache headers to GET and HEAD methods.
 
-    NOTE: /checker and /get-token endpoints are marked as never cache.
+    NOTE: /checker, /get-token and /{healthcheck} endpoints are marked as never cache.
```
boecklic, line 19: `/metrics` as well

app/middleware/cache_headers.py
```diff
@@ -26,9 +30,13 @@ def __call__(self, request):
 
         # Code to be executed for each request/response after
         # the view is called.
-
-        if request.path in ['/checker', '/get-token']:
-            # never cache the /checker and /get-token endpoints.
+        # match /xxx or /stac/api/xxx
+        if (
+            re.match(r'^/((%s/)\w+|\w+)$' % STAC_BASE, request.path) or
+            # DOTO: /get-token could be moved to the base like f.ex /checker
```
boecklic, line 36: it could be moved to `/api/get-token` (it must be publicly visible). Imho we could do this change, GDWH has a token now, so not critical for them

boecklic (changes requested):
Nice cleanup! One thing you need to change, rest perfect

app/middleware/cache_headers.py
ltshb: @boecklic that was my idea to leave it `get-token` as it was to not enter in conflict with GDWH but if you think we are all good then, we should do it :+1:

ltshb (commented):
Very nice clean up. Just some minors comments, see below, but overall very good job.

app/config/urls.py
rebert: ok

app/middleware/cache_headers.py
rebert: ok

app/middleware/cache_headers.py
rebert: ok

app/middleware/cache_headers.py
```diff
@@ -26,9 +30,13 @@ def __call__(self, request):
 
         # Code to be executed for each request/response after
         # the view is called.
-
-        if request.path in ['/checker', '/get-token']:
-            # never cache the /checker and /get-token endpoints.
+        # match /xxx or /stac/api/xxx
```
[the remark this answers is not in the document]
rebert: ok

app/middleware/cache_headers.py
```diff
@@ -26,9 +30,13 @@ def __call__(self, request):
...
+        if (
+            re.match(r'^/((%s/)\w+|\w+)$' % STAC_BASE, request.path) or
```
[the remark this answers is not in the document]
rebert: does fr stand for French? I didn't know that you can combine them (f & r)

rebert:
@ltshb @boecklic as soon as ok from your side I'm going to squash it into one rebased commit

app/config/urls.py
```diff
@@ -15,13 +15,27 @@
...
+
+    return JsonResponse(data)
+
+
 urlpatterns = [
-    path('', include('stac_api.urls')),
-    path('api/stac/admin/', admin.site.urls),
     path('', include('django_prometheus.urls')),
+    path('checker', checker, name='checker'),
+    path('get-token', obtain_auth_token, name='get-token'),
```
ltshb, line 36: Oops still not the correct path from past discussion:
- original: `/api/stac/v0.9/get-tocken`
- new proposal from @boecklic : `/api/stac/get-tocken`
I would be ok to move it at the root, but I'm not sure if this is a good idea, @boecklic and @hansmannj  what are your opinion ?

ltshb (commented):
A part of the path of get-token is ok ;-)

app/config/urls.py
hansmannj: OK for me, it can well be decoupled from the API version, so no need for the `/v0.9/`-part for me.
This would imply, that a token remains also valid for future API versions, which is OK for me.

boecklic: `/api/stac/v0.9/get-token` and `/api/stac/get-token` both work, `/get-token` doesn't work (since not visible to the outside). 

If it makes code simpler (i.e. no exception regarding path parsing to determine cache header) I'd prefer `/api/stac/get-token` (plus it makes even more sense, get-token is independant of the stac version )

rebert: @ltshb sorry

ltshb (commented):
What last comment, the SPEC needs to be updated relating to the `get-token`. If also noticed that also the `get-token` is described in a paragraphe on the spec, its endpoint definition is missing it could/should be also added.

ltshb (approved):

boecklic (approved):

PR #228, opened 2021-02-19
BGDIINF_SB-1682

ltkum:
Issue : We want to cleanup almost everything with the admin css.

Fix : Removing every change made until now in css (first commit) to have a clean plate, then introducing changes to pages styling.

we kept two files from the official bund style repository : the favicon and the logo
we created a stac.css file to store the styling.

testlinks will follow soon to check if the css is acceptable (as it changed a tiny bit : mostly the gap between the footer and the content in api.html and apitransactional.html is now white.

Test links : 
https://service-stac.dev.bgdi.ch/api/stac/static/spec/v0.9/api.html
https://service-stac.dev.bgdi.ch/api/stac/static/spec/v0.9/apitransactional.html
https://service-stac.dev.bgdi.ch/api/stac/v0.9/ <-- this one might not display the logo right now because of the cache, the next one does.
https://service-stac.dev.bgdi.ch/api/stac/v0.9/collections 

I already saw a problem with the logo not being displayed (I'm correcting it right now), feel free to give me your feedbacks and check if this is acceptable.


app/templates/rest_framework/base.html
```diff
@@ -1,19 +1,58 @@
 {% extends "rest_framework/base.html" %}
 {% load static %}
-
 {% block style %}
 {{ block.super }}
-<link rel="icon" href="{% static 'favicon.ico' %}">
+    <link rel="icon" href="{% static 'style/favicon.ico' %}">
+    <link rel="stylesheet" href="{% static 'style/style.css' %}" type="text/css">
+{% endblock %}
+{% block title %}Geoadmin STAC API{% endblock %}
```
ltshb, line 8: Is this block title required ? if yes I would also use the same title as in line 15.

app/templates/rest_framework/base.html
```diff
@@ -1,19 +1,58 @@
...
+
+     <ul class="breadcrumb">
+    {% for breadcrumb_name, breadcrumb_url in breadcrumblist %}
+      {% if forloop.last %}
+        <li class="active"><a href="{{ breadcrumb_url }}">{{ breadcrumb_name }}</a></li>
+      {% else %}
+        <li><a href="{{ breadcrumb_url }}">{{ breadcrumb_name }}</a></li>
+      {% endif %}
+    {% endfor %}
+  </ul>
```
ltshb, line 27: Indentation issue

ltshb (commented):
Note sure if this state is deployed on DEV, but currently there is the following issue on DEV:

REST API (e.g. landing page):
- Footer on REST API (e.g. landing page) doesn't have the 4px rounded corner like other box
- Footer text is not aligned, left column is at top of the box while right column at bottom

API.html:
- Header should not have rounded corner to be consistent with the rest of the design of API (should be different than in REST API)
- Footer of API.html has the disclaimer at the bottom left instead of on the right
- Footer text would require some padding inside the box

app/templates/rest_framework/base.html
ltkum: The default comportment of the title block is to get the title of wherever we are ( as in the breadcrumbs menu. For example, /api/stac/v0.9/collections/ch.swisstopo.swisstlmregio/items would be "Item List") as the page <title>, followed by Django REST framework. using data.geo.admin.ch STAC API is cleaner. So I changed it to that.

app/templates/rest_framework/base.html
ltkum: Done

ltshb (commented):
Some minor issues, but looks already better:
- In api.html and apitransactional.html the footer margin-top could be removed and margin-bottom could be added.
- In REST API remove the background in .footer-columns to allow to have the rounded corner in the bottom of the footer.

ltshb (approved):
If these changes are deployed on DEV, then its all good. It looks now nice and professional :+1:

PR #229, opened 2021-02-19
updated cryptography package

rebert:


Pipfile
```diff
@@ -16,6 +16,7 @@ tblib = "*"  # needed for traceback when running tests in parallel
 responses = "!=0.12.1" # skip version 0.12.1 which has a bug see https://github.com/getsentry/responses/issues/358
 moto = {extras = [ "s3",], version = "*"}
 requests-mock = "*"
+cryptography = "~=3.3.2"
```
ltshb, line 19: Because it is only a dev package I suggest to use `>=3.3.2` this way it will be always updated to the latest fixing any security :-), but test needs first to pass.

ltshb:
@rebert why did you close it ?

PR #230, opened 2021-02-19
BGDIINF_SB-1663 add CORS headers to all responses

boecklic:
this PR adds CORS headers to all responses.

### DIFF
--- a/app/middleware/cors.py
+++ b/app/middleware/cors.py
@@ -1,0 +1,44 @@
+import logging
+import re
+from urllib.parse import urlparse
+
+from django.conf import settings
+from django.utils.cache import add_never_cache_headers
+from django.utils.cache import patch_cache_control
+from django.utils.cache import patch_response_headers
+
+logger = logging.getLogger(__name__)
+
+STAC_BASE = settings.STAC_BASE
+STAC_BASE_V = settings.STAC_BASE_V
+
+
+class CORSHeadersMiddleware:
+    '''Middleware that adds appropriate CORS headers.
+
+    CORS has only effect on browser applications (e.g. STAC browser),
+    not on other systems. Therefore we only allow GET and HEAD requests
+    on all endpoints except /search.
+    '''
+
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        # Code to be executed for each request before
+        # the view (and later middleware) are called.
+
+        response = self.get_response(request)
+
+        # Code to be executed for each request/response after
+        # the view is called.
+        response['Access-Control-Allow-Origin'] = '*'
+        # Access-Control-Allow-Methods:
+        allow_methods = ['GET', 'HEAD']
+        # For /search we allow POST as well
+        if request.path == f'{STAC_BASE_V}/search':
+            allow_methods.append('POST')
+        response['Access-Control-Allow-Methods'] = ','.join(allow_methods)
+        response['Access-Control-Allow-Headers'] = 'Content-Type,Accept'
+
+        return response
--- a/app/middleware/cors.py
+++ b/app/middleware/cors.py
@@ -1,11 +1,6 @@
 import logging
-import re
-from urllib.parse import urlparse
 
 from django.conf import settings
-from django.utils.cache import add_never_cache_headers
-from django.utils.cache import patch_cache_control
-from django.utils.cache import patch_response_headers
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


app/middleware/cors.py
```diff
@@ -0,0 +1,44 @@
+import logging
+import re
+from urllib.parse import urlparse
+
+from django.conf import settings
+from django.utils.cache import add_never_cache_headers
+from django.utils.cache import patch_cache_control
+from django.utils.cache import patch_response_headers
```
ltshb, line 8: These needs to be removed for lint :)

app/tests/test_generic_api.py
```diff
@@ -402,3 +402,80 @@ def test_head_cache_header(self):
...
+        cls.item = cls.factory.create_item_sample(
+            collection=cls.collection.model,
+            name='item-1',
+            db_create=True,
+        )
+        cls.asset = cls.factory.create_asset_sample(
+            item=cls.item.model,
+            name='asset-1',
+            db_create=True,
+        )
```
ltshb, line 426: Just as a side note, here the `name='collection|item|asset-1'` would not be required as it is the default value of the create sample.

app/tests/test_generic_api.py
```diff
@@ -402,3 +402,80 @@ def test_head_cache_header(self):
...
+
+    def test_get_cors_header(self):
+        for endpoint in [
+            '',  # landing page
+            'conformance',
+            f'collections/{self.collection["name"]}',
+            f'collections/{self.collection["name"]}/items/{self.item["name"]}',
+            f'collections/{self.collection["name"]}/items/{self.item["name"]}'
+            f'/assets/{self.asset["name"]}'
```
ltshb, line 435: The following endpoint are missing:
- api/stac/v0.9/collections
- api/stac/v0.9/.../items
- api/stac/v0.9/.../assets

app/tests/test_generic_api.py
```diff
@@ -402,3 +402,80 @@ def test_head_cache_header(self):
...
+
+                self.assertTrue(
+                    response.has_header('Access-Control-Allow-Methods'),
+                    msg="CORS header allow-methods missing"
+                )
+                self.assertEqual(
+                    response['Access-Control-Allow-Methods'],
+                    'GET,HEAD',
+                    msg='Wrong CORS allow-methods value'
+                )
```
ltshb, line 459: The check for `Access-Control-Allow-Headers` is missing

app/tests/test_generic_api.py
```diff
@@ -402,3 +402,80 @@ def test_head_cache_header(self):
...
+
+        self.assertTrue(
+            response.has_header('Access-Control-Allow-Methods'),
+            msg="CORS header allow-methods missing"
+        )
+        self.assertEqual(
+            response['Access-Control-Allow-Methods'],
+            'GET,HEAD,POST',
+            msg='Wrong CORS allow-methods value'
+        )
```
ltshb, line 481: Same as above the check for `Access-Control-Allow-Headers` is missing.

ltshb (commented):
See comments

app/middleware/cors.py
```diff
@@ -0,0 +1,44 @@
...
+STAC_BASE = settings.STAC_BASE
+STAC_BASE_V = settings.STAC_BASE_V
+
+
+class CORSHeadersMiddleware:
+    '''Middleware that adds appropriate CORS headers.
+
+    CORS has only effect on browser applications (e.g. STAC browser),
+    not on other systems. Therefore we only allow GET and HEAD requests
+    on all endpoints except /search.
```
hansmannj, line 21: How about the get-token endpoint? Probably better to not allow POST from origin "*" though..

boecklic: We would only need to allow CORS for POST get-token if this was done/issued directly within a browser that has cors protection enabled. I assume this is not the case, so we leave it out.

app/tests/test_generic_api.py
boecklic: Well, it's a middleware which treats all endpoints except /search equally, so imho it's safe to not include every single endpoint here. But I can add them if you prefer.

app/tests/test_generic_api.py
boecklic: removed it

app/middleware/cors.py
boecklic: removed

app/tests/test_generic_api.py
boecklic: added

app/tests/test_generic_api.py
ltshb: I don't mind leave it as is, but it is just in case if we would change the middleware logic it could be useful to have it then in unittest.

ltshb (approved):

PR #232, opened 2021-02-23
BGDIINF_SB-1490 moved and added unit tests

hansmannj:
* moved CollectionsModelTemporalExtentTestCase to test_collection_extent.py
* renamed test_item_to_collection_propagation.py to test_collections_extent.py
* added unit tests for creation of collections in test_collections_model.py
* added unit tests for asset model


boecklic (approved):

app/tests/test_assets_model.py
```diff
@@ -0,0 +1,90 @@
...
+
+class AssetsModelTestCase(TestCase):
+
+    @classmethod
+    @mock_s3_asset_file
+    def setUpTestData(cls):
+        cls.factory = Factory()
+        cls.collection = cls.factory.create_collection_sample(db_create=True).model
+        cls.item = cls.factory.create_item_sample(
+            collection=Collection.objects.get(name="collection-1"), db_create=True
```
ltshb, line 23: Here you can simply do `collection=cls.collection` which is safer because the name of the created collection is not guaranteed.

app/tests/test_assets_model.py
```diff
@@ -0,0 +1,90 @@
...
+        ).model
+        cls.asset = cls.factory.create_asset_sample(
+            item=cls.item, db_create=True, create_asset_file=False
```
ltshb, line 26: Why using `create_asset_file=False` ? I would not to make the test more realistic.

app/tests/test_assets_model.py
```diff
@@ -0,0 +1,90 @@
...
+                create_asset_file=False
+            )
+
+    def test_create_asset_valid_geoadmin_variant(self):
+        # try to create an asset with valid geoadmin variant. This should not raise any error.
+        self.factory.create_asset_sample(
+            item=self.item,
+            sample="asset-valid-geoadmin-variant",
+            db_create=True,
+            create_asset_file=False
```
ltshb, line 65: Currently there is no asset file existing check in model validation, only in serializer validation but if this is in future added this test would failed. I would recommend to use everywhere here `create_asset_file=True`

app/tests/test_collections_model.py
```diff
@@ -1,737 +1,73 @@
...
+            collection.full_clean()
+            collection.save()
+
+    def test_create_collection_invalid_name(self):
+        # try to create a collection with invalid collection name
+        with self.assertRaises(ValidationError, msg="Collection with invalid name was accepted."):
+            collection_data = CollectionFactory().create_sample(sample="collection-invalid")
+            collection = Collection(**collection_data.attributes)
+            collection.full_clean()
+            collection.save()
```
ltshb, line 32: For consistency I would do it like below, using the `db_create=True`.

ltshb (changes requested):
See inline comment

app/tests/test_assets_model.py
hansmannj: Done

app/tests/test_assets_model.py
hansmannj: changed to True

app/tests/test_assets_model.py
hansmannj: Done

app/tests/test_assets_model.py
ltshb: Oops sorry I forgot about this, like mentioned in slack, that's my fault, the factory is not very intuitive regarding the `create_asset_file`. I might have to search for a better name which would be  more intuitive.

ltshb (approved):

app/tests/test_collections_model.py
hansmannj: Done

PR #233, opened 2021-02-23
BGDIINF_SB-1671: added logging of requests and payloads for write endpoints

hansmannj:
adapted middleware/logging.py and stac_api/apps.py to log payload of requests to write endpoints

### DIFF
--- a/app/middleware/logging.py
+++ b/app/middleware/logging.py
@@ -16,14 +16,12 @@ class RequestResponseLoggingMiddleware:
     def __call__(self, request):
         # Code to be executed for each request before
         # the view (and later middleware) are called.
-        logger.info(
-            "Request %s %s",
-            request.method.upper(),
-            request.path,
-            extra={
-                "request": request, "requestPayload": str(request.body[:200])
-            }
-        )
+        if request.method.upper() in ["PATCH", "POST", "PUT"]:
+            extra = {"request": request, "requestPayload": str(request.body[:200])}
+        else:
+            extra = {"request": request}
+
+        logger.info("Request %s %s", request.method.upper(), request.path, extra=extra)
         start = time.time()
 
         response = self.get_response(request)
--- a/app/middleware/logging.py
+++ b/app/middleware/logging.py
@@ -17,7 +17,12 @@ class RequestResponseLoggingMiddleware:
         # Code to be executed for each request before
         # the view (and later middleware) are called.
         logger.info(
-            "Request %s %s", request.method.upper(), request.path, extra={"request": request}
+            "Request %s %s",
+            request.method.upper(),
+            request.path,
+            extra={
+                "request": request, "requestPayload": str(request.body[:200])
+            }
         )
         start = time.time()
 
@@ -35,7 +40,7 @@ class RequestResponseLoggingMiddleware:
         # HttpResponse and JSONResponse sure have
         # (e.g. WhiteNoiseFileResponse doesn't)
         if isinstance(response, (HttpResponse, JsonResponse)):
-            extra["response"]["content"] = str(response.content)[:200]
+            extra["response"]["payload"] = str(response.content)[:200]
 
         logger.info("Response %s", response.status_code, extra=extra)
         # Code to be executed for each request/response after
--- a/app/stac_api/apps.py
+++ b/app/stac_api/apps.py
@@ -112,11 +112,13 @@ def custom_exception_handler(exc, context):
     response = exception_handler(exc, context)
 
     if response is not None:
+        extra = {"request": context['request']._request, "requestPayload": context['request'].data}  # pylint: disable=protected-access
+        logger.error(repr(exc), extra=extra)
         response.data = {'code': response.status_code, 'description': response.data}
         return response
 
     # If we don't have a response that's means that we have an unhandled exception that needs to
-    # return a 500. We need to log the exception here as it might not be re-raised.ma
+    # return a 500. We need to log the exception here as it might not be re-raised.
     extra = {"request": context['request']._request}  # pylint: disable=protected-access
     logger.critical(repr(exc), extra=extra, exc_info=sys.exc_info())
 
--- a/app/stac_api/apps.py
+++ b/app/stac_api/apps.py
@@ -112,8 +112,15 @@ def custom_exception_handler(exc, context):
     response = exception_handler(exc, context)
 
     if response is not None:
-        extra = {"request": context['request']._request, "requestPayload": context['request'].data}  # pylint: disable=protected-access
-        logger.error(repr(exc), extra=extra)
+        if context['request']._request.method.upper() in ["PATCH", "POST", "PUT"]:  # pylint: disable=protected-access
+            extra = {
+                "request": context['request']._request,  # pylint: disable=protected-access
+                "requestPayload":
+                    context['request'].data  # pylint: disable=protected-access
+            }
+        else:
+            extra = {"request": context['request']._request}  # pylint: disable=protected-access
+        logger.error("Response %s: %s", response.status_code, response.data, extra=extra)
         response.data = {'code': response.status_code, 'description': response.data}
         return response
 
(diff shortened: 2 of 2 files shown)


app/stac_api/views.py
```diff
@@ -301,16 +317,19 @@ def get(self, request, *args, **kwargs):
...
     # Here the etag is only added to support pre-conditional If-Match and If-Not-Match
     @etag(get_item_etag)
     def patch(self, request, *args, **kwargs):
+        debug_print_request_data(self)
         return self.partial_update(request, *args, **kwargs)
 
     # Here the etag is only added to support pre-conditional If-Match and If-Not-Match
     @etag(get_item_etag)
     def delete(self, request, *args, **kwargs):
+        debug_print_request_data(self)
```
ltshb, line 332: In delete we don't expect any body and don't do anything about it so no need to log it.

app/stac_api/views.py
```diff
@@ -31,6 +32,17 @@
 logger = logging.getLogger(__name__)
 
 
+def debug_print_request_data(view):
+    if sys.getsizeof(view.request.data) < 500000:
```
ltshb, line 36: The first access to `.data` on the request of the `rest_framework` trigger the body stream to be fully read and parsed. Looking at the rest_framework docu (https://www.django-rest-framework.org/api-guide/requests/#data) and at its code. The parsing takes already care of edge cases and malformed body. So I would suggest to remove this check. However once the logging is done, I would quickly test the behavior with a script that would do a POST with a valid json payload e.g. 
```python
import requests
requests.post(url, data ={'hugeString': 'This is a huge String' * 1000 * 1000 * 1000})
```

ltshb (commented):
The logging statement could/should be added to the view_mixins `update()` and `create()` method after the `data = self.get_write_request_data(request, *args, **kwargs)` statement. Then some view should be changed to uses our mixins instead of the original from rest framework but I think this should be OK.

app/stac_api/views.py
```diff
@@ -31,6 +32,17 @@
...
+        logger.debug(
+            "Request: %s, request.data: %s",
+            view.request,
+            view.request.data,
+        )
```
ltshb, line 41: The logging should make uses of the extra and uses `info` like in logging middleware; `logger.info('Request %s %s', method, path, extra={'request': request})`) the `data` should then be added to the logging configuration 
```yaml
filters:
  django:
    (): logging_utilities.filters.django_request.JsonDjangoRequest
    include_keys:
      - request.path
      - request.method
      - request.headers
      - request.data
    exclude_keys:
      - request.headers.Authorization
      - request.headers.Proxy-Authorization
      - request.headers.Cookie
``` 
NOTE: one could think that simply changing the log configuration would be enough due to our logging middleware, however in the logging middleware the request object is a django HttpRequest object that don't have a `.data`. It only has a `.body` which is a bytestring.

ltshb (commented):
I would also update the logging middleware to use messages like below:
- `'Request %s %s', method, path` => 'Request GET /ap/stac/v0.9/'
- `'Response to request %s %s', method, path` => 'Response to request GET /ap/stac/v0.9/'
Currently the message are simly:
- `request`
- `request-response`
This would make the search and reading of logs in Kibana easier when scrolling in logs, what do you think @boecklic ?

hansmannj:
> The logging statement could/should be added to the view_mixins `update()` and `create()` method after the `data = self.get_write_request_data(request, *args, **kwargs)` statement. Then some view should be changed to uses our mixins instead of the original from rest framework but I think this should be OK.

That's where I had it in my first tries, after your hints, but then I thought, having a simple `debug_print_request_data()` in `views.py` might do as well. But I can move things into our `views_mixins,py`. For this, I'll have to think about some nice solution for the SearchList's `ListModelMixin`.

hansmannj:
> I would also update the logging middleware to use messages like below:
> 
>     * `'Request %s %s', method, path` => 'Request GET /ap/stac/v0.9/'
> 
>     * `'Response to request %s %s', method, path` => 'Response to request GET /ap/stac/v0.9/'
>       Currently the message are simly:
> 
>     * `request`
> 
>     * `request-response`
>       This would make the search and reading of logs in Kibana easier when scrolling in logs, what do you think @boecklic ?

You mean as in here: https://github.com/geoadmin/service-stac/pull/235/files? ;-)

boecklic (commented):
Just one question: What's the reason for not adding this part to the logging middleware? For debugging it's probably sufficient to just print the first e.g. 200 chars from the `.body` (since `.data` is not available) (the parsed json is not really necessary, it's the same for the `response.content`)

ltshb:
> Just one question: What's the reason for not adding this part to the logging middleware? For debugging it's probably sufficient to just print the first e.g. 200 chars from the `.body` (since `.data` is not available) (the parsed json is not really necessary, it's the same for the `response.content`)

Personally I would prefer having the JSON parsed for logging it makes the reading of the log on Kibana much easier. But you're might be right reading as for the response might work, although looking at the django doc, there is two possible ways of reading the body, which is using the file like `read()` and `readline()` and accessing the `body` after having it read by those raise a `RawPostDataException` (https://docs.djangoproject.com/en/3.1/ref/request-response/#django.http.HttpRequest.body). I'm not if this could be an issue depending on the middleware order ?

hansmannj:
When testing locally with `manage.py test` there's a new issue popping up, but the test runner says `OK` in the end. I guess this is probably a new "feature" and not a bug, introduced by PR #234.
```CRITICAL - stac_api.apps              : KeyError('collection_name') - collection:  - item:  - asset:  - duration: 
[...]
/service-stac/app/stac_api/views.py", line 495, in get_queryset  item__collection__name=self.kwargs['collection_name'],
KeyError: 'collection_name'
```
In the test `test_generic_api.ApiGenericTestCase.test_http_error_500_exception`

rebert (approved):

app/stac_api/apps.py
```diff
@@ -112,11 +112,13 @@ def custom_exception_handler(exc, context):
     response = exception_handler(exc, context)
 
     if response is not None:
+        extra = {"request": context['request']._request, "requestPayload": context['request'].data}  # pylint: disable=protected-access
```
ltshb, line 115: Just formatting issue here

app/stac_api/apps.py
```diff
@@ -112,11 +112,13 @@ def custom_exception_handler(exc, context):
...
+        logger.error(repr(exc), extra=extra)
```
ltshb, line 116: I would not log the exception here, the exception is handled. I would log `Response %s: %s % (status_code, response.data)`

ltshb (changes requested):
See comment below

app/middleware/logging.py
```diff
@@ -17,7 +17,12 @@ def __call__(self, request):
         # Code to be executed for each request before
         # the view (and later middleware) are called.
         logger.info(
-            "Request %s %s", request.method.upper(), request.path, extra={"request": request}
+            "Request %s %s",
+            request.method.upper(),
+            request.path,
+            extra={
+                "request": request, "requestPayload": str(request.body[:200])
```
ltshb, line 24: Same here we need to log the payload only for POST, PUT, PATCH request

app/stac_api/apps.py
hansmannj: :+1:

app/middleware/logging.py
hansmannj: Done

app/stac_api/apps.py
```diff
@@ -112,11 +112,20 @@ def custom_exception_handler(exc, context):
     response = exception_handler(exc, context)
 
     if response is not None:
+        if context['request']._request.method.upper() in ["PATCH", "POST", "PUT"]:  # pylint: disable=protected-access
+            extra = {
+                "request": context['request']._request,  # pylint: disable=protected-access
+                "requestPayload":
+                    context['request'].data  # pylint: disable=protected-access
```
ltshb, line 119: pylint disable not required here

ltshb (approved):
The pylint pragma could be just added once in the start of the `ìf response is not None:` between line 114 and 115

app/stac_api/apps.py
hansmannj: :+1:

PR #234, opened 2021-02-24
BGDIINF_SB-1681: Log exception also on production build

ltshb:
On production build with the DEBUG=False the exception where not logged
because the rest framework custom exception handler in apps.py did
caught all exceptions and returned a proper 500 message (JSON or HTML
depending on Accept header or format query). Due to this the logging exception middleware was not called.

So now we logs the exception in the custom exception handler instead of
in the exception middleware. This means that for all exceptions on the
API endpoints we will have a log message from logger="stac_api.app" with
level CRITICAL and with exc_info=True. 
On the other hand all exceptions happening from
the admin interface endpoints, we will have a log from
logger="django.server" of level ERROR with exc_info=True. This is due to the fact that for those endpoints the rest framework custom exception is not called.

I've tried also to remove the custom exception handler, and leave it to the exception handler to change the HTTP 500 message body to our convention: `{'code': 500, 'description': 'Internal Server Error'}`, but then we will always have a JSON response instead of an API response like for other normal response (meaning the rest framework api html response for browser)

### DIFF
--- a/app/config/settings_prod.py
+++ b/app/config/settings_prod.py
@@ -99,7 +99,6 @@ MIDDLEWARE = [
     'django.contrib.auth.middleware.AuthenticationMiddleware',
     'django.contrib.messages.middleware.MessageMiddleware',
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
-    'middleware.logging.ExceptionLoggingMiddleware',
     'middleware.cache_headers.CacheHeadersMiddleware',
     'middleware.cors.CORSHeadersMiddleware',
     'django_prometheus.middleware.PrometheusAfterMiddleware',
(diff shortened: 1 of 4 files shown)


hansmannj (approved):
:+1: LGTM

boecklic (commented):
What happens with errors e.g. in the /admin? They are not capture anymore when removing the middleware I guess?

ltshb:
> What happens with errors e.g. in the /admin? They are not capture anymore when removing the middleware I guess?

As I mentionned in the description, the errors of the /admin are logged by `django.server` with the `exc_info=True` so no need for the exception middleware. I could leave it but I find it a bit confusing because it will only be used for the `/admin` not for other endpoints.

boecklic:
> > What happens with errors e.g. in the /admin? They are not capture anymore when removing the middleware I guess?
> 
> As I mentionned in the description, the errors of the /admin are logged by `django.server` with the `exc_info=True` so no need for the exception middleware. I could leave it but I find it a bit confusing because it will only be used for the `/admin` not for other endpoints.

Ok. If the middleware cannot be adapted to handle the execption logging from rest responses as well, we remove it.

hansmannj (approved):
:1st_place_medal:

PR #235, opened 2021-02-24
Improved logging

ltshb:
The logging middleware message was not very usefull. It did not helped
at all on local developement because we don't use the json format with
the request dictionary and on Kibana the message 'request' did not
helped neither unless we activate the whole request object.


hansmannj (approved):
:+1:

PR #236, opened 2021-02-25
Improved the data factory for unittest

ltshb:
The data factory was not very intuitive with the `create_asset_file`
option. So now even if `create_asset_file` and `db_create` are true then
automatically correct the `create_asset_file` instead of given an error.
This makes the usage of the interface more intuitive.

### DIFF
--- a/app/tests/data_factory.py
+++ b/app/tests/data_factory.py
@@ -1074,10 +1074,9 @@ class AssetFactory(FactoryBase):
             The data sample
         '''
         if db_create and create_asset_file:
-            raise ValueError(
-                'Cannot have db_create=True and create_asset_file=True, db_create=True implicitely '
-                'create the asset file when the "file" attribute is present'
-            )
+            # the asset file are automatically created by the model when creating the model
+            # therefore here correct the `create_asset_file` to false to avoid issues
+            create_asset_file = False
         return super().create_sample(
             sample,
             name=name,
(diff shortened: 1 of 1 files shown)


app/tests/data_factory.py
```diff
@@ -1074,10 +1074,9 @@ def create_sample(
             The data sample
         '''
         if db_create and create_asset_file:
-            raise ValueError(
-                'Cannot have db_create=True and create_asset_file=True, db_create=True implicitely '
-                'create the asset file when the "file" attribute is present'
-            )
+            # the asset file are automatically created by the model when creating the model
+            # therefore here correct the `create_asset_file` to false to avoid issues
+            create_asset_file = False
```
hansmannj, line 1079: That does the trick, I guess.
But probably it would be more intuitive, instead of forcibly set `create_asset_file = False` to do something like in the following (even if the code would grow a bit), just a suggestion, what do you think?

```python
        if db_create and create_asset_file:
            return super().create_sample(
                sample,
                name=name,
                item=item,
                db_create=db_create,
                required_only=required_only,
                **kwargs
            )
        
        else:
            return super().create_sample(
                sample,
                name=name,
                item=item,
                db_create=db_create,
                required_only=required_only,
                create_asset_file=create_asset_file,
                **kwargs
            )
```

hansmannj: of course with the comment, that you added above, also added somewhere in this snippet

hansmannj: But I can also live with your one-liner ;-)

hansmannj (approved):

hansmannj (approved):
That seems more intuitive for me, thanks :-)

PR #238, opened 2021-02-26
BGDIINF_SB-1685: As the path to /get-token has changed, the spec has …

rebert:
…been adapted

### DIFF
--- a/spec/static/spec/v0.9/openapitransactional.yaml
+++ b/spec/static/spec/v0.9/openapitransactional.yaml
@@ -72,7 +72,7 @@ tags:
     Here is an example using curl:
     ```
     curl --request POST \
-      --url https://service-stac.dev.bgdi.ch/api/stac/v0.9/get-token \
+      --url https://service-stac.dev.bgdi.ch/api/stac/get-token \
       --header 'Content-Type: application/json' \
       --data '{"username": "MickeyMouse", "password": "I_love_Minnie_Mouse"}'
     ```
@@ -2643,6 +2643,8 @@ paths:
         5XX:
           $ref: "#/components/responses/ServerError"
   /get-token:
+    servers:
+    - url: http://data.geo.admin.ch/api/stac/
     post:
       tags:
       - Data Management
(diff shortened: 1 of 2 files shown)


ltshb (changes requested):
You also need to add the POST get-token docu as for other endpoints. 
````
paths:
   /get-token:
      .....
````

rebert:
@ltshb So you don't mean the POST here? http://localhost:8090/apitransactional.html#operation/getToken

rebert:
@ltshb https://github.com/geoadmin/service-stac/pull/238/files#diff-9068ad3bd85681717e47d5c094ae2cf17bbe11d47771013c7bc56e89488df2e7R2645

ltshb:
> 
> 
> @ltshb So you don't mean the POST here? http://localhost:8090/apitransactional.html#operation/getToken

yes I mean the POST, I thought it was missing, but it simply at the bottom and on INT it is still on the wrong path (https://service-stac.int.bgdi.ch/api/stac/static/spec/v0.9/apitransactional.html#operation/getToken). Would be nice to have the POST /get-token just after the curl examples: https://service-stac.int.bgdi.ch/api/stac/static/spec/v0.9/apitransactional.html#section/Token-authentication

hansmannj:
Now all authentications options are grouped under the POST /get-token endpoint, even though, strictly speaking, the basic authentication and session authentication are not related to the get-token endpoint. But then I don't know, if it is possible, to have a kind of "simple, general and not endpoint related" heading in the list on the left, under which all authentication related things, that don't relate to the get-token endpoint, could be grouped.

ltshb:
I've seen that the `tags` definition for `Data Management` is missing. The docu and examples for authentication could be added to the tag description (see https://github.com/geoadmin/service-stac/blob/f6db96a2102597a1d91f84da76bed62c2f135155/spec/base.yml#L15) instead of the POST /get-token. Maybe also consider using https://swagger.io/docs/specification/authentication/ for documenting authentication.

ltshb (approved):

PR #240, opened 2021-03-01
BGDIINF_SB-1578: help text support for search

ltkum:
Issue : We wanted a help text to help users better use the tool within
the admin pages

Fix :
Added help texts to asset and item detail pages for the search.
Added a popup window on hover in the searches of general admin pages
search.

further fixes needed :
Indentation in JS files.

### DIFF
--- a/app/stac_api/admin.py
+++ b/app/stac_api/admin.py
@@ -58,6 +58,11 @@ class CollectionLinkInline(admin.TabularInline):
 
 @admin.register(Collection)
 class CollectionAdmin(admin.ModelAdmin):
+
+    class Media:
+        js = ('js/admin/collection_help_search.js',)
+        css = {'all': ('style/hover.css',)}
+
     readonly_fields = [
         'extent_start_datetime', 'extent_end_datetime', 'summaries', 'extent_geometry'
     ]
@@ -79,6 +84,11 @@ class ItemLinkInline(admin.TabularInline):
 
 @admin.register(Item)
 class ItemAdmin(admin.GeoModelAdmin):
+
+    class Media:
+        js = ('js/admin/item_help_search.js',)
+        css = {'all': ('style/hover.css',)}
+
     inlines = [ItemLinkInline]
     autocomplete_fields = ['collection']
     search_fields = ['name', 'collection__name']
@@ -122,6 +132,11 @@ class ItemAdmin(admin.GeoModelAdmin):
 
 @admin.register(Asset)
 class AssetAdmin(admin.ModelAdmin):
+
+    class Media:
+        js = ('js/admin/asset_help_search.js',)
+        css = {'all': ('style/hover.css',)}
+
     autocomplete_fields = ['item']
     search_fields = ['name', 'item__name', 'item__collection__name']
     readonly_fields = ['item_name', 'collection', 'href', 'checksum_multihash']
--- a/app/stac_api/models.py
+++ b/app/stac_api/models.py
@@ -41,6 +41,74 @@ _BBOX_CH.srid = 4326
 # 'SRID=4326;POLYGON ((5.96 45.82, 5.96 47.81, 10.49 47.81, 10.49 45.82, 5.96 45.82))'
 BBOX_CH = str(_BBOX_CH)
 
+SEARCH_TEXT_HELP_ITEM = '''
+    <div class=SearchUsage> 
+        Search Usage: 
+        <ul>
+            <li> 
+                <i>arg</i> will make a non exact search checking if <i>>arg</i>  
+                is part of the Item path 
+            </li>
+            <li> 
+                Multiple <i>arg</i>  can be used, separated by spaces. This will search 
+                for all elements containing all arguments in their path 
+            </li>
+            <li> 
+                <i>"collectionID/itemID"</i> will make an exact search for the specified item. 
+             </li> 
+        </ul> 
+        Examples : 
+        <ul>
+            <li> 
+                Searching for <i>pixelkarte</i> will return all items which have
+                pixelkarte as a part of either their collection ID or their item ID
+            </li>
+            <li> 
+                Searching for <i>pixelkarte 2016 4</i> will return all items
+                which have pixelkarte, 2016 AND 4 as part of their collection ID or
+                item ID 
+            </li>
+            <li>
+                Searching for <i>"ch.swisstopo.pixelkarte.example/item2016-4-example"</i> 
+                will yield only this item, if this item exists.
+            </li>
+        </ul>
+    </div>'''
+
+SEARCH_TEXT_HELP_COLLECTION = '''
+    <div class=SearchUsage> 
+        Search Usage: 
+        <ul> 
+            <li>
+                <i>arg</i> will make a non exact search checking if <i>arg</i> is part of 
+                the collection ID 
+            </li>
+            <li>
+                Multiple <i>arg</i> can be used, separated by spaces. This will search for all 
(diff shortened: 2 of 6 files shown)


app/stac_api/models.py
```diff
@@ -41,6 +41,35 @@
...
+                              ' arguments.</li><li>["collectionID"] will make an exact search for' \
+                              'the specified collection.</li></ul> Examples : <ul> <li>' \
+                              'Searching for "pixelkarte" will return all collections which have' \
+                              ' pixelkarte as a part of their collection ID </li> <li>' \
+                              'Searching for "pixelkarte 2016 4" will return all collection ' \
+                              'which have pixelkarte, 2016 AND 4 as part of their collection ID' \
+                              '</li><li> Searching for "ch.swisstopo.pixelkarte.example" will ' \
+                              'yield only this collection, if this collection exists. Please ' \
+                              'note that it would not return a collection named ' \
+                              'ch.swisstopo.pixelkarte.example.2.</li> </ul></div>'
```
ltshb: If the `help_text` attribute accept multi-line string, it would be easier here to use python multi-line string `'''Multi-line string'''`

app/stac_api/admin.py
```diff
@@ -142,6 +159,8 @@ class AssetAdmin(admin.ModelAdmin):
     )
 
     def get_search_results(self, request, queryset, search_term):
+        print("HELLO")
```
rebert, line 162: hello

app/stac_api/models.py
```diff
@@ -41,6 +41,35 @@
 # 'SRID=4326;POLYGON ((5.96 45.82, 5.96 47.81, 10.49 47.81, 10.49 45.82, 5.96 45.82))'
 BBOX_CH = str(_BBOX_CH)
 
+SEARCH_TEXT_HELP_ITEM = '<div class=search-usage> Search Usage: <ul> ' \
```
rebert, line 44: This could be wrapped by _ 
from django.utils.translation import gettext_lazy as _ for the future, if multilangual

app/stac_api/templates/js/admin/collection_help_search.js
```diff
@@ -0,0 +1,58 @@
+window.onload = function() {
```
ltshb, line 1: Code style (indentation issues, line length, ...)

app/stac_api/templates/js/admin/item_help_search.js
```diff
@@ -0,0 +1,63 @@
+window.onload = function() {
```
ltshb, line 1: Code style (indentation, line-length)

ltshb (commented):
The js and css are not loaded on Firefox, I got those messages:
```
The resource from “http://127.0.0.1:8000/api/stac/static/style/hover.css” was blocked due to MIME type (“text/html”) mismatch (X-Content-Type-Options: nosniff).
The resource from “http://127.0.0.1:8000/api/stac/static/js/admin/asset_help_search.js” was blocked due to MIME type (“text/html”) mismatch (X-Content-Type-Options: nosniff).

```

app/stac_api/admin.py
```diff
@@ -142,6 +159,8 @@ class AssetAdmin(admin.ModelAdmin):
...
+        logging.info("hello")
```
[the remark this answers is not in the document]
ltkum: done. I forgot those.

app/stac_api/models.py
ltkum: This is a good question for @boecklic : do we want multilingual support for the admin pages in the future or not ?

app/stac_api/models.py
ltkum: I tried it locally and it seems to be accepted. I'll be pushing this shortly.

app/stac_api/models.py
boecklic: If it doesn't cost anything to be prepared for multilingual, I'd go for it, even if I think that this kind of admin interfaces don't need to be translated

app/stac_api/templates/js/admin/collection_help_search.js
ltkum: should be better now.

app/stac_api/templates/js/admin/asset_help_search.js
```diff
@@ -0,0 +1,72 @@
...
+            Multiple [arg] can be used, separated by spaces. This will search for all assets ID,
+            items ID / or collections ID containing all arguments.
+        </li>
+        <li>
+            ["arg"] will make an exact search on the asset ID field.
+        </li>
+        <li>
+          ["collectionID/itemID/assetID"] will make an exact search on a whole path.
+        </li>
+    </ul>
```
boecklic, line 19: Would be nice if you could reuse here the text from `SEARCH_TEXT_HELP_ITEM` somehow so it's just defined in one place

app/stac_api/templates/js/admin/collection_help_search.js
```diff
@@ -0,0 +1,62 @@
...
+		<li>
+		    Multiple [arg] can be used, separated by spaces. This will search for all
+		    collections ID containing all arguments.
+		</li>
+		<li>
+		    ["arg"] will make an exact search on the collection ID field.
+		</li>
+
+    </ul>
+    Examples :
```
boecklic, line 17: dito

app/stac_api/templates/js/admin/item_help_search.js
```diff
@@ -0,0 +1,72 @@
...
+		</li>
+		<li>
+		    Multiple [arg] can be used, separated by spaces. This will search for all
+		    collections ID and item ID containing all arguments.
+		</li>
+		<li>
+		    ["arg"] will make an exact search on the Item ID field.
+		</li>
+        <li>
+            ["collectionID/itemID/"] will make an exact search on a whole item path.
```
boecklic, line 17: dito

boecklic (changes requested):

app/stac_api/models.py
ltkum: I put it on the help_text argument in the models, since all other models with help texts have them there. (line 332 and 509)

app/stac_api/templates/js/admin/asset_help_search.js
ltkum: My "problem" is that all texts are slightly different to account for small variations in how the search work depending on what you can search for. That being said, I think it would be possible to reuse the collections and items text. I'll look into it (the asset text is used only in the general asset page)

app/stac_api/models.py
```diff
@@ -41,6 +41,35 @@
...
+                              the specified collection.</li></ul> Examples : <ul> <li>
+                              Searching for "pixelkarte" will return all collections which have
+                               pixelkarte as a part of their collection ID </li> <li>
+                              Searching for "pixelkarte 2016 4" will return all collection 
+                              which have pixelkarte, 2016 AND 4 as part of their collection ID
+                              </li><li> Searching for "ch.swisstopo.pixelkarte.example" will 
+                              yield only this collection, if this collection exists. Please 
+                              note that it would not return a collection named 
+                              ch.swisstopo.pixelkarte.example.2.</li> </ul></div>'''
+
```
ltshb, line 72: Because the multi-line seems to work here, I would suggest to do some styling here in order to make the code easier to read. When running locally it seems that the styling is not applied to this help text, like it is in the search hove ?

boecklic (approved):

ltshb (approved):

PR #241, opened 2021-03-01
add url query to logs

boecklic:
The url query params should appear in the logs as well.

### DIFF
--- a/app/middleware/logging.py
+++ b/app/middleware/logging.py
@@ -8,6 +8,8 @@ logger = logging.getLogger(__name__)
 
 
 class RequestResponseLoggingMiddleware:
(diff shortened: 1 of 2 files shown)


rebert (approved):

hansmannj (approved):
:+1:

ltshb (approved):
If this works and has been tested on DEV, I'm ok

boecklic:
I haven't found a way to obtain the raw query string from the request object, so something like this in the logging-cfg
```
  django:
    (): logging_utilities.filters.django_request.JsonDjangoRequest
    include_keys:
      - request.path
      - request.method
      - request.headers
      - request.raw_query_string
```
is not possible. Hence I'll leave it like it is.

PR #242, opened 2021-03-02
Bug fix BGDIINF_SB-1690: raise error on invalid query parameters

hansmannj:
The `collections` and `items` endpoints currently did not return errors on GET requests containing invalid parameters, e.g. on
https://service-stac.dev.bgdi.ch/api/stac/v0.9/collections/ch.swisstopo.swissbuildings3d_2/items?limit=50&random_nonsense_parameter=blablablubb or
https://service-stac.dev.bgdi.ch/api/stac/v0.9/collections?no_limits=50
This fix now raises a ValidationError in case invalid query parameters are found.

The corresponding unit tests are added, too.

### DIFF
--- a/app/stac_api/validators_serializer.py
+++ b/app/stac_api/validators_serializer.py
@@ -208,6 +208,25 @@ def _validate_asset_file_checksum(href, expected_multihash, asset_multihash):
         )
 
 
+def validate_collection_item_endpoint_get_requests(view, request):
+
+    queried_parameters = list(request.GET.copy().keys())
+
+    if type(view).__name__ == "CollectionList":
+        accepted_query_parameters = ['cursor', 'limit']
+    elif type(view).__name__ == "ItemsList":
+        accepted_query_parameters = ['bbox', 'cursor', 'datetime', 'limit']
+
+    # make sure that all queried_parameters are in the accepted_query_parameters
+    if not all(parameter in accepted_query_parameters for parameter in queried_parameters):
+        wrong_query_parameters = np.setdiff1d(queried_parameters,
+                                              accepted_query_parameters).tolist()
+        logger.error('Query contains the non-allowed parameter(s): %s', str(wrong_query_parameters))
+        message = f"The query contains the following non-queriable" \
+        f"parameter(s): {str(wrong_query_parameters)}."
+        raise ValidationError(code='query-invalid', detail=_(message))
+
+
 class ValidateSearchRequest:
     '''Validates the query parameter for the search endpoint.
 
--- a/app/stac_api/validators_serializer.py
+++ b/app/stac_api/validators_serializer.py
@@ -209,6 +209,18 @@ def _validate_asset_file_checksum(href, expected_multihash, asset_multihash):
 
 
 def validate_collection_item_endpoint_get_requests(view, request):
+    '''Check GET requests on the collections and items endpoints (for the
+    CollectionList and ItemsList view) for invalid query parameters. Raise a
+    Validation error in case of invalid parameters.
+    Args:
+        view: view instance
+            view instance of the calling view
+        request: RequestDict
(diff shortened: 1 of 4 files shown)


ltshb (changes requested):
I would not add such validator, for the following reasons:
1. From an architecture point of view, validation is done either in the model, or in the serializer or in the manager (queryset). We don't have any validation at the view level
2. In the validation functions, the query are copied, which is not very efficient
3. I would differentiate between query parameter with invalid value, which should return an error and invalida parameter (e.g. `foo=bar`). In the second case most of the API are just ignoring such parameters, they could even be simply blocked by the CDN like Cloudfront which might not forward unknown query parameter.
4. Because invalid query parameter (`foo=bar`) are not whitelisted in CF we might have different behavior when the request is handled from cache or from origin.
I would revert and leave it as it is, doing nothing if providing unknown parameter is not too bad in my opinion, I first thought it was a bug because we were giving a correct parameter with incorrect value. But in this case the Items and Collections endpoint don't have any parameter.

hansmannj:
> I would not add such validator, for the following reasons:
> 
>     1. From an architecture point of view, validation is done either in the model, or in the serializer or in the manager (queryset). We don't have any validation at the view level
> 
>     2. In the validation functions, the query are copied, which is not very efficient
> 
>     3. I would differentiate between query parameter with invalid value, which should return an error and invalida parameter (e.g. `foo=bar`). In the second case most of the API are just ignoring such parameters, they could even be simply blocked by the CDN like Cloudfront which might not forward unknown query parameter.
> 
>     4. Because invalid query parameter (`foo=bar`) are not whitelisted in CF we might have different behavior when the request is handled from cache or from origin.
>        I would revert and leave it as it is, doing nothing if providing unknown parameter is not too bad in my opinion, I first thought it was a bug because we were giving a correct parameter with incorrect value. But in this case the Items and Collections endpoint don't have any parameter.

To 1: I implemented the validation in analogy to the SearchListView's validation. That's why its in the view level, but I could also move that elsewhere.

To 3: Invalid values are already handled, this is only about invalid parameters. Simply ignoring could be an option, but I thought, it would be "nice" of us, to return an error instead. E.g. when someone has a spell error when trying to filter e.f. for `datime` instead of `datetime`. In this case simply the first 100 items of all collections will be returned, instead of an error. The user gets data but the data doesn't match the filter he applied. If he's lucky, the user will notice that by himself. Otherwise he'll work with wrong data.. That's why I would opt for returning an error rather than ignoring and just returning "something".

IsabelleBzr (commented):
I agree with the comment Jürgen made. It would be "nice" of us to send an error message. We are already doing that in search, why also not expand for the collections and items endpoint

ltshb:
The issue that I see is with CloudFront caching, looking with these examples below:

1. GET /collections?no_limit=50 => return HTTP 400 and the result is cached by CF under the key `GET /collections`
2. GET /collections => return 400 from cached answer above

Other example

1. GET /collections => return HTTP 200, and the result is cached by CF under the key `GET /collections`
2. GET /collections?no_limit=50 => return 200 from cached answer above.

So if you want to correctly return an error with invalid/unknown query parameter without issue with caching, we need to change the CF caching configuration to cache every query parameters, which will reduce the CF caching performance and cache Hit ratio. I agree that this PR would be nice for the user but it will have impact on the performance and therefore it is in my opinion not worth it. @boecklic and @sontags what are your opinion ?

ltshb:
The cache issue already on prod ! See https://jira.swisstopo.ch/browse/BGDIINF_SB-1700

hansmannj:
After discussions in the service-stac DEV team, it was decided to not implement this PR, hence I will close this now.
Implementation would negatively influence caching performance/behaviour and also potentially open DOS vectors.

PR #243, opened 2021-03-05
BGDIINF_SB-1700 removed check for invalid query parameters and accord…

hansmannj:
As discussed yesterday:

* removed the check for invalid query parameters on search endpoint, as it potentially opens DOS vectors and negatively influences performance and behaviour of caching
* also removed the according unit tests


app/stac_api/validators_serializer.py
```diff
@@ -246,20 +245,6 @@ def validate(self, request):
         '''
         # harmonize GET and POST
         query_param = harmonize_post_get_for_search(request)
```
ltshb, line 247: Pay attention, the `harmonize_post_get_for_search()` method returns the query either from the URL query or from the POST data ! Not sure it is a good idea to also remove this check in the POST data ? What would happen if we provide a POST with invalid attributes ?

hansmannj:
The check for invalid query parameters for POST requests on search endpoint was re-added again (also the corresponding unit tests). We check the payload of POST requests to other endpoints, too. So to be consistent and also, as it does not cause any troubles for caching, for POST requests this could be re-added.

app/tests/test_search_endpoint.py
```diff
@@ -139,14 +107,51 @@ def test_query_multiple_non_allowed_parameters(self):
...
+        )
+        self.assertStatusCode(200, response)
+        json_data_url = response.json()
+
+        # compare limit in payload and limit in url
+        self.assertEqual(
+            json_data_payload['features'][0]['properties']['created'],
+            json_data_url['features'][0]['properties']['created'],
+            msg="Limit parameter in payload and in URL yielded different responses."
+        )
```
ltshb, line 153: This assert is not very strong and too hazardous (could be two different features that have been created at the same time). Both request above MUST returns exactly the same result ! So checking the whole response would be much better; `self.assertEqual(json_data_payload
, json_data_url)`. Also the goal of the test should be to check that both we can use `limit` either in payload or in query, however in this test we send it first only in payload and then both in payload and in query, I would in the second request only send limit in the query and not in the payload.

app/stac_api/utils.py
```diff
@@ -183,6 +190,25 @@ def harmonize_post_get_for_search(request):
...
+        ):
+            wrong_query_parameters = np.setdiff1d(
+                list(query_param.keys()), accepted_query_parameters
+            ).tolist()
+            logger.error(
+                'Query contains the non-allowed parameter(s): %s', str(wrong_query_parameters)
+            )
+            message = f"The query contains the following non-queriable" \
+            f"parameter(s): {str(wrong_query_parameters)}."
+            raise ValidationError(code='query-invalid', detail=_(message))
```
ltshb, line 210: For code clarity and avoiding clash between django ValidationError and Rest Framework ValidationError this part of the code should be moved inside a function and put into `validators_serializer.py` we should try to keep all validation code inside the validations modules.

hansmannj: `Validators_serializer.py` is where it originally was, before you suggested to move it here ;-) But I can as well move it back again

hansmannj: Makes sense to put it in a separate function anyways

ltshb: I originally propose to change the code execution context into harmonize_post_get_for_search
 ;-).

hansmannj: I see, makes sense, I'll implement it that way this morning. Function in validations.py, call here. 
😀

app/tests/test_search_endpoint.py
hansmannj: Good point, I'll implement it like that

hansmannj: I probably misunderstand you here. In the second test it is already implemented as you suggest. `Limit` is only in the URL and not in the payload (lines 134:141 don't contain a `limit`). Or what else do you mean?

ltshb: Oops my bad I didn't pay attention that the limit was not part anymore of the payload in the second request, I missed out the second payload definition in line 135. But this is dangerous, I would suggest to have the base payload in a variable without limit and using a dictionary instead of a string. Then for the first request you can make a copy of the payload and add the limit to it. I would also directly pass the dictionary to request and not a string with `json` attribute: ```
requests.post(url, json={'limit': 1, ...})
```
instead of 
```
requests.post(url, data='{"limit": 1}', content_type="application/json")
```
See https://requests.readthedocs.io/en/master/user/quickstart/#more-complicated-post-requests

Comment (hansmannj) [reply app/tests/test_search_endpoint.py:153]: * `self.assertEqual(json_data_payload , json_data_url, msg=...)` would not work, as the `next links` and the `timestamps` would differ. I changed it to compare the item `id`. As it is unique, this assertion should do, I guess.
* using `json=...` in a requests.post does not work in the Django test client. But with `content_type=application/json` the `data=` will be serialized using `json.dumps()`.

File: app/stac_api/validators_serializer.py [hunk 4]
```diff
@@ -521,3 +509,36 @@ def validate_intersects(self, geojson):
...
+        '''
+        accepted_query_parameters = [
+            "bbox", "collections", "datetime", "ids", "intersects", "limit", "query"
+        ]
+        # make sure that all queried parameters (list(query_param.keys())) are in
+        # the accepted_query_parameters
+        if not all(
+            parameter in accepted_query_parameters for parameter in list(query_param.keys())
+        ):
+            wrong_query_parameters = set(list(query_param.keys())
```
boecklic, line 530: `set(query_param.keys())` and even `set(query_param)` is sufficient here

app/stac_api/validators_serializer.py
```diff
@@ -521,3 +509,36 @@ def validate_intersects(self, geojson):
...
+            query_param: dict
+                Copy of the harmonized QueryDict
+        '''
+        accepted_query_parameters = [
+            "bbox", "collections", "datetime", "ids", "intersects", "limit", "query"
+        ]
+        # make sure that all queried parameters (list(query_param.keys())) are in
+        # the accepted_query_parameters
+        if not all(
+            parameter in accepted_query_parameters for parameter in list(query_param.keys())
```
boecklic, line 528: `... for parameter in query_param.keys():` and even `... for parameter in query_param:` is sufficient (no need for the `list`)

app/stac_api/validators_serializer.py
```diff
@@ -521,3 +509,36 @@ def validate_intersects(self, geojson):
...
+                'Query contains the non-allowed parameter(s): %s', list(wrong_query_parameters)
+            )
+            self.errors.update(
+                {
+                    wrong_query_param:
+                    _(
+                        f"The query contains the following non-queriable parameter: " \
+                            f" {wrong_query_param}."
+                    )
+                    for wrong_query_param in list(wrong_query_parameters)
```
boecklic, line 542: also here I think the `list` is not necessary

boecklic (commented):
There are a few unnecessary `list()` conversions, otherwise nice!

app/stac_api/validators_serializer.py
ltshb, line 542: This could/should be much simpler
```python
wrong_query_parameters = set(query_param.keys()).difference(set(accepted_query_parameters))
if wrong_query_parameters:
    self.errors.update({
        invalid_query: 
           _(f"The query contains the following invalid query parameters: {','.join(wrong_query_parameters)}")
    })
    logger.error('Query contains the non-allowed parameter(s): %s', wrong_query_parameters)
```

ltshb (commented):
A part of the comment below it is fine but before going further here I would wait on the decision of https://jira.swisstopo.ch/browse/BGDIINF_SB-1715

app/tests/test_search_endpoint.py
```diff
@@ -139,14 +107,38 @@ def test_query_multiple_non_allowed_parameters(self):
...
+        self.assertStatusCode(200, response)
+        json_data_url = response.json()
+
+        # compare limit in payload and limit in url
+        self.assertEqual(
+            json_data_payload['features'][0]['id'],
+            json_data_url['features'][0]['id'],
+            msg="Limit parameter in payload and in URL yielded different responses."
+        )
```
ltshb, line 140: Here I would make a stronger comparison
```python
self.assertEqual(json_data_payload['features'], json_data_url['features'])
```
To make sure that both answer are exaclty the same, not only checking the first item id

app/stac_api/validators_serializer.py
hansmannj: Thanks, the list() conversions evolved with the code and I did not simplify in the end.

ltshb (approved):
:+1:

PR #245, opened 2021-03-10
BGDIINF_SB-1711: Fixed the POST /search pagination

ltshb:
The `POST /search` did not worked as defined in the spec. The original body with the DB query needs to be passed to the next request. 
To be consistent the spec changed as follow:

`GET /search`:
  - DB query in URL query
 - limit and cursor for pagination are part of the URL query and are set to the link['href']

`POST /search`:
 - DB query is in the POST body/payload as JSON
 - limit and cursor for pagination are also in the POST body/payload and only accepted there (ignored if passed in the URL queryparam)

Test link: https://service-stac.dev.bgdi.ch/api/stac/v0.9/search
API Spec test link: https://service-stac.dev.bgdi.ch/api/stac/static/spec/v0.9/api.html

The last commit also fixed a missleading CRITICA log in unittest:

    CRITICAL - stac_api.apps: KeyError('collection_name') ...

This message was due to the TestHttp500 view wich raised a KeyError
before raising the expected AttributeError

### DIFF
--- a/app/stac_api/pagination.py
+++ b/app/stac_api/pagination.py
@@ -9,9 +9,9 @@ from rest_framework import pagination
 from rest_framework.exceptions import ValidationError
 from rest_framework.request import Request
 from rest_framework.response import Response
-from rest_framework.utils.urls import remove_query_param
 
-from stac_api.utils import get_query_param
+from stac_api.utils import get_query_params
+from stac_api.utils import remove_query_params
 
 logger = logging.getLogger(__name__)
 
@@ -77,7 +77,7 @@ class CursorPagination(pagination.CursorPagination):
                 extra={'request': request}
             )
             raise ValidationError(
-                _('limit query parameter to small, must be in range 1..%d') % (self.max_page_size),
+                _('limit query parameter too small, must be in range 1..%d') % (self.max_page_size),
                 code='limit'
             )
         if self.max_page_size and page_size > self.max_page_size:
@@ -88,7 +88,7 @@ class CursorPagination(pagination.CursorPagination):
                 extra={'request': request}
             )
             raise ValidationError(
-                _('limit query parameter to big, must be in range 1..%d') % (self.max_page_size),
+                _('limit query parameter too big, must be in range 1..%d') % (self.max_page_size),
                 code='limit'
             )
 
@@ -135,12 +135,17 @@ class GetPostCursorPagination(CursorPagination):
 
     def patch_link(self, link, request):
         if link and request and request.method == 'POST':
-            cursor = get_query_param(link['href'], self.cursor_query_param)
-            limit = get_query_param(link['href'], self.page_size_query_param)
-            if cursor:
-                link['href'] = remove_query_param(link['href'], self.cursor_query_param)
-            if limit:
-                link['href'] = remove_query_param(link['href'], self.page_size_query_param)
+            cursor, limit = get_query_params(
+                link['href'], [self.cursor_query_param, self.page_size_query_param]
+            )
+            if cursor or limit:
+                link['href'] = remove_query_params(
+                    link['href'],
+                    [
+                        self.page_size_query_param if limit else None,
+                        self.cursor_query_param if cursor else None
+                    ]
+                )
             body = {}
             if limit:
                 body['limit'] = limit[0]
--- a/app/stac_api/pagination.py
+++ b/app/stac_api/pagination.py
@@ -1,0 +1,81 @@
+import logging
+
+from django.conf import settings
+from django.utils.translation import gettext_lazy as _
+
+from rest_framework import pagination
+from rest_framework.exceptions import ValidationError
+from rest_framework.response import Response
+
+logger = logging.getLogger(__name__)
+
+
+class CursorPagination(pagination.CursorPagination):
+    ordering = 'id'
+    page_size_query_param = 'limit'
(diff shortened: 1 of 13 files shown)


rebert (approved):
thanks for this fix

IsabelleBzr (approved):
Good documentation, I could understand the changes well :+1:

app/stac_api/pagination.py
```diff
@@ -0,0 +1,150 @@
...
+        return self.patch_link(next_link, request)
+
+    def get_previous_link(self, request=None):
+        previous_link = super().get_previous_link(request)
+        return self.patch_link(previous_link, request)
+
+    def patch_link(self, link, request):
+        if link and request and request.method == 'POST':
+            cursor = get_query_param(link['href'], self.cursor_query_param)
+            limit = get_query_param(link['href'], self.page_size_query_param)
```
hansmannj, line 139: Not sure, if it would have a significant influence on performance, just a thought:
Would it make sense, to make `get_query_param` accept a list of keys instead of a single key?
Then you'd only need one call here, with [`cursor`, `limit`] and in `get_query_param()` the parsing of the url and the query would only need to be done once.

app/stac_api/views.py
```diff
@@ -527,7 +529,8 @@ def delete(self, request, *args, **kwargs):
         return self.destroy(request, *args, **kwargs)
 
 
-class TestHttp500(AssetDetail):
+class TestHttp500(generics.GenericAPIView):
+    queryset = LandingPage.objects.all()
```
hansmannj, line 533: :+1:
Guess this stops the `CRITICAL` in the log when running tests

app/tests/test_search_endpoint.py
```diff
@@ -17,6 +20,153 @@
...
+        json_data_previous = response.json()
+        self.assertEqual(len(json_data_previous['features']), limit)
+
+        # make sure the previous data is identical to the first page
+        self.assertEqual(
+            json_data_previous['features'],
+            json_data['features'],
+            msg='previous page should be the same as the first one'
+        )
+
```
hansmannj, line 168: Probably it might make sense to also implement this in `check_next_and_previous_link()` in the e2e tests in here:
https://github.com/geoadmin/infra-e2e-tests/blob/master/tests/test_service_stac/test_pagination.py

app/tests/test_search_endpoint.py
```diff
@@ -122,176 +267,113 @@ def test_limit_in_post(self):
...
-        )
-        self.assertStatusCode(200, response)
-        json_data_url = response.json()
-
-        # compare limit in payload and limit in url
         self.assertEqual(
-            json_data_payload['features'],
-            json_data_url['features'],
-            msg="Limit parameter in payload and in URL yielded different responses."
+            len(json_data_payload['features']), limit, msg="More than one item returned."
```
hansmannj, line 271: Might be safer to have `limit` as a variable in `msg`, something like: "More than {limit} item(s) returned.", as `limit` is used as a variable in the payload and not hardcoded?

app/stac_api/pagination.py
```diff
@@ -0,0 +1,150 @@
...
+            )
+
+        if page_size <= 0:
+            logger.error(
+                'Invalid query parameter limit=%d: negative number not allowed',
+                page_size,
+                extra={'request': request}
+            )
+            raise ValidationError(
+                _('limit query parameter to small, must be in range 1..%d') % (self.max_page_size),
```
hansmannj, line 80: Its a copy&pasted message here, I know, but being picky here, we could use the chance and correct `to small` to `too small"

app/stac_api/pagination.py
```diff
@@ -0,0 +1,150 @@
...
+            )
+        if self.max_page_size and page_size > self.max_page_size:
+            logger.error(
+                'Invalid query parameter limit=%d: number bigger than the max size of %d',
+                page_size,
+                self.max_page_size,
+                extra={'request': request}
+            )
+            raise ValidationError(
+                _('limit query parameter to big, must be in range 1..%d') % (self.max_page_size),
```
hansmannj, line 91: same here: `too big` instead of `to big`

hansmannj (approved):
Very nice, thanks :+1:
Just a few minor comments

app/stac_api/views.py
ltshb: Not really it always have a CRITICAL message because that what we want to test, but the message would be about the `AttributeError('test exception')`

app/stac_api/pagination.py
ltshb: Good idea I'll modify the function as follow:
```python
def get_query_params(url, keys):
    '''Get URL query parameters by keys

    Args:
        url: string
            url to parse and retrieve query parameter
        keys: string | [string]
            query parameter key(s) to retrieve
    Returns: string | [string]
        Query parameter value(s)
    '''
    (scheme, netloc, path, query, fragment) = parse.urlsplit(url)
    query_dict = parse.parse_qs(query, keep_blank_values=True)
    if isinstance(keys, str):
        return query_dict.get(keys, None)
    return [query_dict.get(key, None) for key in keys]
```

app/stac_api/views.py
hansmannj: OK, but probably it sounds less "panicy" now :wink:

hansmannj (approved):
:+1:
Nice, also the "nicifications" in the unit test :smile:

PR #247, opened 2021-03-11
BGDIINF_SB-1715: Removed the Browsable API view

ltshb:
This view has currently a bug in the /search endpoint, where the POST
form is wrong, it is using the POST form of the /items view. This is due
to the fact that we reuse the ItemSerializer for the /search view.

Because the browsable view doesn't bring much advantages and because we
know that DRF has a performance issue and that we might have migrate the
views in other serializer framework, we decided to simply remove it for
now as it not yet very known to the customer.


hansmannj (approved):
:+1:

PR #249, opened 2021-03-15
BGDIINF_SB-1646: 1646 admin interface item filtering

hansmannj:
using https://github.com/farhan0581/django-admin-autocomplete-filter to allow filtering for items of a specified collection.
This adds a dropdown list with available collection names as well as an autocomplete textfield for the search.

PR #251, opened 2021-03-17
Updated pylint and fixed duplicate-code, raise-missing-from

ltshb:
In PR #249, django has been updated and it triggered a lots of no-member
pylint errors when referencing django models. This can be solved by
updating pylint to a version >= 2.6.

The update of 2.6 added a new raise-missing-from check which produced some
errors in our current code. As this check is very usefull, see also
https://stefan.sofa-rockers.org/2020/10/28/raise-from/ I fixed these
errors properly. Note I used `from None` for the ValidationError as
these ones are caught and translated into HTTP 400 errors, they don't
have any backtrace.

The current pylint version 2.7.0-2.7.2 have fixed an issue with the
similarities check, where the check was not done at all when runnings
pylint --jobs=2+, which we do. This showed several duplicate-code issue,
some of them where very relevant, like the pagination duplicate unittest
code and also the bbox conversion to geometry that I moved into a
functions. Unfortunately these pylint version still suffers from a bug
concerning the similarities, the config min-similarity-lines is ignored
! For our case the default value of 4 is too small and this cause some
false positive, testing pylint version 2.6.2 --jobs=1 showed that having
a min-similarity-lines=8 don't have false positive, therefore I change
it to this value which seems to me a convenable value.
We need to see then in the next 2.7 release if the bug is fixed or if we
need to either add the next 2.7 to the ignore list or totally disable
the duplicate-code checker.

I personally find this checker very good as it allow to improves code
and I would try to keep it, unfortunately, currently the only way to
have it working correctly would be to set the --job=1 and then the
lintings takes ~50s instead of ~15s...

### DIFF
--- a/app/stac_api/utils.py
+++ b/app/stac_api/utils.py
@@ -3,6 +3,7 @@ import json
 import logging
 from datetime import datetime
 from datetime import timezone
+from decimal import Decimal
 from urllib import parse
 
 import boto3
@@ -10,6 +11,8 @@ import multihash
 from botocore.client import Config
 
 from django.conf import settings
+from django.contrib.gis.geos import Point
+from django.contrib.gis.geos import Polygon
 
 logger = logging.getLogger(__name__)
 
@@ -267,3 +270,27 @@ class CommandHandler():
 
     def print_error(self, message, *args):
         self.stderr.write(self.style.ERROR(message % (args)))
+
+
+def geometry_from_bbox(bbox):
+    '''Returns a Geometry from a bbox
+
+    Args:
+        bbox: string
+            bbox as string comma separated or as float list
+
+    Returns:
+        Geometry
+
+    Raises:
+        ValueError, IndexError, GDALException
+    '''
+    list_bbox_values = bbox.split(',')
+    if (
+        Decimal(list_bbox_values[0]) == Decimal(list_bbox_values[2]) and
+        Decimal(list_bbox_values[1]) == Decimal(list_bbox_values[3])
+    ):
+        bbox_geometry = Point(float(list_bbox_values[0]), float(list_bbox_values[1]))
+    else:
+        bbox_geometry = Polygon.from_bbox(list_bbox_values)
+    return bbox_geometry
--- a/app/tests/data_factory.py
+++ b/app/tests/data_factory.py
@@ -132,7 +132,7 @@ class SampleData:
         try:
             sample = self.samples_dict[sample]
         except KeyError as error:
-            raise KeyError(f'Unknown {self.sample_name} sample: {error}')
+            raise KeyError(f'Unknown {self.sample_name} sample: {error}') from None
 
         # Sets attributes from sample
         for key, value in sample.items():
--- a/app/tests/test_generic_api.py
+++ b/app/tests/test_generic_api.py
@@ -39,95 +39,116 @@ class ApiPaginationTestCase(StacBaseTestCase):
 
     @classmethod
     def setUpTestData(cls):
-        Factory().create_collection_samples(3, db_create=True)
+        cls.factory = Factory()
+        cls.collections = cls.factory.create_collection_samples(3, db_create=True)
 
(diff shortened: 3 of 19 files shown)


app/stac_api/utils.py
```diff
@@ -267,3 +270,27 @@ def print_success(self, message, *args, level=1):
 
     def print_error(self, message, *args):
         self.stderr.write(self.style.ERROR(message % (args)))
+
+
+def geometry_from_bbox(bbox):
```
rebert, line 275: nice

app/tests/data_factory.py
```diff
@@ -132,7 +132,7 @@ def __init__(self, sample, required_only=False, **kwargs):
         try:
             sample = self.samples_dict[sample]
         except KeyError as error:
-            raise KeyError(f'Unknown {self.sample_name} sample: {error}')
+            raise KeyError(f'Unknown {self.sample_name} sample: {error}') from None
```
hansmannj, line 135: Is `from None` correct here, or could/should it be `from error`?

app/tests/test_generic_api.py
```diff
@@ -39,95 +39,116 @@ class ApiPaginationTestCase(StacBaseTestCase):
...
-        )
+                # Check next link
+                next_link = get_link(json_data['links'], 'next')
+                self.assertIsNotNone(next_link, msg='Pagination next link missing')
+                self.assertTrue(isinstance(next_link['href'], str), msg='href is not a string')
+                self.assertTrue(
+                    next_link['href'].
+                    startswith(f'http://testserver/api/stac/v0.9/{endpoint}?cursor='),
+                    msg='Invalid href link pagination string'
+                )
```
hansmannj, line 97: Not really relevant for this PR, but when in a call with Isabelle about the e2e tests, we decided to implement a `_check_link()` and pass the response and the `link_rel` to it.
So the lines after the comments `#check next link` and `#check previous link` (each called twice here) can be handled by that new function.

hansmannj (approved):
:1st_place_medal: 
Thanks!

app/tests/data_factory.py
ltshb: Yes it is correct, using `from error` would  give two backtrace, one from the original exception and one from the second. As the original exception doesn't provide more information as the one we re-raise (original exception message is part of the new one), we can omit it, this would make the reading of the exception easier.

app/tests/test_generic_api.py
ltshb: good point I'll do it.

ltshb: Done see commit https://github.com/geoadmin/service-stac/pull/251/commits/fbff8f3b6855a48a662ff6911772f26e66489ae3

rebert (approved):

hansmannj (approved):
:1st_place_medal:

app/tests/test_generic_api.py
```diff
@@ -44,6 +44,32 @@ def setUpTestData(cls):
 
     def setUp(self):
         self.client = Client()
+        self.maxDiff = None  # pylint: disable=invalid-name
+
+    def _get_check_link(self, links, rel, endpoint):
```
hansmannj, line 49: nice

app/tests/test_generic_api.py
```diff
@@ -44,6 +44,32 @@ def setUpTestData(cls):
...
+        link = get_link(links, rel)
+        self.assertIsNotNone(link, msg=f'Pagination {rel} link missing')
+        self.assertTrue(isinstance(link['href'], str), msg='href is not a string')
+        self.assertTrue(
+            link['href'].startswith(f'http://testserver/api/stac/v0.9/{endpoint}?cursor='),
+            msg='Invalid href link pagination string'
+        )
+        return link
+
+    def _read_link(self, link, rel, other_pages, result_attribute):
```
hansmannj, line 59: nice, too :+1:

PR #252, opened 2021-03-17
BGDIINF_SB-1566: Added unittest for asset media type validation

ltshb:
Still waiting an answer concerning the file extension for application/vnd.mapbox-vector-tile which I set to `*.mvt` based on https://github.com/mapbox/vector-tile-spec/tree/master/2.1#21-file-extension

### DIFF
--- a/app/stac_api/serializers.py
+++ b/app/stac_api/serializers.py
@@ -24,6 +24,8 @@ from stac_api.utils import build_asset_href
 from stac_api.utils import isoformat
 from stac_api.validators import MEDIA_TYPES_MIMES
 from stac_api.validators import validate_asset_multihash
+from stac_api.validators import validate_asset_name
+from stac_api.validators import validate_asset_name_with_media_type
 from stac_api.validators import validate_geoadmin_variant
 from stac_api.validators import validate_item_properties_datetimes
 from stac_api.validators import validate_name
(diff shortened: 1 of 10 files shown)


app/tests/data_factory.py
```diff
@@ -1109,6 +1115,21 @@ def create_samples(self, samples, item, db_create=False, create_asset_file=False
             samples, item=item, db_create=db_create, create_asset_file=create_asset_file, **kwargs
         )
 
+    def get_extension(self, sample_name, kwargs):
```
rebert, line 1118: some docstring

app/stac_api/serializers.py
```diff
@@ -615,6 +617,14 @@ class Meta:
     updated = serializers.DateTimeField(read_only=True)
 
     def validate(self, attrs):
```
rebert, line 619: some docstring

rebert (approved):
tested successfully manually as well

app/tests/data_factory.py
ltshb: Its an internal method which should not be used externally, so i'll rename it to `_get_extension` instead of adding a docstring.

app/stac_api/serializers.py
ltshb: Well the name is quite explicit and it is a django method which is already documented by django, so I don't think docstring add any value here. It is and should only be called by django internals.

hansmannj (approved):
Good test coverage :+1:

PR #253, opened 2021-03-19
BGDIINF_SB-1647 filter assets for their collection and/or item

hansmannj:
Added autocomplete search fields and dropdown list filters for filtering assets for their collection and/or items.
Following behaviour is implemented:
If a collection is defined in the `filter by collection name` filter, only those items will appear
in the dropdown list of the `filter by item name` filter, that are in the currently selected collection.
ItemAdmin's `get_search_results()` was overriden with a few "hacky" lines for this.
Otherwise, all items would appear in the list, which does not make sense.

Note: The responsiveness after clicking on those filters could be better. Its a bit slow when there are many items or collections. Probably will improve this soon. The performance problem is due to the large number of assets, items and collections and was already a problem before the "hacky" lines were inserted.

P.S.: Any ideas for a cleaner approach or improving the performance are welcome :slightly_smiling_face:

### DIFF
--- a/app/stac_api/admin.py
+++ b/app/stac_api/admin.py
@@ -134,11 +134,11 @@ class ItemAdmin(admin.GeoModelAdmin):
         # make sense.
 
         # this asserts that the request comes from the autocomplete filters.
-        if "autocomplete" in request.__dict__['environ']['PATH_INFO']:
-            if "item__collection" in request.__dict__['environ']['HTTP_REFERER']:
-                current_collection_pk = request.__dict__['environ']['HTTP_REFERER'].split(
-                    'item__collection='
-                )[1].split("&")[0]
+        environ = request.environ.copy()
+        if "autocomplete" in environ['PATH_INFO']:
+            if "item__collection" in environ['HTTP_REFERER']:
+                current_collection_pk = environ['HTTP_REFERER'].split('item__collection='
+                                                                     )[1].split("&")[0]
                 queryset = self.model.objects.filter(collection__pk__exact=current_collection_pk)
         elif search_term.startswith('"') and search_term.endswith('"'):
             search_terms = search_term.strip('"').split('/', maxsplit=2)
--- a/app/stac_api/admin.py
+++ b/app/stac_api/admin.py
@@ -1,4 +1,5 @@
 from admin_auto_filters.filters import AutocompleteFilter
+from admin_auto_filters.filters import AutocompleteFilterFactory
 
 from django.contrib.gis import admin
 from django.contrib.postgres.fields import ArrayField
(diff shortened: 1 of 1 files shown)


app/stac_api/admin.py
```diff
@@ -120,11 +121,26 @@ class Media:
...
+        # dropdown list. Otherwise all items would appear in the dropdown list, which does not
+        # make sense.
+
+        # this asserts that the request comes from the autocomplete filters.
+        if "autocomplete" in request.__dict__['environ']['PATH_INFO']:
+            if "item__collection" in request.__dict__['environ']['HTTP_REFERER']:
+                current_collection_pk = request.__dict__['environ']['HTTP_REFERER'].split(
+                    'item__collection='
+                )[1].split("&")[0]
+                queryset = self.model.objects.filter(collection__pk__exact=current_collection_pk)
```
hansmannj, line 142: This queryset will serve as input for the dropdown list of the "filter by item name" filter.

app/stac_api/admin.py
```diff
@@ -120,11 +121,26 @@ class Media:
...
+
+        # The following few lines are a bit hacky and are needed for the item dropdown list
+        # to depend on the currently selected collection in the collection dropdown filter.
+        # With this "hack", only those items appear in the "filter by item name" dropdown list,
+        # that belong to the currently selected collection in the "filter by collection name"
+        # dropdown list. Otherwise all items would appear in the dropdown list, which does not
+        # make sense.
+
+        # this asserts that the request comes from the autocomplete filters.
+        environ = request.environ.copy()
```
boecklic, line 137: just wondering: why is it necessary to copy the environ here?

ltshb: It shouldn't be necessary and should be removed.

app/stac_api/admin.py
```diff
@@ -120,11 +121,26 @@ class Media:
...
+        if "autocomplete" in environ['PATH_INFO']:
+            if "item__collection" in environ['HTTP_REFERER']:
+                current_collection_pk = environ['HTTP_REFERER'].split('item__collection='
+                                                                     )[1].split("&")[0]
+                queryset = self.model.objects.filter(collection__pk__exact=current_collection_pk)
```
ltshb, line 142: Here we already have a filter to the correct items list (by item name), just filter the current queryset by collection instead of recreating a query set which doesn't take the item name into account !

ltshb (changes requested):

app/stac_api/admin.py
hansmannj: OK, I'll revert that. Just wanted to be on the safe side, in case it is treated as a stream, like request.data, but it isn't.

app/stac_api/admin.py
hansmannj: I made this part now a bit more "stable" and "nice", using `request.headers` and parsing the querystring using `get_query_params()` from our `stac_api.utils`.

ltshb (approved):

PR #254, opened 2021-03-21
BGDIINF_SB-1724 items (and also assets already) upsert

hansmannj:
Now this PR is consistent with the upsert for collections (https://github.com/geoadmin/service-stac/pull/255)
Slightly adapted `UpsertInsertModelMixin` to pass an instance and the serialize_kwargs to the serializer, in case the object already exists (upsert update) and to catch the exception, in case the object doesn't already exist. Otherwise this would lead to a crash (uncaught 404 exception) in the atomic transaction.


hansmannj:
I'll add unit tests for upserting invalid items and assets and also
add error handling for the atomic transactions in next commit

ltshb (changes requested):
The upsert should not be implemented within the `UpdateModelMixin`, this mixin is for the update model. But we need a new mixin `UpsertModelMixin` does any already exists in Rest Framework ?
Then we migth want to make use of the `get_or_create` method of the django model manager which might/could be more efficient than using `.exists()` but would anyway leave this handling to django.
Regarding the atomic transaction we should discuss it within a meeting, we aslo needs to take the rest framework custom exception handler into account, see https://stackoverflow.com/questions/52715005/atomic-transactions-in-drf-action.
If we use a correct mixin approach it doens't make sense to work all ticket for upsert in parallel...

ltshb:
To correctly test the atomic operation, an idea would be to create a new view (similar to `TestHttp500` or simply extend this one, but it could/should be moved out of the regular views.py module maybe in views_test.py that is only added in `DEBUG=True`). This  view could be used to test an insert and update in DB and before returning the response simply raise a HTTP 500, then the unittest can check that the ressource was not created respectively not modified.

app/stac_api/views_mixins.py
```diff
@@ -57,28 +59,46 @@ class UpdateModelMixin:
...
+
+        # else create
+        serializer = self.get_serializer(data=data)
+        serializer.is_valid(raise_exception=True)
+        self.perform_upsert(serializer)
+        headers = self.get_success_headers(serializer.data)
+        return Response(serializer.data, status=status.HTTP_201_CREATED, headers=headers)
 
-    def perform_update(self, serializer):
+    def perform_upsert(self, serializer):
```
boecklic, line 89: What's the reason for putting this one call into a dedicated function?

app/stac_api/views.py
```diff
@@ -491,10 +499,18 @@ def get_queryset(self):
...
+        return queryset.exists()
+
+    def get_object(self, called_from_within_atomic=False):  # pylint: disable=arguments-differ):
+        asset_name = self.kwargs.get(self.lookup_url_kwarg)
+        queryset = self.get_queryset().filter(name=asset_name)
+        if not called_from_within_atomic:
+            obj = get_object_or_404(queryset)
+        else:
+            obj = Asset.objects.get(name=asset_name)
         return obj
```
boecklic, line 514: I believe you can use `get_object_or_404` in an atomic transaction, so imho there's no need to distinguish behaviour here. Am I wrong?

boecklic (commented):
I'm not sure you need to explicitly check for existence, see comments

app/stac_api/views_mixins.py
hansmannj: The reason is to be consistent with the way, it currently is implemented. Currently, there's the `perform_update()`, which I simply renamed to `perform_upsert()`. And the currently implemented way is simply an overriden original mixin. So the `perform_XYZ()` comes from the original Django mixins.

app/stac_api/views.py
hansmannj: Probably you're right. This is just, because I wasn't sure how to handle exceptions within an atomic context, yet. Probably it is ok, if the exceptions are already handled inside the atomic block. But I thought, they need to be handled in a special way. I'll check that

hansmannj:
> I'm not sure you need to explicitly check for existence, see comments

Currently we need to explicitly check for existence, as the serializer needs to be passed the `instance` and the `serializer_kwargs` in case of an already existing instance, and only the `data` in case of a non-existing object.
There might as well be other solutions for this, but it seems to work :-)

hansmannj:
Currently the unit tests run locally but crash during build. I'll dig deeper here, but perhaps woon't fix this today. Seems to be related to the tests for race conditions..

app/stac_api/views_mixins.py
```diff
@@ -86,7 +86,14 @@ def update(self, request, *args, **kwargs):
...
     def upsert(self, request, *args, **kwargs):
         data = self.get_write_request_data(request, *args, **kwargs)
-        serializer = self.get_serializer(data=data)
+        if self.check_existence():
+            instance = self.get_object()
+            partial = kwargs.pop('partial', False)
+            serializer_kwargs = {'partial': partial}
+            serializer = self.get_serializer(instance, data=data, **serializer_kwargs)
+        else:
+            serializer = self.get_serializer(data=data)
```
ltshb, line 95: That's what we need to avoid here by using the django update_or_create() method. This code above don't treat race condition, and thats probably why the unittest are failing. Moreover the `check_existence()` is not correct as it returns a filter (`exists()` missing). To achieve our goal we don't want to provide the instance to the serializer.

ltshb (changes requested):

app/stac_api/views_mixins.py
hansmannj: I see, I'll revert the check_existence(). And yes, I forgot to add the `.exists()`, thanks. But the unit tests were failing also before, that's why I re-added the `check_existence()` and also passed the instance to the serializer. Without this, the tests sometimes fail and sometimes pass. But also with this. So there must be something else. I am on it :-)

app/tests/test_items_endpoint.py
```diff
@@ -648,25 +656,155 @@ def test_item_endpoint_patch_invalid_datetimes(self):
...
+    def setUp(self):
+        self.username = 'user'
+        self.password = 'dummy-password'
+        get_user_model().objects.create_superuser(self.username, password=self.password)
+
+    def test_item_upsert_race_condition(self):
+        workers = 5
+        status_201 = 0
+        collection_sample = CollectionFactory().create_sample(sample='collection-2')
+        item_sample = ItemFactory().create_sample(collection_sample.model, sample='item-1')
```
ltshb, line 750: Same as for asset unittest, make use of `setUpTestData()`

app/tests/test_items_endpoint.py
```diff
@@ -648,25 +656,155 @@ def test_item_endpoint_patch_invalid_datetimes(self):
...
+                msg=f'Unexpected response status code {response.status_code} for worker {worker}'
+            )
+            self.check_stac_item(item_sample.json, response.json(), collection_sample['name'])
+        self.assertEqual(status_201, 1, msg="Not only one upsert did a create !")
+
+    def test_item_post_race_condition(self):
+        workers = 5
+        status_201 = 0
+        collection_sample = CollectionFactory().create_sample(sample='collection-2')
+        item_sample = ItemFactory().create_sample(collection_sample.model, sample='item-1')
```
ltshb, line 781: Same as above: `setUpTestData`

ltshb:
Looking at the unittest I think there is a bug where you can create an asset/item using the PUT or POST even if the underlying parent (collection and/or item) doesn't exists ! Or if not there is something not working as expected with the unittest and the data factory.

ltshb:
@hansmannj here below is a proposal to solve the non existing collection issue:
```python
# serializer.py
class UpsertModelSerializerMixin:
    def upsert(self, look_up, **kwargs):
        """
        Update or insert an instance and return it.

        Args:
            look_up: dict
                Must be a unique query to be used in the objects.update_or_create(**look_up) method. 
            **kwargs:
                Extra key=value pairs to pass as validated_data to update_or_create(). For example relationship
                that are not serialized but part of the request path can be given as kwargs.
        """
        validated_data = {**self.validated_data, **kwargs}
        self.instance, created = self.update_or_create(look_up, validated_data)
        return self.instance, created

    def update_or_create(self, look_up, validated_data):
        """This method must be implemented by the serializer and must make use of the DB
        objects.update_or_create() method.

        Args:
           look_up: dict
                Must be a unique query to be used in the objects.update_or_create(**look_up) method. 
            validated_data: dict
                Copy of the validated_data to be used as defaults in the objects.update_or_create(defaults=validated_data) method.
        """
        raise NotImplementedError("update_or_create() not implemented")

class CollectionSerializer(NonNullModelSerializer, UpsertModelSerializerMixin):
    def update_or_create(self, look_up, validated_data):
        ...
        collection, created = Collection.objects.update_or_create(**look_up, defaults=validated_data)
        ...


class AssetSerializer(NonNullModelSerializer, UpsertModelSerializerMixin):
    # remove the `item = serializers.SlugRelatedField()` field !!!
    def update_or_create(self, look_up, validated_data):
        ...
        asset, created = Asset.objects.update_or_create(**look_up, defaults=validated_data)
        ...

class ItemSerializer(NonNullModelSerializer, UpsertModelSerializerMixin):
    # remove the `asset = serializers.SlugRelatedField()` field !!!
    def update_or_create(self, look_up, validated_data):
        ...
        item, created = Item.objects.update_or_create(**look_up, defaults=validated_data)
        ...

# views_mixins.py
class UpdateInsertModelMixin:
    def perform_upsert(self, serializer, lookup):
        return serializer.upsert(lookup)

# views.py

class ItemsList(generics.GenericAPIView, views_mixins.CreateModelMixin):
    # remove `def get_write_request_data(self, request, *args, **kwargs):`
   
   def perform_create(self, serializer):
        # this DB hit used to done by the serializer due to the SlugRelatedField during deserialization
        collection = Collection.objects.get(name=self.kwargs['collection_name'])
        serializer.save(collection=collection)

class ItemDetail():
   # remove `def get_write_request_data(self, request, *args, **kwargs):`

   def perform_update(self, serializer):
        collection = Collection.objects.get(name=self.kwargs['collection_name'])
        serializer.save(collection=collection)

    def perform_upsert(self, serializer, lookup):
        collection = Collection.objects.get(name=self.kwargs['collection_name'])
        return serializer.upsert(lookup, collection=collection)

class AssetsList(generics.GenericAPIView, views_mixins.CreateModelMixin):
    # remove `def get_write_request_data(self, request, *args, **kwargs):`
   
   def perform_create(self, serializer):
        # this DB hit used to done by the serializer due to the SlugRelatedField during deserialization
        item = Item.objects.get(collection__name=self.kwargs['collection_name'], name=self.kwargs['item_name'])
        serializer.save(item=item)

class AssetDetail():
   # remove `def get_write_request_data(self, request, *args, **kwargs):`

   def perform_update(self, serializer):
        item = Item.objects.get(collection__name=self.kwargs['collection_name'], name=self.kwargs['item_name'])
        serializer.save(item=item)

    def perform_upsert(self, serializer, lookup):
        item = Item.objects.get(collection__name=self.kwargs['collection_name'], name=self.kwargs['item_name'])
        return serializer.upsert(lookup, item=item)
```
Off course the code above lack of proper error handling, for example we need to check for `DoesNotExist` when performing an `objects.get()` and return an HTTP 404 in this case, this could be wrapped with the `get_or_404()` method. But this approach should solve the issue in a clean manner. Note the `def get_write_request_data()` has been added by me to solve the relationship issue, but after looking deeper in code I found out that this kind of relationship should be solved using the `**kwargs` in the `serializer.save()` method, at least it is my understanding of DRF.

hansmannj:
Nice suggestion @ltshb, adding extra key-value pairs to the validated_data might do the trick.
The standard `update_or_create()` seems to have troubles with foreign keys or even foreign key's foreign keys.

app/tests/test_assets_endpoint.py
```diff
@@ -646,6 +671,155 @@ def test_asset_endpoint_patch_read_only_in_payload(self):
...
+    def setUp(self):
+        self.username = 'user'
+        self.password = 'dummy-password'
+        get_user_model().objects.create_superuser(self.username, password=self.password)
+
+    def test_asset_upsert_race_condition(self):
+        workers = 5
+        status_201 = 0
+        collection_sample = CollectionFactory().create_sample(sample='collection-2')
+        item_sample = ItemFactory().create_sample(collection_sample.model, sample='item-2')
```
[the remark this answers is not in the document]
ltshb: This comment is still valid

app/tests/test_assets_endpoint.py
```diff
@@ -646,6 +671,155 @@ def test_asset_endpoint_patch_read_only_in_payload(self):
...
+        asset_sample = AssetFactory().create_sample(item_sample.model, sample='asset-2')
```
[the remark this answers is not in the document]
ltshb: Also still valid

app/tests/test_assets_endpoint.py
```diff
@@ -646,6 +671,155 @@ def test_asset_endpoint_patch_read_only_in_payload(self):
...
+                item_sample['name'],
+                ignore=['item']
+            )
+        self.assertEqual(status_201, 1, msg="Not only one upsert did a create !")
+
+    def test_asset_post_race_condition(self):
+        workers = 5
+        status_201 = 0
+        collection_sample = CollectionFactory().create_sample(sample='collection-2')
+        item_sample = ItemFactory().create_sample(collection_sample.model, sample='item-2')
```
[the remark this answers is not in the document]
ltshb: Still valid

app/tests/test_items_endpoint.py
ltshb: Still valid

app/tests/test_items_endpoint.py
ltshb: Still valid

ltshb (approved):
Very good. Just some comment about the unittest code that could/should be simplified.

PR #255, opened 2021-03-22
BGDIINF_SB-1723: Collection upsert

ltshb:
We also make use of the DB atomic transaction to do proper DB rollback
in case of errors.

The get_object() of the views have been removed in order to use the
default implementation using the lookup_field.

Also made all endpoints atomic to assure proper db rollback in case of
error.

### DIFF
--- a/app/stac_api/serializers.py
+++ b/app/stac_api/serializers.py
@@ -463,7 +463,19 @@ class CollectionSerializer(NonNullModelSerializer, UpsertModelSerializerMixin):
 
     def update_or_create(self, validated_data, **kwargs):
         """
-        Create and return a new `Collection` instance, given the validated data.
+        Update or create the collection object selected by kwargs and return the instance.
+
+        When no collection object matching the kwargs selection, a new object is created.
+
+        Args:
+            validated_data: dict
+                Copy of the validated_data to use for update
+            kwargs: dict
+                Object selection arguments (NOTE: the selection arguments must match a unique
+                object in DB otherwise an IntegrityError will be raised)
+
+        Returns: tuple
+            Collection instance and True if created otherwise false
         """
         providers_data = validated_data.pop('providers', [])
         links_data = validated_data.pop('links', [])
--- a/app/stac_api/serializers.py
+++ b/app/stac_api/serializers.py
@@ -31,6 +31,7 @@ from stac_api.validators import validate_item_properties_datetimes
 from stac_api.validators import validate_name
 from stac_api.validators_serializer import validate_asset_file
 from stac_api.validators_serializer import validate_json_payload
+from stac_api.validators_serializer import validate_uniqueness_and_create
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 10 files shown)


app/stac_api/validators_serializer.py
```diff
@@ -527,3 +529,33 @@ def validate_query_parameters_post_search(self, query_param):
             logger.error(
                 'Query contains the non-allowed parameter(s): %s', list(wrong_query_parameters)
             )
+
+
+def validate_uniqueness_and_create(model_class, validated_data):
```
hansmannj, line 534: The idea of this method is nice and I guess there's no way around it, in order to properly treat the integrity error in the atomic context. It's just a bit "un-nice", that is validates and also creates in one step an inside the `validators_serializer.py`..

app/stac_api/serializers.py
```diff
@@ -439,6 +461,22 @@ def update(self, instance, validated_data):
         )
         return super().update(instance, validated_data)
 
+    def update_or_create(self, validated_data, **kwargs):
+        """
+        Create and return a new `Collection` instance, given the validated data.
```
hansmannj, line 466: Docstring needs to be updatd. Something like "If the collection already exists, it will be updated, otherwise a new instance will be created" or the like.

app/tests/test_collections_endpoint.py
```diff
@@ -319,13 +327,123 @@ def test_collection_patch_read_only_in_payload(self):
...
+
+
+class CollectionRaceConditionTest(StacBaseTransactionTestCase):
+
+    def setUp(self):
+        self.username = 'user'
+        self.password = 'dummy-password'
+        get_user_model().objects.create_superuser(self.username, password=self.password)
+
+    def test_collection_upsert_race_condition(self):
```
hansmannj, line 387: :+1: Nice!

app/tests/test_collections_endpoint.py
```diff
@@ -319,13 +327,123 @@ def test_collection_patch_read_only_in_payload(self):
...
+            if response.status_code == 201:
+                status_201 += 1
+            self.assertIn(
+                response.status_code, [200, 201],
+                msg=f'Unexpected response status code {response.status_code} for worker {worker}'
+            )
+            self.check_stac_collection(sample.json, response.json())
+        self.assertEqual(status_201, 1, msg="Not only one upsert did a create !")
+
+    def test_collection_post_race_condition(self):
```
hansmannj, line 417: :1st_place_medal:

hansmannj (approved):
Seems fine to me, nice implementation. And also nice having tests for race conditions :-)
Once this PR is merged, I'll make the PR for items/assets consistent with this one

app/stac_api/serializers.py
ltshb: good point

app/stac_api/validators_serializer.py
ltshb: @hansmannj  Exactly we cannot go without this function, however we could move it into serializer.py ?

hansmannj: The world will keep on turning no matter if it stays here or is moved into `serializer.py` :smile: 
Just leave it here.

PR #259, opened 2021-03-25
Improved code for auto generated links

ltshb:
The new code make use of django reverse url functionality and improved code reusability


hansmannj (approved):
Nice!

PR #263, opened 2021-03-30
BGDIINF_SB-1739: Bug fixes and improvements

ltshb:
The view validation code to check if parent exists in list views has
been moved in a separate module in order to be better reused. The
performance has been slightly improved due to the fact that we only do
one DB hit by checking the direct parent which can only exists if its
parent exists.

The Asset Etag was also not computed correctly if two asset in two
different items would have had the same name which is allowed. So now
the query to compute the asset etag takes also the parents into account.


app/stac_api/urls.py
```diff
@@ -16,36 +17,33 @@
...
 HEALTHCHECK_ENDPOINT = settings.HEALTHCHECK_ENDPOINT
 
+asset_urls = [
+    path("<asset_name>", AssetDetail.as_view(), name='asset-detail'),
+]
+
+item_urls = [
+    path("<item_name>", ItemDetail.as_view(), name='item-detail'),
+    path("<item_name>/assets", AssetsList.as_view(), name='assets-list'),
+    path("<item_name>/assets/", include(asset_urls))
```
rebert, line 27: for a human user, this makes it easier

rebert (approved):
lgtm

PR #264, opened 2021-03-30
BGDIINF_SB-1754: wrong media type 415

rebert:


ltshb (commented):
Maybe a unittest to check for 415 would be good to avoid such issue in future.

ltshb (commented):
I would not mixup unittest of the wrong media type and the preconditions ! The wrong media type unittest should be in a separate test method: test_wrong_media_type(). Done for all endpoints and methods (except for GET and DELETE).

app/tests/test_generic_api.py
```diff
@@ -274,10 +276,70 @@ def test_put_precondition(self):
...
+            (['post'], 'search', {
+                "query": {
+                    "title": {
+                        "eq": "My item 1"
+                    }
+                }
+            }),
+        ]:
+            with self.subTest(endpoint=endpoint):
+                client_requests = [getattr(self.client, i) for i in request_methods]
```
ltshb, line 335: I like that :+1: 
Just for a better code readability, I would change the variable `i` to `method` as `i` is usually for index or number.

ltshb (approved):

PR #265, opened 2021-04-06
BGDIINF_SB-1670: Correct transactional examples in STAC spec

IsabelleBzr:
Corrected the transactional examples in STAC spec:
* removed read only properties
* changed id to small letters
* changed image type
* removed non supported properties

### DIFF
--- a/spec/components/schemas.yml
+++ b/spec/components/schemas.yml
@@ -231,8 +231,6 @@ components:
           type: array
         providers:
           $ref: "#/components/schemas/providers"
-        stac_extensions:
-          $ref: "#/components/schemas/stac_extensions"
         stac_version:
           $ref: "#/components/schemas/stac_version"
         summaries:
@@ -605,8 +603,6 @@ components:
           $ref: "#/components/schemas/geometryGeoJSON"
         properties:
           $ref: "#/components/schemas/itemProperties"
-        stac_extensions:
-          $ref: "#/components/schemas/stac_extensions"
           readOnly: true
         stac_version:
           $ref: "#/components/schemas/stac_version"
@@ -868,8 +864,6 @@ components:
           items:
             $ref: "#/components/schemas/link"
           type: array
-        stac_extensions:
-          $ref: "#/components/schemas/stac_extensions"
         stac_version:
           $ref: "#/components/schemas/stac_version"
         title:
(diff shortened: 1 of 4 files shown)


ltshb (approved):
:+1:

hansmannj (commented):
:+1:

Just one little point:
We discussed this a while ago, and I can't remember exactly, what the final result was.
But this PR would be a chance to remove the `stac_extensions` from the response schema and also the response example here: https://data.geo.admin.ch/api/stac/static/spec/v0.9/apitransactional.html#operation/createCollection

Currently, `stac_extensions` is hard-coded to be empty. And as it is empty, it will not be part of the response. So we can simply delete it from the response schema and also the response example. If necessary in future, we can simply revert it and re-add it to the schema and the sample. What do you think about that @IsabelleBzr @ltshb ?

hansmannj (approved):

ltshb:
@hansmannj  I would leave the `stac_extensions` in the SPEC as it is also present in the original radianearth SPEC. It is correct to have it the SPEC and having it in the service hard coded to an empty list. However related to this there is still some clean up in unittest to do see https://jira.swisstopo.ch/browse/BGDIINF_SB-1765

hansmannj:
> @hansmannj I would leave the `stac_extensions` in the SPEC as it is also present in the original radianearth SPEC. It is correct to have it the SPEC and having it in the service hard coded to an empty list. However related to this there is still some clean up in unittest to do see https://jira.swisstopo.ch/browse/BGDIINF_SB-1765

Good point to check the radiant earth specs @ltshb. However I think it is more transparent to remove `stac_extensions` from the response schema and also the response example, as it currently will not appear in the response. So having it in the spec and not in the response would be strange.
`stac_extensions` are not a required field in the radiant earth specs. And we also have our own custom way to deal with other non-required fields (e.g. `keywords`). So I think it would be OK to remove it. What's your take @boecklic ?

boecklic:
> `stac_extensions` are not a required field in the radiant earth specs. And we also have our own custom way to deal with other non-required fields (e.g. `keywords`). So I think it would be OK to remove it. What's your take @boecklic ?
If `stac_extensions` is not required (I haven't checked) and since it's not present in the payload I'd remove it from the spec.

PR #266, opened 2021-04-07
BGDIINF_SB-1747: No more move asset operations

rebert:
Admin Interface

### DIFF
--- a/app/stac_api/admin.py
+++ b/app/stac_api/admin.py
@@ -226,6 +226,7 @@ class AssetAdmin(admin.ModelAdmin):
             # of None. We use None for empty value, None value are stripped
             # then in the output will empty string not.
             obj.description = None
+
         super().save_model(request, obj, form, change)
 
     # Note: this is a bit hacky and only required to get access
@@ -237,3 +238,12 @@ class AssetAdmin(admin.ModelAdmin):
     def href(self, instance):
         path = instance.file.name
         return build_asset_href(self.request, path)
+
+    # We don't want to move the assets on S3
+    # That's why some fields like the name of the asset are set readonly here
+    # for update operations
+    def change_view(self, request, object_id, form_url='', extra_context=None):
+        self.readonly_fields = self.get_readonly_fields(request)
+        self.readonly_fields.extend(['name', 'item'])
+
+        return super().change_view(request, object_id, form_url, extra_context)
--- a/app/stac_api/validators_view.py
+++ b/app/stac_api/validators_view.py
@@ -92,3 +92,28 @@ def validate_upload_parts(request):
         message = f'Required "parts" must be a list, not a {type(request.data["parts"])}'
         logger.error(message, extra={'request': request})
         raise ValidationError({'parts': _(message)}, code='invalid')
+
+
+def validate_renaming(serializer, id_field='', original_id='', extra_log=None):
+    '''Validate that the asset name is not different from the one defined in
+       the data.
+
+    Args:
+        serializer: serializer object
+            The serializer to derive the data from
+        id_field: string
+            The key to get the name/id in the data dict
+        original_id: string
+            The id/name derived from the request kwargs
+        extra: djangoHttpRequest object
+            The request object for logging purposes
+
+    Raises:
+        Http400: when the asset will be renamed/moved
+    '''
+    data = serializer.validated_data
+    if id_field in data.keys():
+        if data[id_field] != original_id:
+            message = 'Renaming object is not allowed'
+            logger.error(message, extra={'request': extra_log})
+            raise ValidationError(_(message), code='invalid')
--- a/app/stac_api/views.py
+++ b/app/stac_api/views.py
@@ -30,6 +30,7 @@ from stac_api.utils import utc_aware
 from stac_api.validators_serializer import ValidateSearchRequest
 from stac_api.validators_view import validate_collection
 from stac_api.validators_view import validate_item
+from stac_api.validators_view import validate_renaming
 
 logger = logging.getLogger(__name__)
 
@@ -484,11 +485,23 @@ class AssetDetail(
         item = get_object_or_404(
             Item, collection__name=self.kwargs['collection_name'], name=self.kwargs['item_name']
         )
+        validate_renaming(
+            serializer,
+            id_field='name',
+            original_id=self.kwargs['asset_name'],
+            extra_log=self.request
+        )
         serializer.save(item=item, file=get_asset_path(item, self.kwargs['asset_name']))
 
     def perform_upsert(self, serializer, lookup):
         item = get_object_or_404(
             Item, collection__name=self.kwargs['collection_name'], name=self.kwargs['item_name']
+        )
+        validate_renaming(
+            serializer,
+            id_field='name',
+            original_id=self.kwargs['asset_name'],
+            extra_log=self.request
         )
         return serializer.upsert(
             lookup, item=item, file=get_asset_path(item, self.kwargs['asset_name'])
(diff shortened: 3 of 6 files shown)


app/stac_api/admin.py
```diff
@@ -237,3 +237,11 @@ def get_form(self, request, obj=None, **kwargs):  # pylint: disable=arguments-di
     def href(self, instance):
         path = instance.file.name
         return build_asset_href(self.request, path)
+
+    # We don't want to move the assets on S3
+    # That's why some fields like the name of the asset are set readonly here
+    # for update operations
+    def change_view(self, request, object_id, form_url='', extra_context=None):
+        self.readonly_fields = self.get_readonly_fields(request)
+        self.readonly_fields.extend(['name', 'item', 'file'])
```
ltshb, line 246: The `file` should not be read only, we should be able to change the file content from the admin page. This does actually a new upload and is needed if we want to update an asset file from admin page.

app/tests/test_admin_page.py
```diff
@@ -606,60 +606,6 @@ def setUp(self):
...
-        self.assertEqual(response.status_code, 302, msg="Admin page failed to update asset")
-        asset.refresh_from_db()
-        for key, value in data.items():
-            if key in ['item', 'id', 'file']:
-                continue
-            self.assertEqual(getattr(asset, key), value, msg=f"Failed to update field {key}")
-
-        # Check that file content has changed
-        with asset.file.open() as fd:
-            self.assertEqual(filecontent, fd.read())
```
ltshb, line 640: This test should not be removed it is important we still need to be able to update an asset from admin page. Renaming is not allowed but updating metadata and asset file must be allowed

app/tests/test_admin_page.py
```diff
@@ -606,60 +606,6 @@ def setUp(self):
...
-        response = self.client.post(reverse('admin:stac_api_asset_change', args=[asset.id]), data)
-        self.assertEqual(response.status_code, 302)
-
-        # Un-comment with BGDIINF_SB-1625
-        # Assert that the location on s3 has been changed
-        # new_path = f"{asset.item.collection.name}/{asset.item.name}/{data['name']}"
-        # self.assertS3ObjectExists(new_path)
-
-        # asset.refresh_from_db()
-        # self.assertEqual(asset.file.name, new_path)
```
ltshb, line 661: This test should also be left but changed to make sure that renaming is not allowed. If I remember well in case of failure the status code is 200 and 302 in case of success for admin page request.

app/stac_api/views.py
```diff
@@ -481,6 +483,24 @@ def get_serializer(self, *args, **kwargs):
         kwargs.setdefault('context', self.get_serializer_context())
         return serializer_class(*args, **kwargs)
 
+    def perform_upsert(self, serializer, lookup):
+        if serializer.validated_data['name'] != self.kwargs['asset_name']:
+            raise ValidationError(
+                _('This query leads to a rename of the asset'
+                  ' and this is forbidden (in case you want to rename,'
+                  ' you have to delete the asset first).')
```
boecklic, line 491: actually with the presigned url it will be 'creating new' and then 'deleting old'

app/tests/test_assets_endpoint.py
```diff
@@ -554,25 +555,31 @@ def test_asset_endpoint_put_rename_asset(self):
...
         )
-        self.assertStatusCode(200, response)
-        json_data = response.json()
-        self.assertEqual(changed_asset.json['id'], json_data['id'])
-        self.check_stac_asset(
-            changed_asset.json, json_data, collection_name, item_name, ignore=['item']
-        )
+        self.assertStatusCode(400, response)
+
+        self.assertStatusCode(400, response)
```
ltshb, line 560: Duplicates lines. It would good also to check after the status code that the description is the expected one (e.g. `Renamed not allowed`) to make sure that we got a 400 for the good reason (there is so many reasons for 400).

app/tests/test_assets_endpoint.py
```diff
@@ -585,25 +592,27 @@ def test_asset_endpoint_patch_rename_asset(self):
         response = self.client.patch(
             path, data=changed_asset.get_json('patch'), content_type="application/json"
         )
-        json_data = response.json()
-        self.assertStatusCode(200, response)
-        self.assertEqual(changed_asset.json['id'], json_data['id'])
-        self.check_stac_asset(
-            changed_asset.json, json_data, collection_name, item_name, ignore=['item']
-        )
+        self.assertStatusCode(400, response)
```
ltshb, line 595: Same as above also test the response description to make sure that the 400 is due to a  `rename not allowed`

ltshb (changes requested):
Also remove the `move_asset()` and `remove_asset()` deprecated and unused method in model.

app/stac_api/views.py
ltshb: This code could be actually be moved inside a generic function for all renaming:
```python
# validators_view.py
def validate_renaming(serializer, original_name, log_extra):
  if serializer.validated_data['name'] !=original_name:
    logger.error(..., extra=log_extra)
     raise ValidationError(_('Renaming object is not allowed'), code='invalid')
```
This should also then be used fro all renaming like for the collection (@IsabelleBzr)

@rebert  note also that in your implementation logging has been forgotten.

app/stac_api/validators_view.py
```diff
@@ -92,3 +92,11 @@ def validate_upload_parts(request):
         message = f'Required "parts" must be a list, not a {type(request.data["parts"])}'
         logger.error(message, extra={'request': request})
         raise ValidationError({'parts': _(message)}, code='invalid')
+
+
+def validate_renaming(serializer, id_field, original_name):
+    data = serializer.validated_data
+    if data[id_field] != original_name:
+        message = 'Renaming object is not allowed'
+        logger.error(message, extra={'request': data})
```
ltshb, line 101: `extra={'request': ...}` must be a django HttpRequest object, here data is a dictionary of the deserialized payload. In the extra it would be good to have the object name for this I suggest to pass the extra as argument and depending on which view we validate set different extra. E.g.
```python
def validate_renaming(serializer, id_field, original_name, log_extra):
   ... 
  logger.error(message, extra=log_extra)

class AssetsDetail:

def perform_upsert():
   validate_renaming(serializer, 'name', self.kwargs['asset_name'], {'asset': self.kwargs['asset_name']}
```

app/tests/test_admin_page.py
ltshb: This still needs to be re-added and modified

app/tests/test_admin_page.py
ltshb: This still needs to be re-added and modified

boecklic (approved):
:+1:

app/tests/test_admin_page.py
ltshb: @rebert I guess you didn't find a way to easily test if the name field in admin page is read only ?

app/stac_api/admin.py
```diff
@@ -237,3 +238,12 @@ def get_form(self, request, obj=None, **kwargs):  # pylint: disable=arguments-di
...
         return build_asset_href(self.request, path)
+
+    # We don't want to move the assets on S3
+    # That's why some fields like the name of the asset are set readonly here
+    # for update operations
+    def change_view(self, request, object_id, form_url='', extra_context=None):
+        self.readonly_fields = self.get_readonly_fields(request)
+        self.readonly_fields.extend(['name', 'item'])
+
+        return super().change_view(request, object_id, form_url, extra_context)
```
hansmannj, line 249: Nice :smile:

hansmannj (approved):
nice solution :+1:

ltshb (approved):

IsabelleBzr (approved):

PR #267, opened 2021-04-08
feature_BGDIINF_SB-1625_remove_checksum

ltshb:
With the new asset management the checksum:multihash is set when
creating a new asset upload and not anymore when creating/updating
the asset metadata.

Also some minor clean up and improvements:
* Updated the README.md table of content.
* Removed unused variable in Makefile
* Allowed module name with numbers in pylintrc: e.g. s3_multipart_upload
before only migrations module where allowed to have number in there
names. However a valid name must always start with a letter (except for
migrations).
* Also improved a unittest error message
* Fixed pagination bug in case the links would only contain the
paginations links. In this case `links` would have been a list of list
of link instead of a list of link. This use case what currently not used
but will be with the asset uploads view.


Makefile
```diff
@@ -5,7 +5,6 @@ SHELL = /bin/bash
 SERVICE_NAME := service-stac
 
 CURRENT_DIR := $(shell pwd)
-VENV := $(CURRENT_DIR)/.venv
```
rebert, line 8: thanks

rebert (approved):
thanks!

hansmannj (approved):
LGTM :+1:

PR #268, opened 2021-04-08
BGDIINF_SB-1739: Presigned url

ltshb:
Implemented Asset upload endpoint for presigned url.


app/stac_api/s3_multipart_upload.py
```diff
@@ -0,0 +1,266 @@
...
+            },
+        )
+        raise ValueError(response)
+
+    def abort_multipart_upload(self, key, asset, upload_id):
+        '''Abort a multipart upload on the backend
+
+        Args:
+            key: string
+                key on the S3 backend for which we want to complete the multipart upload
```
hansmannj, line 223: `abort` instead of `complete`

app/stac_api/urls.py
```diff
@@ -17,8 +21,17 @@
 STAC_VERSION_SHORT = settings.STAC_VERSION_SHORT
 HEALTHCHECK_ENDPOINT = settings.HEALTHCHECK_ENDPOINT
 
+asset_upload_urls = [
+    path("<upload_id>", AssetUploadDetail.as_view(), name='asset-upload-detail'),
+    # path("<upload_id>/parts/<part_number>", AssetUploadPart.as_view(), name='asset-upload-part'),
```
hansmannj, line 26: Intentionally commented out?

app/stac_api/views.py
```diff
@@ -274,6 +300,11 @@ def put(self, request, *args, **kwargs):
     def patch(self, request, *args, **kwargs):
         return self.partial_update(request, *args, **kwargs)
 
+    # Here the etag is only added to support pre-conditional If-Match and If-Not-Match
+    # @etag(get_collection_etag)
+    # def delete(self, request, *args, **kwargs):
+    #     return self.destroy(request, *args, **kwargs)
+
```
hansmannj, line 307: Intentionally commented out?

app/stac_api/urls.py
ltshb: Yes it will be implemented later

app/stac_api/views.py
ltshb: Yes the implementation of the collection delete is not complete and there is a ticket to do that, see https://jira.swisstopo.ch/browse/BGDIINF_SB-1742. I've added it here to be able to do some E2E test allowing me to clean up the DB after the E2E test.

app/stac_api/serializers.py
```diff
@@ -933,7 +933,7 @@ class UploadPartSerializer(serializers.Serializer):
     # pylint: disable=abstract-method
     etag = serializers.CharField(source='ETag', allow_blank=False, required=True)
     part_number = serializers.IntegerField(
-        source='PartNumber', min_value=1, max_value=10000, required=True, allow_null=False
+        source='PartNumber', min_value=1, max_value=100, required=True, allow_null=False
```
hansmannj, line 936: Does the max file size limit also need to be checked?

app/stac_api/serializers.py
```diff
@@ -913,3 +915,79 @@ def validate(self, attrs):
...
+    def data(self):
+        ret = super(serializers.ListSerializer, self).data
+        return ReturnDict(ret, serializer=self)
+
+
+class UploadPartSerializer(serializers.Serializer):
+    # pylint: disable=abstract-method
+    etag = serializers.CharField(source='ETag', allow_blank=False, required=True)
+    part_number = serializers.IntegerField(
+        source='PartNumber', min_value=1, max_value=100, required=True, allow_null=False
```
hansmannj, line 936: Would it make sense to use a global setting or env var for this, as it is also used in `models.py` to ensure consistence?

app/stac_api/signals.py
```diff
@@ -1,13 +1,34 @@
...
 
 from stac_api.models import Asset
+from stac_api.models import AssetUpload
 
 logger = logging.getLogger(__name__)
 
 
+@receiver(pre_delete, sender=AssetUpload)
+def check_on_going_upload(sender, instance, **kwargs):
+    if instance.status == AssetUpload.Status.IN_PROGRESS:
```
hansmannj, line 15: Would it make sense to force-abort the ongoing upload and perform the deletion afterwards?

ltshb: I would not because it is usually not a use case and it increase complexity.

app/stac_api/serializers.py
ltshb: Well this value should not changed because of the lack of pagination for the urls that depend on this number. And also if we would do it here, then we should do for all validator parameters in models and serializer. I think we should know that model and serializer must remain in sync and we don't need to define variables for this for simplicity.

hansmannj: Makes sense, I guess you mean my comment on the gobal setting or env var for this.
What about checking the max allowed file size? Does that need to be validated for each part?

ltshb: The actual upload is not done via our api, but directly with the S3 API through the presigned URL. So we have no control about the file size at all. However if the file size are not in the allowed range; 5MB - 5GB for each parts except the last one which don't have a minimum, S3 will complain with a 400 and a proper message. I already added comment about file size in the spec in the next coming PR.

hansmannj: Another stupid question: If I understood correctly, the value for `max_value` here depends on our `PAGE_SIZE` settings, right? Then we could probably simply use `os.environ.get('PAGE_SIZE', 100)` here to be consistent.

app/stac_api/models.py
```diff
@@ -6,13 +6,18 @@
 
 # import botocore.exceptions # Un-comment with BGDIINF_SB-1625
 import multihash
+from multihash import to_hex_string
```
boecklic, line 9: any particular reason for importing multihash twice in different ways?

app/stac_api/models.py
```diff
@@ -736,7 +742,93 @@ def delete(self, *args, **kwargs):  # pylint: disable=signature-differs
...
+            raise ValidationError(error.args[0]) from None
 
     def clean(self):
         validate_asset_name_with_media_type(self.name, self.media_type)
+
+
+class AssetUpload(models.Model):
+
+    class Meta:
+        constraints = [
```
boecklic, line 767: Maybe a short description/explanation about these constraints could help a future dev

app/stac_api/utils.py
```diff
@@ -107,6 +107,17 @@ def get_s3_resource():
...
 
 
+def get_s3_client():
+    '''Returns an AWS S3 client
+
+    Returns:
+        AWS S3 client
+    '''
+    return boto3.client(
+        's3', endpoint_url=settings.AWS_S3_ENDPOINT_URL, config=Config(signature_version='s3v4')
```
boecklic, line 117: maybe add a comment how authentication works (env variables)

boecklic (approved):
Very nice! Added a few comments, mostly regarding comments

app/stac_api/serializers.py
ltshb: There is no dependency with the page size. The size of `urls` array depends on the `number_parts` and we don't have any pagination for the array `urls`. That's why I limited the `number_parts` to 100 has it would allow to upload a file up to 500GB (a part cannot be more than 5GB) and the answer of the create endpoint would still have an acceptable size. Originaly I set the `number_parts` to the S3 limit which was 10'000 but this would have cause trouble because we would then loop 10'000 times to create presigned url and the create multipart upload endpoint answer would have had a 10'000 size `urls` array.

hansmannj: OK, so i misinterpreted "pagination", sorry ;-)

app/stac_api/models.py
ltshb: Historical, but I'll change it as follow:
```python
from multihash import encode as multihash_encode
from multihash import to_hex_string
```
Which makes more sense

hansmannj (approved):
Yes, very nice and a nice test coverage, too :smile:

boecklic (approved):

hansmannj (approved):
LGTM :1st_place_medal:

rebert (approved):
Tested via admin GUI - LGTM

PR #269, opened 2021-04-12
BGDIINF_SB-1739: Added asset upload list parts endpoint

ltshb:
Aslo splitted the serializers module into two as it became quite big.

### DIFF
--- a/app/stac_api/pagination.py
+++ b/app/stac_api/pagination.py
@@ -9,6 +9,8 @@ from rest_framework import pagination
 from rest_framework.exceptions import ValidationError
 from rest_framework.request import Request
 from rest_framework.response import Response
+from rest_framework.utils.urls import remove_query_param
+from rest_framework.utils.urls import replace_query_param
 
 from stac_api.utils import get_query_params
 from stac_api.utils import remove_query_params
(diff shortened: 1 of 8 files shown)


app/stac_api/pagination.py
```diff
@@ -9,103 +9,185 @@
...
+        logger.error(
+            'Invalid query parameter limit=%d: negative number not allowed',
+            page_size,
+            extra=log_extra
+        )
+        raise ValidationError(
+            _('limit query parameter too small, must be in range 1..%d') % (max_page_size),
+            code='limit'
+        )
+    if max_page_size and page_size > max_page_size:
```
hansmannj, line 93: wouldn't simply `if page_size > max_page-size` do here?

app/stac_api/s3_multipart_upload.py
```diff
@@ -43,28 +43,19 @@ def create_multipart_upload(self, key, asset, checksum_multihash):
...
-                error,
-                extra={
-                    'collection': asset.item.collection.name,
-                    'item': asset.item.name,
-                    'asset': asset.name,
-                    's3_error': error.response
-                }
-            )
-            raise
+        response = self.call_s3_api(
```
hansmannj, line 46: No need for catching `ClientError` here anymore?

app/stac_api/s3_multipart_upload.py
```diff
@@ -93,31 +84,25 @@ def create_presigned_url(self, key, asset, part, upload_id):
...
-                extra={
-                    'collection': asset.item.collection.name,
-                    'item': asset.item.name,
-                    'asset': asset.name,
-                    'upload_id': upload_id,
-                    's3_error': error.response
-                }
-            )
-            raise
+        url = self.call_s3_api(
```
hansmannj, line 87: And here, too. No more exception handling (`ClientError`) needed?

app/stac_api/s3_multipart_upload.py
hansmannj: I see, sorry, its handled in `call_s3_api()` now

app/stac_api/s3_multipart_upload.py
hansmannj: I see, sorry, its handled in `call_s3_api()` now

app/stac_api/s3_multipart_upload.py
```diff
@@ -146,74 +131,32 @@ def complete_multipart_upload(self, key, asset, parts, upload_id):
         Raises:
             ValidationError: when the parts are not valid
```
hansmannj, line 132: or ValueError when multipart upload fails

app/stac_api/s3_multipart_upload.py
```diff
@@ -226,41 +169,110 @@ def abort_multipart_upload(self, key, asset, upload_id):
...
+            log_extra={
+                'collection': asset.item.collection.name,
+                'item': asset.item.name,
+                'asset': asset.name,
+                'upload_id': upload_id
+            }
+        )
+        return response, response.get('IsTruncated', False)
+
+    def call_s3_api(self, func, *args, **kwargs):
```
hansmannj, line 219: Nice :smile:

hansmannj (approved):
:+1:
LGTM, thanks :1st_place_medal: 
Just two small questions

app/stac_api/pagination.py
ltshb: `max_page_size` comes from `settings.REST_FRAMEWORK['PAGE_SIZE_LIMIT']` which could be None, that's why I'm checking it first.

hansmannj: OK, I see, thanks. So it could be passes as None to this function. Probably might make sense to add `| None` in the docstring for `max_page_size`?

app/stac_api/s3_multipart_upload.py
ltshb: @hansmannj well I didn't document the ValueError() because it is not clear to me when it happens. From the AWS docu it can happen but don't know in which case and/or circumstances. I couldn't force such issues from a user perspective. So I let this exception happen with a log error, and if it would happen the service would simply return an HTTP 500 with hopefully enough logs and information to understand why this happens. Only then we could want to change the logic to return eventually a 400 or any other more appropriate error code. For the moment I think it is better to have a 500 as we don't know in which circumstance it could happen.

PR #270, opened 2021-04-14
BGDIINF_SB-1739: Added management command to list all asset uploads

ltshb:
This command won't work correctly with S3 minio server as it doesn't implement the ListMultipartUploads correctly.

### DIFF
--- a/app/stac_api/management/commands/list_asset_uploads.py
+++ b/app/stac_api/management/commands/list_asset_uploads.py
@@ -43,7 +43,6 @@ class ListAssetUploadsHandler(CommandHandler):
                 limit=limit, key=s3_key_start, start=s3_upload_id_start
             )
 
-        db_uploads = []
         if not self.options['s3_only']:
             queryset = AssetUpload.objects.filter_by_status(self.options['status'])
             count = queryset.count()
@@ -55,7 +54,7 @@ class ListAssetUploadsHandler(CommandHandler):
 
         if not self.options['db_only'] and not self.options['s3_only']:
 
-            def are_uploads_equals(s3_upload, db_upload):
+            def are_uploads_equal(s3_upload, db_upload):
                 if (
                     s3_upload['UploadId'] == db_upload.upload_id and
                     s3_upload['Key'] == get_asset_path(db_upload.asset.item, db_upload.asset.name)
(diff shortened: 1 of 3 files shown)


app/stac_api/management/commands/list_asset_uploads.py
```diff
@@ -0,0 +1,166 @@
...
+            (
+                s3_uploads,
+                s3_has_next,
+                s3_next_key,
+                s3_next_upload_id,
+            ) = self.s3.list_multipart_uploads(
+                limit=limit, key=s3_key_start, start=s3_upload_id_start
+            )
+
+        db_uploads = []
```
hansmannj, line 46: `db_uploads` seems to be unused in the rest of this module

app/stac_api/management/commands/list_asset_uploads.py
```diff
@@ -0,0 +1,166 @@
...
+            count = queryset.count()
+            if count > limit:
+                queryset = queryset[start:start + limit]
+            db_uploads_qs = queryset
+            if start + limit < count:
+                db_has_next = True
+
+        if not self.options['db_only'] and not self.options['s3_only']:
+
+            def are_uploads_equals(s3_upload, db_upload):
```
hansmannj, line 58: probably `are_uploads_equal` instead of `equalS`?

app/stac_api/management/commands/list_asset_uploads.py
```diff
@@ -0,0 +1,166 @@
...
+                if (
+                    s3_upload['UploadId'] == db_upload.upload_id and
+                    s3_upload['Key'] == get_asset_path(db_upload.asset.item, db_upload.asset.name)
+                ):
+                    return True
+                return False
+
+            # Add all db uploads
+            for db_upload in db_uploads_qs:
```
hansmannj, line 67: If I understand this correctly, it depends on the same ordering of `db_uploads_qs` and `s3_uploads`, right? Is the same ordering guaranteed?

ltshb: It depends of the size, if there is less then 100 uploads (both in S3 and in DB) then the ordering is not relevant. However if we have more than 100 uploads, either in S3 or DB then not the ordering is relevant but the pagination where we should search for each page which is not done... If we have more than 100 active uploads then we would need to implement this but I don't think that it will be needed. I implement this command more as a clean up if we stil have uploads that are not synced between DB and S3. On S3 the uploads are automatically deleted once they have been completed, therefore we should never have too many in progress uploads on S3 and this command will then works.

hansmannj (approved):
:1st_place_medal:

PR #275, opened 2021-04-19
BGDIINF_SB-1742: Added collection delete endpoint

ltshb:
Collection delete is only allowed if the collection doesn't have
anymore items.


hansmannj (approved):
:+1:
Thanks!

PR #276, opened 2021-04-21
BGDIINF_SB-1764: Reorganized the openapi files

ltshb:
Also make uses of local file reference which allows to jump into
references with an IDE like VSCode.

NOTE: somehow VSCode still mark some reference as unresolved even though the reference is correct and we can jump into it with CRTL+click, I don't know why, might be a bug in vscode openapi extension.


spec/components/schemas.yaml
```diff
@@ -1,74 +1,75 @@
...
-    #              - type: number
-    #          type: array
-    #        startsWith:
-    #          description: >-
-    #            Find items with a property that begins with the specified string. A case-insensitive
-    #            comparison must be performed.
-    #          type: string
-    #      type: object
-    #  description: Apply query operations to a specific property
+    assetQuery:
```
hansmannj, line 4: Asset's POST search endpoint is not yet implemented, probably keep this commented out until it will be?

spec/components/schemas.yaml
```diff
@@ -891,10 +869,10 @@ components:
           type: string
         links:
           items:
-            $ref: "#/components/schemas/link"
+            $ref: "./schemas.yaml#/components/schemas/link"
           type: array
         stac_version:
-          $ref: "#/components/schemas/stac_version"
+          $ref: "./schemas.yaml#/components/schemas/stac_version"
```
hansmannj, line 875: Two lines below (line 877) from here and also elsewhere in the spec-code there's still some (copy&paste I guess) references to a sample address in Bonn. Probably we should replace them everywhere in the spec-code with "Musterstrasse 1 in Berne" or similar. :wink:

hansmannj (approved):
LGTM, reorganizing and being able to follow references in vscode makes life easier, good idea!
:+1:

spec/components/schemas.yaml
ltshb: We don't gain anything by commenting this out, code is still here. However the loose the folding functionality in IDE like vscode when the code is commented. Because this schema is nowhere references it will not be showed anywhere, therefore I prefer keeping it un-commented to have the possibility to fold all schemas with an IDE which make the navigation in code easier.

spec/components/schemas.yaml
ltshb: I don't mind having an example in Deutschland and I think this is not related to this PR which is more of reorganizing files in order to make use of IDE code navigation. But feel free to change it in the completion of the validation of the openapi schema.

PR #278, opened 2021-04-23
BGDIINF_SB-1753 update collections upstream

hansmannj:
* collection's `updated` property is now updated, when a child item/asset of the collection is inserted, updated or deleted.
* corresponding unit tests were added to `test_collection_model.py`

### DIFF
--- a/app/stac_api/collection_temporal_extent.py
+++ b/app/stac_api/collection_temporal_extent.py
@@ -55,7 +55,9 @@ class CollectionTemporalExtentMixin():
             logger.critical(
                 'Failed to update collection temporal extent; invalid trigger parameter %s',
                 trigger,
-                extra={'collection', self.name, 'item', item.name}
+                extra={
+                    'collection': self.name, 'item': item.name
+                }
             )
             raise ValueError(f'Invalid trigger parameter; {trigger}')
 
--- a/app/tests/test_collection_model.py
+++ b/app/tests/test_collection_model.py
@@ -1,11 +1,13 @@
 import logging
+from time import sleep
 
 from django.core.exceptions import ValidationError
 from django.test import TestCase
 
 from stac_api.models import Collection
 
-from tests.data_factory import CollectionFactory
+from tests.data_factory import Factory
+from tests.utils import mock_s3_asset_file
 
 logger = logging.getLogger(__name__)
 
@@ -13,8 +15,10 @@ logger = logging.getLogger(__name__)
 class CollectionsModelTestCase(TestCase):
 
     @classmethod
+    @mock_s3_asset_file
     def setUpTestData(cls):
-        cls.collection = CollectionFactory().create_sample(db_create=True)
+        cls.factory = Factory()
+        cls.collection = Factory().create_collection_sample(db_create=True)
 
     def test_create_already_existing_collection(self):
         # try to create already existing collection twice
@@ -26,14 +30,14 @@ class CollectionsModelTestCase(TestCase):
     def test_create_collection_invalid_name(self):
         # try to create a collection with invalid collection name
         with self.assertRaises(ValidationError, msg="Collection with invalid name was accepted."):
-            CollectionFactory().create_sample(sample="collection-invalid", db_create=True)
+            Factory().create_collection_sample(sample="collection-invalid", db_create=True)
 
     def test_create_collection_missing_mandatory_fields(self):
         # try to create a collection with invalid collection name
         with self.assertRaises(
             ValidationError, msg="Collection with missing mandatory fields was accepted."
         ):
-            CollectionFactory().create_sample(
+            Factory().create_collection_sample(
                 name="collection-missing-mandatory-fields",
                 sample="collection-missing-mandatory-fields",
                 db_create=True
@@ -42,7 +46,7 @@ class CollectionsModelTestCase(TestCase):
     def test_create_collection_invalid_links(self):
         # try to create a collection with invalid collection name
         with self.assertRaises(ValidationError, msg="Collection with invalid links was accepted."):
-            CollectionFactory().create_sample(
+            Factory().create_collection_sample(
                 name="collection-invalid-links", sample="collection-invalid-links", db_create=True
             )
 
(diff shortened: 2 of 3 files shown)


IsabelleBzr (approved):
:+1:

app/tests/test_collection_model.py
```diff
@@ -1,20 +1,24 @@
...
 
 
 class CollectionsModelTestCase(TestCase):
 
     @classmethod
+    @mock_s3_asset_file
     def setUpTestData(cls):
-        cls.collection = CollectionFactory().create_sample(db_create=True)
+        cls.factory = Factory()
+        cls.collection = Factory().create_collection_sample(db_create=True)
```
ltshb, line 21: Use `cls.factory` instead of `Factory()`

app/tests/test_collection_model.py
```diff
@@ -26,14 +30,14 @@ def test_create_already_existing_collection(self):
     def test_create_collection_invalid_name(self):
         # try to create a collection with invalid collection name
         with self.assertRaises(ValidationError, msg="Collection with invalid name was accepted."):
-            CollectionFactory().create_sample(sample="collection-invalid", db_create=True)
+            Factory().create_collection_sample(sample="collection-invalid", db_create=True)
```
ltshb, line 33: Use the `self.factory` instead of a new `Factory()` instance.

app/tests/test_collection_model.py
```diff
@@ -51,20 +55,105 @@ def test_create_collection_invalid_providers(self):
...
+    def test_collection_update_on_item_write_operations(self):
+        # assert that collection's updated property is updated when an item is
+        # added to the collection, this item is updated and this item is deleted
+
+        # check collection's update on item insertion
+        initial_last_modified = self.collection.model.updated
+        sleep(0.01)
+        item = Factory().create_item_sample(self.collection.model, sample='item-1', db_create=True)
+        item.model.full_clean()
+        item.model.save()
```
ltshb, line 87: This should not be needed as the model is already created and saved!

app/tests/test_collection_model.py
```diff
@@ -51,20 +55,105 @@ def test_create_collection_invalid_providers(self):
...
+        self.assertGreater(
+            self.collection.model.updated,
+            initial_last_modified,
+            msg="Collection's updated property was not updated on item insert"
+        )
+
+        # check collection's update on item update
+        initial_last_modified = self.collection.model.updated
+        sleep(0.01)
+        item.model.name = "new_randon_name"
```
ltshb, line 98: Renaming item is not allowed ! Currently this check is only done at view level but if we change the validation in a lower level then this will failed. Update another attribute that is allowed instead of the name, e.g. the `properties_title`

app/tests/test_collection_model.py
```diff
@@ -51,20 +55,105 @@ def test_create_collection_invalid_providers(self):
...
         )
+
+    def test_collection_update_on_item_write_operations(self):
+        # assert that collection's updated property is updated when an item is
+        # added to the collection, this item is updated and this item is deleted
+
+        # check collection's update on item insertion
+        initial_last_modified = self.collection.model.updated
+        sleep(0.01)
+        item = Factory().create_item_sample(self.collection.model, sample='item-1', db_create=True)
```
ltshb, line 85: Use `self.factory`

app/tests/test_collection_model.py
```diff
@@ -51,20 +55,105 @@ def test_create_collection_invalid_providers(self):
...
+            initial_last_modified,
+            msg="Collection's updated property was not updated on item deletion"
+        )
+
+    def test_collection_update_on_asset_write_operations(self):
+        # assert that collection's updated property is updated when an asset is
+        # added to an item of the collection, this asset is updated and this asset is deleted
+
+        # check collection's update on asset insertion
+        item = Factory().create_item_sample(self.collection.model, sample='item-1', db_create=True)
```
ltshb, line 124: Use `self.factory`

app/tests/test_collection_model.py
```diff
@@ -51,20 +55,105 @@ def test_create_collection_invalid_providers(self):
...
+        initial_last_modified = self.collection.model.updated
+        sleep(0.01)
+        asset = Factory().create_asset_sample(item=item.model, sample='asset-1', db_create=True)
```
ltshb, line 127: Use `self.factory`

app/tests/test_collection_model.py
```diff
@@ -51,20 +55,105 @@ def test_create_collection_invalid_providers(self):
...
+        asset.model.full_clean()
+        asset.model.save()
```
ltshb, line 129: These should not be needed, already part of `create_asset_sample`

app/tests/test_collection_model.py
```diff
@@ -51,20 +55,105 @@ def test_create_collection_invalid_providers(self):
...
+        self.assertGreater(
+            self.collection.model.updated,
+            initial_last_modified,
+            msg="Collection's updated property was not updated on asset insert"
+        )
+
+        # check collection's update on asset update
+        initial_last_modified = self.collection.model.updated
+        sleep(0.01)
+        asset.model.name = f"new-{asset.model.name}"
```
ltshb, line 140: Renaming asset is not allowed (see comment about item renaming). Update another attribute instead like the `title`

ltshb (commented):
Good just some clean ups.

app/tests/test_collection_model.py
hansmannj: Sure, should have remembered that :see_no_evil: 
Guess the test passes, as the "renaming not allowed" issue is not  handled on model level, so it is not an error that the test passes anyways, right?

ltshb: Yes that's correct test pass because the validation is not on model level but on view level and on admin page the renaming is avoided due to the usage of read only field.

app/stac_api/collection_temporal_extent.py
```diff
@@ -55,7 +55,9 @@ def update_temporal_extent(self, item, trigger, original_item_values):
             logger.critical(
                 'Failed to update collection temporal extent; invalid trigger parameter %s',
                 trigger,
-                extra={'collection', self.name, 'item', item.name}
+                extra={
+                    'collection': self.name, 'item': item.name
+                }
```
ltshb, line 60: Should be on one line otherwise yapf will change it on next format

ltshb (approved):
:+1:

PR #279, opened 2021-04-27
BGDIINF_SB-1764 GeoJSON

hansmannj:
* copied JSON schema from https://geojson.org/schema/Geometry.json in our spec and adapted it to be compliant with our OpenAPI yaml.
* made all geometry types (i.e. GeoJSON Point, GeoJSON LineString, GeoJSON Polygon, GeoJSON MultiPoint, GeoJSON MultiLineString or GeoJSON MultiPolygon) valid options for all methods on item endpoint (GET, POST, PATCH and PUT). Currently only for PATCH all geometry options are available in the productive specs. For all other methods only `polygonGeoJSON` is a valid option in the specs.


spec/components/schemas.yaml
```diff
@@ -575,15 +575,15 @@ components:
...
-        # - $ref: "./schemas.yaml#/components/schemas/geometrycollectionGeoJSON"
+    # geometryGeoJSON:
+    #   oneOf:
+    #     # - $ref: "./schemas.yaml#/components/schemas/pointGeoJSON"
+    #     # - $ref: "./schemas.yaml#/components/schemas/multipointGeoJSON"
+    #     # - $ref: "./schemas.yaml#/components/schemas/linestringGeoJSON"
+    #     # - $ref: "./schemas.yaml#/components/schemas/multilinestringGeoJSON"
+    #     - $ref: "./schemas.yaml#/components/schemas/polygonGeoJSON"
+    #     # - $ref: "./schemas.yaml#/components/schemas/multipolygonGeoJSON"
+    #     # - $ref: "./schemas.yaml#/components/schemas/geometrycollectionGeoJSON"
```
hansmannj, line 586: I was not exactly sure, why there was only polygonGeoJSON available here and the rest was commented out.
In the currently productive specs for item's PATCH all geometry types are described. For all other methods, only `polygonGeoJSON` is available.
I supposed that all options should be valid for any method and implemented it that way in this PR.
So I'll delete this commented out section, once this PR is approved.

spec/components/schemas.yaml
```diff
@@ -644,11 +644,12 @@ components:
         bbox:
           $ref: "./schemas.yaml#/components/schemas/bbox"
         geometry:
+          $ref: "./schemas.yaml#/components/schemas/itemGeometry"
           # we could use the 'original' schema, but it doesn't
           # contain a useful example, hence we use our own
           # schema with appropriate example
           # $ref: https://geojson.org/schema/Geometry.json
-          $ref: "./schemas.yaml#/components/schemas/geometryGeoJSON"
+          # $ref: "./schemas.yaml#/components/schemas/geometryGeoJSON"
```
hansmannj, line 652: This section will be deleted, if this PR is approved, i.e. if you agree, that all geometry types should be valid options in the specs for all methods on item endpoints.

spec/components/schemas.yaml
```diff
@@ -818,6 +818,150 @@ components:
...
+            coordinates:
+              type: array
+              minItems: 2
+              items:
+                type: number
+            bbox:
+              type: array
+              minItems: 4
+              items:
+                type: number
```
ltshb, line 843: I would suggest to break down the schema for each object type in the `onOf`:

```yaml
components:
  schemas:
    itemGeometry:
      onOf:
        - $ref: "#/components/schemas/geoJsonPoint"
       ...
    geoJsonPoint:
      title: GeoJSON Point
       type: object
       required:
          - type
          - coordinates
       properties:
         type:
           type: string
           enum:
             - Point
         coordinates:
           type: array
           minItems: 2
           items:
             type: number
         bbox:
           type: array
           minItems: 4
           items:
             type: number
```
Then we can keep all GeoJSON schemas and only remove refs in `itemGeometry` if needed. This can be useful for future uses instead of removing all parts from the itemGeometry or commenting out.

spec/components/schemas.yaml
```diff
@@ -818,6 +818,150 @@ components:
...
+              items:
+                type: array
+                minItems: 2
+                items:
+                  type: number
+            bbox:
+              type: array
+              minItems: 4
+              items:
+                type: number
```
ltshb, line 891: There is already a schema for the `bbox` that should be reused here and in any GeoJSON types. Its description is not 100% correct as it is currently used in the item response schemas and in item query parameter. For the response the description sentence `Only features that have a geometry that intersects the bounding box are selected.` doesn't make sense and should be removed. However this sentence make sense for the query parameter. But for this one we can use the `bboxfilter` shemas instead.

spec/components/schemas.yaml
```diff
@@ -818,6 +818,150 @@ components:
...
+              enum:
+                - LineString
+            coordinates:
+              type: array
+              minItems: 2
+              items:
+                type: array
+                minItems: 2
+                items:
+                  type: number
```
ltshb, line 861: Might be a good idea to add a short description for the coordinates, for example also add the type of coordinate we support like in the bbox description.

ltshb (commented):
About the type of GeoJson that we support I don't know, @rebert and/or @boecklic should better know than me on this :-)

spec/components/schemas.yaml
rebert: why not

rebert:
To filter the GeoJSON we use the library GEOS and following geometry types are supported:
https://docs.djangoproject.com/en/3.2/ref/contrib/gis/geos/#django.contrib.gis.geos.GEOSGeometry.geom_typeid

GeoJSON Point, GeoJSON LineString, GeoJSON Polygon, GeoJSON MultiPoint, GeoJSON MultiLineString or GeoJSON MultiPolygon are supported

boecklic:
@hansmannj @rebert since we support only polygon type geometries in the db I believe we should only accept polygon type geometries in the payload and only mention this type in the spec.

rebert:
Sorry. I had the search endpoint in mind

hansmannj:
> @hansmannj @rebert since we support only polygon type geometries in the db I believe we should only accept polygon type geometries in the payload and only mention this type in the spec.

@boecklic and @rebert 
So I'll leave the geometry for the search endpoint's intersect as is and only change the geometry type for all other methods to the polygon types (i.e. only this two: `GeoJSONPolygon` and `GeoJSONMultiPolygon`), right?

rebert:
@hansmannj @boecklic For all other methods just GeoJSONPolygon:
https://github.com/geoadmin/service-stac/blob/develop/app/stac_api/models.py#L381

hansmannj:
@rebert @ltshb @boecklic
The `bbox` is an optional property of the geometry object (see: https://tools.ietf.org/html/rfc7946#section-5).
I'd suppose we do not add it in the specs, as we already have a `bbox` property on item level. Otherwise there would be two bboxes, one on the item level and one on the item's geometry level, which could be confusing I guess.
What's your take on that?

spec/components/schemas.yaml
```diff
@@ -644,11 +626,11 @@ components:
...
           $ref: "./schemas.yaml#/components/schemas/bbox"
         geometry:
+          $ref: "./schemas.yaml#/components/schemas/itemGeometry"
           # we could use the 'original' schema, but it doesn't
           # contain a useful example, hence we use our own
-          # schema with appropriate example
-          # $ref: https://geojson.org/schema/Geometry.json
-          $ref: "./schemas.yaml#/components/schemas/geometryGeoJSON"
+          # schema with appropriate example. Original schemas
+          # were defined here: https://geojson.org/schema/Geometry.json
```
ltshb, line 633: Comment should be before the reference or totally removed, actually for such case the comment in git message is more appropriate and would always be up to date.

spec/components/schemas.yaml
```diff
@@ -818,6 +800,209 @@ components:
...
       example: smr200-200-4-2019
       type: string
+    itemGeometry:
+      # oneOf:
+        # - $ref: "#/components/schemas/geoJsonPoint"
+        # - $ref: "#/components/schemas/geoJsonLineString"
+      $ref: "#/components/schemas/geoJsonPolygon"
+        # - $ref: "#/components/schemas/geoJsonMultiPoint"
+        # - $ref: "#/components/schemas/geoJsonMultiLineString"
+        # - $ref: "#/components/schemas/geoJsonMultiPolygon"
```
ltshb, line 810: I would either remove all comments of at least leave it with the oneOf:
```yaml
oneOf:
  - $ref: "#/components/schemas/geoJsonPolygon"
  #- $ref: "#/components/schemas/geoJsonPoint"
 ...
```

spec/components/schemas.yaml
```diff
@@ -644,11 +626,11 @@ components:
...
           # we could use the 'original' schema, but it doesn't
           # contain a useful example, hence we use our own
-          # schema with appropriate example
-          # $ref: https://geojson.org/schema/Geometry.json
-          $ref: "./schemas.yaml#/components/schemas/geometryGeoJSON"
+          # schema with appropriate example. Original schemas
+          # were defined here: https://geojson.org/schema/Geometry.json
         properties:
           $ref: "./schemas.yaml#/components/schemas/itemProperties"
           readOnly: true
```
ltshb, line 636: Not related to this PR but could be corrected here. The reference are local in the file therefore the prefix `./schemas.yaml` could be removed here

ltshb (approved):

spec/components/schemas.yaml
hansmannj: If `oneOf` was not commented out here, then we'd have a text in the spec like `one of` + only one choice, which I found did not make too much sense. That's why I commented `oneOf` out until in future probably at least one of the other choices becomes available, too. I don't mind removing all commented out lines here, just left them for convenience, in case they are needed in future.

spec/components/schemas.yaml
hansmannj: I moved it before the reference and shortened it

spec/components/schemas.yaml
hansmannj: removed those prefixes in the whole `schemas.yaml`

hansmannj:
Open point left: the item's `geometry` property has itself an optional `bbox` property, but the item already has its own `bbox` in our implementation. I suggest to drop the item's geometry bbox and only have the item's bbox. Do you agree @boecklic @ltshb @rebert ?
Otherwise we'd have something like:
```
item
     bbox
     geometry
          bbox
```
which could be a bit confusing.

PR #281, opened 2021-05-05
BGDIINF_SB-1810: Added admin page for asset upload

ltshb:
This gives the possibility to list all asset uploads.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@
 Prerequisites on host for development and build:
 
 - python version 3.7
+- libgdal-dev
 - [pipenv](https://pipenv-fork.readthedocs.io/en/latest/install.html)
 - `docker` and `docker-compose`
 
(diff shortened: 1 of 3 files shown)


rebert:
image
@ltshb testing the functionality manually - it look like this for me. What do I miss?

ltshb:
> image
> @ltshb testing the functionality manually - it look like this for me. What do I miss?

Nothing if you don't have any asset upload locally. Asset uploads needs to be created with the API in order to download asset files from the API. There is currently a version on DEV for testing. However after some thought I'll improve it a bit by making all fields read only. We should not use the Admin page to create/update asset upload but only do them from the API. But the admin view would help to debug in case of problems where we could easily see what is on the DB. I'll do some modication and and deploy it again on DEV for testing.

rebert (approved):
seems to work

PR #284, opened 2021-05-19
BGDIINF_SB-1827: Readded item geometry field in admin page

ltshb:
The item geometry field was missing in the admin page. This was due to
the ItemAdmin.get_fieldsets() method which removed the geometry from the
first fieldset.

Now the geometry field has its own fieldset which is also a preparation
for further improvement in geometry field (see BGDIINF_SB-1650)

### DIFF
--- a/app/stac_api/admin.py
+++ b/app/stac_api/admin.py
@@ -125,7 +125,10 @@ class ItemAdmin(admin.GeoModelAdmin):
     readonly_fields = ['collection_name']
     fieldsets = (
         (None, {
-            'fields': ('name', 'collection', 'geometry')
+            'fields': ('name', 'collection')
+        }),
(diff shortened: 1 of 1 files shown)


rebert (commented):
There is a bug. When deleting features in the field, the background map disappears as well... Maybe this will have to be treated in a bug ticket?

ltshb:
> There is a bug. When deleting features in the field, the background map disappears as well... Maybe this will have to be treated in a bug ticket? Maybe it is just about defining the max extend around Switzerland...

@rebert I think this is not related to this PR and ticket, so I suggest to create a new ticket. The solution might take a bit more time as this small fix and we need then to plan it in a future sprint.

rebert (approved):

PR #286, opened 2021-05-20
#patch, BGDIINF_SB-1826, BGDIINF_SB-1827: Hotfix Item Admin page - v1.4.1

ltshb:
The item geometry field was missing in the admin page. This was due to
the ItemAdmin.get_fieldsets() method which removed the geometry from the
first fieldset.

Now the geometry field has its own fieldset which is also a preparation
for further improvement in geometry field (see BGDIINF_SB-1650)


hansmannj (approved):
Merci :+1:

PR #290, opened 2021-06-04
Bug_fix BGDIINF_SB-1850 large file upload via admin GUI

hansmannj:
Currently uploading assets files, which are larger than the default 2.5 MB, which is the default for `DATA_UPLOAD_MAX_MEMORY_SIZE`  fails when using the admin GUI.

The `RequestDataTooBig` exception was raised, when the middleware logging tried to access the `request.body` and hence Djangos `request.py` evaluated the `content_length` against the `DATA_UPLOAD_MAX_MEMORY_SIZE`.
This is now fixed by NOT adding the first 200 characters of the request.body in the middleware logging in case the request.path starts with /api/stac/admin (thanks @ltshb ;-)).

In addition a minor change in `test_admin_page.py`'s `_create_asset()` was done in order to have a more consistent file naming.

Once this PR is merged, I'll deploy the changes toDEV and test to upload a large asset file (probably 60 GB?). If this works, it should also work on INT and probably no need for adding an `emptyDir` in the k8s repo then.

### DIFF
--- a/app/tests/test_admin_page.py
+++ b/app/tests/test_admin_page.py
@@ -224,11 +224,11 @@ class AdminBaseTestCase(TestCase):
         start = time.time()
         filecontent = b'mybinarydata'
         filelike = BytesIO(filecontent)
-        filelike.name = 'testname.tiff'
+        filelike.name = 'testname.zip'
 
         data = {
             "item": item.id,
-            "name": "test_asset.zip",
+            "name": filelike.name,
             "description": "This is a description",
             "eo_gsd": 10,
             "geoadmin_lang": "en",
--- a/app/middleware/logging.py
+++ b/app/middleware/logging.py
@@ -23,7 +23,11 @@ class RequestResponseLoggingMiddleware:
             "request.query": request.GET.urlencode(RequestResponseLoggingMiddleware.url_safe)
         }
 
-        if request.method.upper() in ["PATCH", "POST", "PUT"]:
+        if request.method.upper() in [
+            "PATCH", "POST", "PUT"
+        ] and request.content_type == "application/json" and not request.path.startswith(
+            '/api/stac/admin'
+        ):
             extra["request.payload"] = str(request.body[:200])
 
         logger.info(


app/config/settings_prod.py
```diff
@@ -209,6 +209,9 @@
 UPLOAD_FILE_CHUNK_SIZE = 1024 * 1024  # Size in Bytes
 DEFAULT_FILE_STORAGE = 'stac_api.storages.S3Storage'
 
+FILE_UPLOAD_MAX_MEMORY_SIZE = 100 * 1024 * 1024  # size in Bytes
```
rebert, line 212: maybe comment =100 MByte

app/config/settings_prod.py
```diff
@@ -209,6 +209,9 @@
...
+DATA_UPLOAD_MAX_MEMORY_SIZE = 60 * 1024 * 1024 * 1024  # size in Bytes
```
rebert, line 213: maybe comment =60GByte

app/tests/test_admin_page.py
```diff
@@ -220,11 +221,26 @@ def _create_asset_minimal(self, item):
...
-    def _create_asset(self, item, extra=None):
+    def _create_asset(self, item, extra=None, large_file=False):
         start = time.time()
-        filecontent = b'mybinarydata'
-        filelike = BytesIO(filecontent)
-        filelike.name = 'testname.tiff'
+        if large_file:
+            logger.warning(
+                "Running test for admin page with very large asset file. "
+                "This can take a while (about 1.5 minutes), sorry."
```
rebert, line 229: as you mentioned.... maybe just skipping in the default config

rebert (approved):
with some minor comments

app/config/settings_prod.py
hansmannj: :+1:

app/config/settings_prod.py
hansmannj: :+1:

app/tests/test_admin_page.py
hansmannj: Done :-)
But I set the default to `True`, so that one has to explicitly disable it in `.env.local`, just to be on the safe side.

hansmannj:
Thanks @ltrod for increasing the resources on the instance so that the `MemoryError` now shouldn't be raised anymore.

app/tests/test_admin_page.py
```diff
@@ -220,11 +222,26 @@ def _create_asset_minimal(self, item):
...
+            )
+            asset_file = './tests/sample_data/large_asset_file.tiff'
+            if not os.path.isfile(asset_file):
+                with open(asset_file, "wb") as dummy_file:
+                    dummy_file.seek(1073741824 - 1)
+                    dummy_file.write(b'\0')
```
boecklic, line 236: maybe add a comment about how you get to this number

boecklic (commented):
Looks good. But I wonder if the same testing effect couldn't be achieved as well with a small file and those settings overwritten for that test to much smaller values? It's not really testing the X Gb, but it's testing the effect of those settings. For me this would be acceptable.

app/tests/test_admin_page.py
hansmannj: Sounds like a good idea, I'll give it a try.
If it works, it would be much better and the resources for the build instance could be reduced to "normal" again :-)

app/config/settings_dev.py
```diff
@@ -22,6 +22,10 @@
     strtobool(os.getenv('DEBUG_ENABLE_DB_EXPLAIN_ANALYZE', 'False'))
 )
 
+RUN_LARGE_ASSET_FILE_UNIT_TEST = bool(
+    strtobool(os.getenv('RUN_LARGE_ASSET_FILE_UNIT_TEST', 'True'))
+)
+
```
ltshb, line 28: I would not differentiate unit test between dev and prod, this don't make sense, unit test are for local dev, and are to avoid merging, deploying non working code. This settings should go then in settings_prod.py for simplicity.

app/tests/test_admin_page.py
```diff
@@ -220,11 +222,26 @@ def _create_asset_minimal(self, item):
...
-        filelike.name = 'testname.tiff'
+        if large_file:
+            logger.warning(
+                "Running test for admin page with very large asset file. "
+                "This can take a while (about 1.5 minutes), sorry."
+            )
+            asset_file = './tests/sample_data/large_asset_file.tiff'
+            if not os.path.isfile(asset_file):
+                with open(asset_file, "wb") as dummy_file:
+                    dummy_file.seek(1073741824 - 1)
```
ltshb, line 235: Avoid magic number or comment them

app/tests/test_admin_page.py
```diff
@@ -723,3 +740,8 @@ def test_add_update_asset_invalid_media_type(self):
             Asset.objects.filter(item=self.item, name=sample["name"]).exists(),
             msg="Asset with invalid data has been added to db"
         )
+
+    @mock_s3_asset_file
+    def test_add_very_large_asset(self):
+        if settings.RUN_LARGE_ASSET_FILE_UNIT_TEST:
+            self._create_asset(self.item, large_file=True)
```
ltshb, line 747: Uses the unittest `@unittest.skipIf(settings.RUN_LARGE_ASSET_FILE_UNIT_TEST == False)` instead here, this way you also have a proper test output. You could maybe then also change the setting like this `SKIP_LARGE_ASSET_FILE_UNITTEST` ;-)

ltshb (changes requested):

app/tests/test_admin_page.py
hansmannj: Now this whole problem is fixed in a much shorter way

app/config/settings_dev.py
hansmannj: Now this whole problem is fixed in a much shorter way

app/tests/test_admin_page.py
hansmannj: Now this whole problem is fixed in a much shorter way

app/tests/test_admin_page.py
hansmannj: Now this whole problem is fixed in a much shorter way

rebert (approved):

ltshb (approved):
I did not test, but I trust you did ;-)

PR #293, opened 2021-06-07
BGDIINF_SB-1844: #bugfix empty description field

ltkum:
Issue : when creating a provider through the admin UI and leaving the
description field empty, a description field would be present in JSON
outputs.

Fix : As the problem was that the admin interface saved an empty string
("") rather than a None object, we transform empty string descriptions
into None object before saving.


app/stac_api/models.py
```diff
@@ -242,6 +242,9 @@ def __str__(self):
     def save(self, *args, **kwargs):  # pylint: disable=signature-differs
         # It is important to use `*args, **kwargs` in signature because django might add dynamically
         # parameters
+        if self.description == "":
+            logger.debug("transforming empty description string into None for database storage")
+            self.description = None
```
ltshb, line 247: This is too hacky and would probably breaks the API, the field needs to be changed instead to have its default to `None` like for the Asset:

```
# here we need to set blank=True otherwise the field is as required in the admin interface
description = models.TextField(blank=True, null=True, default=None)
```

And the description needs to be changed in the admin interface like for `AssetAmin`
```python
def save_model(self, request, obj, form, change):
        if obj.description == '':
            # The admin interface with TextArea uses empty string instead
            # of None. We use None for empty value, None value are stripped
            # then in the output will empty string not.
            obj.description = None

        super().save_model(request, obj, form, change)
```

ltshb (changes requested):
Also add unittest for all cases

- admin no description
- admin with description
- api no description
- api with description

ltshb:
Please do not merge develop into private branch but rebase !

PR #295, opened 2021-06-09
added more details to log msg in models.py

hansmannj:
added collection, item and asset to the extra
added size of asset file to log msg

### DIFF
--- a/app/stac_api/models.py
+++ b/app/stac_api/models.py
@@ -510,7 +510,14 @@ def upload_asset_to_path_hook(instance, filename=None):
     Returns:
         Asset file path to use on S3
     '''
-    logger.debug('Start computing asset file %s multihash', filename)
+    logger.debug(
+        'Start computing asset file %s multihash (file size: %s MB)',
+        filename,
+        int(instance.file.size / 1024**2),
+        extra={
+            "collection": instance.item.collection, "item": instance.item, "asset": instance.name
+        }
+    )
     start = time.time()
     ctx = hashlib.sha256()
     for chunk in instance.file.chunks(settings.UPLOAD_FILE_CHUNK_SIZE):
--- a/app/stac_api/models.py
+++ b/app/stac_api/models.py
@@ -511,9 +511,9 @@ def upload_asset_to_path_hook(instance, filename=None):
         Asset file path to use on S3
     '''
     logger.debug(
-        'Start computing asset file %s multihash (file size: %s MB)',
+        'Start computing asset file %s multihash (file size: %.1f MB)',
         filename,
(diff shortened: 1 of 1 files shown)


app/stac_api/models.py
```diff
@@ -510,7 +510,14 @@ def upload_asset_to_path_hook(instance, filename=None):
     Returns:
         Asset file path to use on S3
     '''
-    logger.debug('Start computing asset file %s multihash', filename)
+    logger.debug(
+        'Start computing asset file %s multihash (file size: %s MB)',
+        filename,
+        int(instance.file.size / 1024**2),
+        extra={
+            "collection": instance.item.collection, "item": instance.item, "asset": instance.name
```
ltshb, line 518: `instance.item.collection` return the object not the name, same for `instance.item` ! It need to be `instance.item.collection.name` and `instance.item.name`.

ltshb (changes requested):
Also add the `extra` to the log in line 529 for consistency.

app/stac_api/models.py
hansmannj: :+1:

app/stac_api/models.py
```diff
@@ -510,7 +510,16 @@ def upload_asset_to_path_hook(instance, filename=None):
     Returns:
         Asset file path to use on S3
     '''
-    logger.debug('Start computing asset file %s multihash', filename)
+    logger.debug(
+        'Start computing asset file %s multihash (file size: %s MB)',
```
ltshb, line 514: Uses `%.1f` or `%.3f` instead of `%s`.

app/stac_api/models.py
```diff
@@ -510,7 +510,16 @@ def upload_asset_to_path_hook(instance, filename=None):
...
+        filename,
+        int(instance.file.size / 1024**2),
```
ltshb, line 516: Based on comment above remove the type casting to `int`

app/stac_api/models.py
```diff
@@ -523,7 +532,12 @@ def upload_asset_to_path_hook(instance, filename=None):
         'Set uploaded file %s multihash %s to checksum:multihash; computation done in %ss',
```
ltshb, line 532: You may also change the `%ss` to `%.3fs` here as well

ltshb (approved):
Just small improvements but otherwise is ok

app/stac_api/models.py

app/stac_api/models.py

app/stac_api/models.py
hansmannj: :+1:

PR #296, opened 2021-06-10
Removed logging configuration for deployment

ltshb:
These configuration have been moved into the k8s repository.


rebert (approved):
What shall I say :-)

PR #298, opened 2021-06-21
Bugfix : BGDIINF_SB-1844 # Bugfix empty provider descriptions

ltkum:
Issue : On admin Interface, we want empty descriptions to be treated as
Null objects, and on the API calls, we want emtpy strings in
descriptions to still remain empty strings

Fix : Added property 'empty_value' to admin inline providers fields
Added "None" as default in the provider model
Added Unit tests to check out these cases

adding migration

make format employed


app/tests/test_admin_page.py
```diff
@@ -342,6 +342,145 @@ def test_add_update_collection_with_provider(self):
...
+        data = {
+            "name": "test_collection",
+            "license": "free",
+            "description": "some very important collection",
+            "providers-TOTAL_FORMS": "0",
+            "providers-INITIAL_FORMS": "0",
+            "links-TOTAL_FORMS": "0",
+            "links-INITIAL_FORMS": "0"
+        }
+        data.update({
```
ltshb, line 358: From the admin interface, no provider description returns a `"providers-0-description": ""` in the payload.

app/tests/test_admin_page.py
```diff
@@ -342,6 +342,145 @@ def test_add_update_collection_with_provider(self):
...
+        )
+        collection = Collection.objects.get(name=data["name"])
+
+        self.assertTrue(
+            Provider.objects.filter(collection=collection, name=data["providers-0-name"]).exists(),
+            msg="Admin page Provider added not found in DB"
+        )
+        provider = Provider.objects.get(collection=collection, name=data["providers-0-name"])
+
+        provider.refresh_from_db()
```
ltshb, line 382: This `refres_from_db()` is not necessary here, we just read the value from db the line above.

app/tests/test_admin_page.py
```diff
@@ -342,6 +342,145 @@ def test_add_update_collection_with_provider(self):
...
```
ltshb, line 426: This is not needed the provider has been read just the line before.

app/tests/test_admin_page.py
```diff
@@ -342,6 +342,145 @@ def test_add_update_collection_with_provider(self):
             msg="Admin page wrong provider.roles after update"
         )
 
+    def test_add_collection_with_provider_no_description(self):
```
ltshb, line 345: Actually this use case is not needed, we cannot have `no description` from the admin interface, but simply a collection empty string.

app/tests/test_admin_page.py
```diff
@@ -342,6 +342,145 @@ def test_add_update_collection_with_provider(self):
...
+        # Login the user first
+        self.client.login(username=self.username, password=self.password)
+
+        collection, data, link, provider = self._create_collection(with_provider=True)
+
+        # update some data in provider
+        data["providers-INITIAL_FORMS"] = 1
+        data["providers-0-id"] = provider.id
+        data["providers-0-collection"] = collection.id
+        data.pop("providers-0-description")
```
ltshb, line 442: This is not a realistic use case the `"providers-0-description"` is always part of the admin interface payload.
This test could be change to test a provider having initialy a description which is removed by the admin interface using `"providers-0-description": ""`

app/tests/test_admin_page.py
```diff
@@ -342,6 +342,145 @@ def test_add_update_collection_with_provider(self):
...
+        )
+        provider = Provider.objects.get(collection=collection, name=data["providers-0-name"])
+
+        provider.refresh_from_db()
+
+        self.assertEqual(
+            provider.description, None, msg="Admin page wrong provider.description on creation"
+        )
+
+    def test_add_update_collection_remove_provider_description(self):
```
ltshb, line 432: This use case is not realistic and should be removed.

app/tests/test_admin_page.py
```diff
@@ -342,6 +342,145 @@ def test_add_update_collection_with_provider(self):
...
+        self.assertEqual(response.status_code, 302, msg="Admin page failed to update provider")
+
+        provider.refresh_from_db()
+        self.assertEqual(
+            provider.description,
+            None,
+            msg="Admin page wrong provider.description after update. Should be None"
+        )
+
+    def test_add_update_collection_empty_provider_description(self):
```
ltshb, line 458: This uses doesn't work and I couldn't figure out why. But if you add  `data["providers-0-description"] = ""` in `test_add_update_collection_with_provider` above and then check for provider.description null it works.

ltshb (changes requested):
Some issues to fixes in unittest.

app/tests/test_admin_page.py
ltkum: Which is the test that starts at line 458. But I'm removing this test case here.

app/tests/test_admin_page.py
ltkum: I don't exactly understand the problem here.

app/tests/test_admin_page.py
ltkum: All my tests works locally without any problems, so I don't know what's wrong. I'll check the aws logs for failures in the tests.

ltkum: I did not remove the test and it seems to work on AWS. perhaps you have some modifications in your environment for some reasons. I'll resolve this conversation now.

app/tests/test_admin_page.py
ltshb: Sorry was not very clear, I mean the admin interface client will always send a `providers-x-description` either with a value or with an empty string. So testing with a payload that don't contains `providers-x-description` doesn't make sense.

app/tests/test_admin_page.py
```diff
@@ -342,6 +342,74 @@ def test_add_update_collection_with_provider(self):
...
+            "links-INITIAL_FORMS": "0"
+        }
+        data.update({
+            "providers-TOTAL_FORMS": "1",
+            "providers-INITIAL_FORMS": "0",
+            "providers-0-name": "my-provider",
+            "providers-0-description": "",
+            "providers-0-roles": "licensor",
+            "providers-0-url": "http://www.example.com",
+        })
```
ltshb, line 365: What is the point of the `update` method here ? simply put alltogether.

app/tests/test_admin_page.py
```diff
@@ -342,6 +342,74 @@ def test_add_update_collection_with_provider(self):
...
+        self.assertEqual(
+            provider.description, None, msg="Admin page wrong provider.description on creation"
+        )
+
+    def test_add_update_collection_empty_provider_description(self):
+        # Login the user first
+        self.client.login(username=self.username, password=self.password)
+
+        collection, data, link, provider = self._create_collection(with_provider=True)
+
```
ltshb, line 392: A check that the provider has a description after creation would be good. To be sure that it is not already `None`.

ltshb (approved):
🥇

PR #299, opened 2021-06-21
BGDIINF_SB-1863: Added collection published flag

ltshb:
Added a flag in admin interface to un-published a collection. By default all collection are published. When un-published then the collection and its items won't be listed in the /collections and /search endpoints. However they are still accessible through their exact path.
The publish flag can also be set/unset via the API via an optional field (default to True)


app/stac_api/models.py
```diff
@@ -274,8 +274,16 @@ class Collection(
...
-        indexes = [models.Index(fields=['name'], name='collection_name_idx')]
+        indexes = [
+            models.Index(fields=['name'], name='collection_name_idx'),
+            models.Index(fields=['published'], name='collection_published_idx')
+        ]
 
+    published = models.BooleanField(
+        default=True,
+        help_text=
+        "When not published the collection doesn't appear on the api/stac/v0.9/collections endpoint"
```
hansmannj, line 285: Probably add in the help_text, that it is still accessible via the exact pass though, even when publishes=False, just to be clear about that.

app/tests/test_admin_page.py
```diff
@@ -53,6 +53,7 @@ def _create_collection(self, with_link=False, with_provider=False, extra=None):
             "name": "test_collection",
             "license": "free",
             "description": "some very important collection",
+            "published": "on",
```
hansmannj, line 56: Is it really `on` here, or should that be `true`?

ltshb: Yes and to disable it is not `"off"` like everyone would thought, but the field needs to be removed from the payload ;-)

Makefile
```diff
@@ -66,8 +66,13 @@ help:
 	@echo "- ci                       Create the python virtual environment and install requirements based on the Pipfile.lock"
 	@echo -e " \033[1mFORMATING, LINTING AND TESTING TOOLS TARGETS\033[0m "
 	@echo "- format                   Format the python source code"
+	@echo "- ci-check-format          Format the python source code and check if any files has changed. This is meant to be used by the CI."
 	@echo "- lint                     Lint the python source code"
 	@echo "- test                     Run the tests"
+	@echo -e " \033[1mSPEC TARGETS\033[0m "
+	@echo "- lint-spec                Lint the openapi spec"
+	@echo "- build-spec               Build the openapi spec"
```
rebert, line 74: S : maybe .PHONY build-spec instead of build-spec**s**

Makefile
```diff
@@ -159,6 +172,10 @@ build-specs:
 	cd spec && make build-specs
 
 
+.PHONY: lint-spec
```
rebert, line 175: nice !!!

rebert (approved):
LGTM - thanks for finding a validator!

Makefile
ltshb: Actually plural is more correct since we have 2 specs, openapi.yaml and openapitransactional.yaml. I corrected to have all specs target in plural.

hansmannj (approved):
Nice :1st_place_medal:

boecklic (approved):
:+1:

PR #300, opened 2021-06-22
BGDIINF_SB-1765: Clean up unittest stac_extension

ltshb:
We don't use stac_extension (see BGDIINF_SB-1410) there remove them from
unittest.


hansmannj (approved):
:+1:

PR #301, opened 2021-06-22
BGDIINF_SB-1864: Order API collections/items/assets by name

ltshb:
For the view using pagination, the ordering is done in the pagination class.
This one did not had the ability to easily configure different ordering
per view. This has been added via the get_ordering() method which checks
for ordering attribute in the view.

For the assets ordering within items, we needed to do the ordering in
the prefetch related and in the get_queryset() because assets don't uses
pagination.

I don't think this have a big impact on performance but those needs definitely to be tested with k6.io.

NOTE: to avoid conflict with migrations files with the PR #299  this PR is based on feature_BGDIINF_SB-1863_publish_flag. Once #299 is merged I'll set the branch to `develop` before merging this one.

### DIFF
--- a/app/stac_api/pagination.py
+++ b/app/stac_api/pagination.py
@@ -153,7 +153,7 @@ class CursorPagination(pagination.CursorPagination):
     def get_ordering(self, request, queryset, view):
         '''Get the ordering for the pagination.
 
-        The base class doesn't not allowed to use the view configuration for
+        The base class doesn't allow to use the view configuration for
         ordering, therefore it is here implemented.
         '''
         if hasattr(view, 'ordering'):
--- a/app/stac_api/pagination.py
+++ b/app/stac_api/pagination.py
@@ -150,6 +150,19 @@ class CursorPagination(pagination.CursorPagination):
     page_size_query_param = 'limit'
     max_page_size = settings.REST_FRAMEWORK['PAGE_SIZE_LIMIT']
 
+    def get_ordering(self, request, queryset, view):
+        '''Get the ordering for the pagination.
+
+        The base class doesn't not allowed to use the view configuration for
+        ordering, therefore it is here implemented.
+        '''
+        if hasattr(view, 'ordering'):
+            ordering = getattr(view, 'ordering')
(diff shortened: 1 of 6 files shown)


app/stac_api/pagination.py
```diff
@@ -150,6 +150,19 @@ class CursorPagination(pagination.CursorPagination):
     page_size_query_param = 'limit'
     max_page_size = settings.REST_FRAMEWORK['PAGE_SIZE_LIMIT']
 
+    def get_ordering(self, request, queryset, view):
+        '''Get the ordering for the pagination.
+
+        The base class doesn't not allowed to use the view configuration for
```
rebert, line 156: does not

app/tests/test_items_endpoint.py
```diff
@@ -35,20 +35,56 @@ class ItemsReadEndpointTestCase(StacBaseTestCase):
...
+        item_3 = self.factory.create_item_sample(self.collection, name='item-0', db_create=True)
+        assets = self.factory.create_asset_samples(
+            3, item_3.model, name=['asset-1.tiff', 'asset-0.tiff', 'asset-2.tiff'], db_create=True
+        )
         response = self.client.get(f"/{STAC_BASE_V}/collections/{self.collection.name}/items")
         self.assertStatusCode(200, response)
         json_data = response.json()
 
-        self.assertEqual(len(json_data['features']), 2, msg='Output should only have two features')
+        self.assertEqual(len(json_data['features']), 3, msg='Output should only have two features')
```
rebert, line 57: have **three** features

rebert (approved):
LGTM

boecklic (commented):
General question: is ordering not better defined at queryset/model level instead of pagination?

ltshb:
> General question: is ordering not better defined at queryset/model level instead of pagination?

@boecklic it would be indeed more intuitive, that's what I tried first, but the REST Framework Pagination is then reordering the query. So without hacking the REST Framework we need to do the ordering in the pagination.

ltshb:
Tested by @p1d1d1  with  browser stac.

PR #303, opened 2021-06-23
BGDIINF_SB-1725, BGDIINF_SB-1726: Removed POST collections/items/assets endpoints

ltshb:
Those endpoints are not needed anymore since we have the UPSERT PUT
method.

Addapted also unittests.


rebert (approved):
LGTM - you updated code, unittest and spec ...

PR #307, opened 2021-07-09
BGDIINF_SB-1898: Avoid crash if request is missing the content-type header.

ltshb:
On prod we had some non authorized requests without content-type header
that crashed the service.


boecklic (approved):

rebert (approved):

rebert:
thx

PR #309, opened 2021-07-12
BGDIINF_SB-1897: Don't crash in case of removing missing summaries

ltshb:
When updating the collection summaries and trying to remove a summary
from a list, if the summary (eo:gsd, geoadmin:variant or proj:epsg) is
missing from the list, then the update crash preventing the collection
summaries update.

This case should not happen, but if the collection is
corrupted for any reason like another small bug or data have been
manually altered, we cannot recover from it.

Now in this case we can recover from such situation, removing an non
existent summary is simply ignored as the result is what we expect
anyway.

Also correct a possible bug with the geoadmin:variant where checking
with `if variant:` is not true when variant == ''. In theory the variant
is never an empty string but if it would be we would end up again in a
corrupted summaries that we cannot correct.


app/stac_api/collection_summaries.py
```diff
@@ -130,7 +147,7 @@ def _update_summaries_on_asset_insert(self, asset):
         '''
         updated = False
         if (
-            asset.geoadmin_variant and
+            asset.geoadmin_variant is not None and
```
ltkum, line 150: This is more explicit, I like this.

ltkum (approved):
It looks to me as if this should do what we expect it to (stop crashing with an impossibility to recover). 

I suggest we create a few tests to check that we are not "Locked out" when the situation occurs, but this can be its own ticket / PR.

hansmannj (approved):
:+1: LGTM

ltshb:
This PR is deprecated due to PR #310 there I close it

PR #318, opened 2021-09-22
Added Asset PUT presigned URL error message in SPEC

ltshb:
Specified the possible error message on the PUT presigned URL operation in the SPEC.

### DIFF
--- a/spec/static/spec/v0.9/openapitransactional.yaml
+++ b/spec/static/spec/v0.9/openapitransactional.yaml
@@ -899,7 +899,9 @@ paths:
           example: yLLiDqX2OL7mcIMTjob60A==
       responses:
         "200":
-          description: Asset file uploaded part successfully
+          description: Asset file part uploaded part successfully (Response has no
+            content).
+          content: {}
           headers:
             ETag:
               schema:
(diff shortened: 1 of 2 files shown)


spec/static/spec/v0.9/openapitransactional.yaml
```diff
@@ -899,7 +899,8 @@ paths:
           example: yLLiDqX2OL7mcIMTjob60A==
       responses:
         "200":
-          description: Asset file uploaded part successfully
+          description: Asset file uploaded part successfully (No Content).
```
hansmannj, line 902: Could it also be "asset file part uploaded" (order of "uploaded" and "part" changed)?

spec/transaction/paths.yaml
```diff
@@ -580,7 +580,8 @@ paths:
             example: yLLiDqX2OL7mcIMTjob60A==
       responses:
         "200":
-          description: Asset file uploaded part successfully
+          description: Asset file uploaded part successfully (Response as no content).
```
hansmannj, line 583: Again, probably "asset file part uploaded" instead of "asset file uploaded part"?
And "response **h**as no content".

hansmannj (approved):
Thanks :+1:
Just two very minor comments

spec/transaction/paths.yaml
ltshb: Done, thanks.

PR #320, opened 2021-10-28
BGDIINF_SB-1983: `md5_parts` is now a mandatory parameter

ltkum:
Issue : md5_parts was an optional parameter while waiting for the users to adapt their scripts.
Fix: Now that those scripts are adapted, md5_parts is required for all uploads.

TODO: 
I'll check the tests, and ensure they work. 
I'll add a test that verifies that sending an upload request without the md5 fails

### DIFF
--- a/app/stac_api/serializers.py
+++ b/app/stac_api/serializers.py
@@ -4,6 +4,7 @@ from urllib.parse import urlparse
 
 from django.conf import settings
 from django.contrib.gis.geos import GEOSGeometry
+from django.core.exceptions import ValidationError
 
 from rest_framework import serializers
 from rest_framework.utils.serializer_helpers import ReturnDict
@@ -785,8 +786,7 @@ class AssetUploadSerializer(NonNullModelSerializer):
         allow_blank=False,
         validators=[validate_checksum_multihash_sha256]
     )
-    # TODO BGDIINF_SB-1983 make the md5_parts required
-    md5_parts = serializers.JSONField(required=False)
+    md5_parts = serializers.JSONField(required=True)
 
     # write only fields
     ended = serializers.DateTimeField(write_only=True, required=False)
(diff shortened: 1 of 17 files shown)


app/stac_api/serializers.py
```diff
@@ -802,10 +802,13 @@ class Meta:
...
 
     def validate(self, attrs):
-        # TODO BGDIINF_SB-1983 md5_parts should be required not optional
+        # get partial from kwards (if partial true and no md5 : ok, if false no md5 : error)
         # Check the md5 parts length
-        if 'md5_parts' in attrs:
+        if attrs.get('md5_parts') is not None:
             validate_md5_parts(attrs['md5_parts'], attrs['number_parts'])
+        elif not self.partial:
+            raise ValidationError('')
```
ltshb, line 810: Here the error message should be set, for example `'md5_parts' argument missing`

app/tests/test_asset_upload_endpoint.py
```diff
@@ -180,10 +180,16 @@ def test_asset_upload_create_abort_multipart(self):
         self.assertS3ObjectNotExists(key)
         number_parts = 2
         file_like, checksum_multihash = self.get_file_like_object(1 * KB)
+        offset = 1 * KB // number_parts
+        md5_parts = [{
+            'part_number': i + 1, 'md5': base64_md5(file_like[i * offset:(i + 1) * offset])
+        } for i in range(number_parts)]
```
ltshb, line 186: Because this snippet comes often, it would be better to have it as a helper method
```python
def compute_md5_parts(number_parts, offset, file_like):
    return [{
        'part_number': i + 1, 'md5': base64_md5(file_like[i * offset:(i + 1) * offset])
    } for i in range(number_parts)]
```

ltshb (changes requested):

app/stac_api/serializers.py
```diff
@@ -807,7 +807,7 @@ def validate(self, attrs):
         if attrs.get('md5_parts') is not None:
             validate_md5_parts(attrs['md5_parts'], attrs['number_parts'])
         elif not self.partial:
-            raise ValidationError('')
+            raise ValidationError('md5 parameter not set')
```
ltshb, line 810: Sorry to be picky, but the parameter name is `md5_parts`

app/stac_api/serializers.py
```diff
@@ -802,10 +802,13 @@ class Meta:
     aborted = serializers.SerializerMethodField()
 
     def validate(self, attrs):
-        # TODO BGDIINF_SB-1983 md5_parts should be required not optional
+        # get partial from kwards (if partial true and no md5 : ok, if false no md5 : error)
```
hansmannj, line 805: picky, sorry, just a small typo here. Guess it should be kwargs instead of kwards ;-)

hansmannj (approved):
Nice, thanks! Looks good to me :+1:

ltshb (approved):
@ltkum I've added a commit to consolidate the ValidationError and make use of some best practice from rest_framework.

PR #322, opened 2021-11-11
BGDIINF_SB-2033: adding generic zip media type

ltkum:
Issue : Generic zip files were not accepted as media types
Fix : We accept those as media types


boecklic (approved):
thanks :+1:

PR #328, opened 2021-12-01
BGDIINF_SB-2083: Changed parent link to match STAC spec

ltshb:
The parent link for a Collection or an Item, based on the STAC
spec should be respectively the landing page `/` (catalog) and the
collection `/<collection>` instead of the `/collections` and
`/collections/<collection>/items`. This is also required by the STAC
Browser, see (https://github.com/radiantearth/stac-browser/issues/105)

The generated links have also been consolidated in one method to
simplify the maintenance. For this simplification we base the generation
of the links based on the route name given in the urls.py. For the
DestroyMixin as it is used by several vues we could not use directly the
route name, therefore we added the route name to every view as name.


boecklic (approved):
:+1:

PR #330, opened 2022-01-10
Feature_KOGIS_PRJ-1382: script for multipart uploads for large asset files via api

hansmannj:
* Added script for uploading large asset files as multipart upload via API
* Added script for generating a 20 MB dummy asset file for testing
* Added the 20 MB dummy asset file for testing

- [x] Update the README.md describing how to call this script using summon to inject gopass secrets
- [x] Add the according `secrets.yml`


scripts/multipart_upload_via_API.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+def get_args():
+    parser = argparse.ArgumentParser(
+        description="Utility script for uploading large asset files as a multipart upload " \
+            "using the STAC API. When calling this script, summon can be used to inject secrets " \
+                "from gopass into the environment variables. To do this, call this script like " \
+                    "this: summon -p gopass -e dev|int|prod python3 multipart_upload_via_API.py " \
+                        "localhost|DEV|INT|PROD collection-name item-name assset-name " \
+                            "asset-file-path --part-size."
```
hansmannj, line 31: I'll move the description on how to use summon from this description here into the `README.md`

scripts/create_20MB_random_asset_file.py
```diff
@@ -0,0 +1,8 @@
+# this will create a 20 MB dummy asset file for testing the upload via
+# the admin GUI.
+
+import os
+
+FILE_SIZE = 20 * 1024**2  # this is 20 MB
+with open("20MB_sample_asset_file.zip", "wb") as dummy_file:
+    dummy_file.write(os.urandom(FILE_SIZE))
```
ltshb, line 8: You use `zip` extension but this is not a valid zip file which could be misleading while testing.  Might be a good idea to really zip the file to have a valid zip file.

ltshb: The file size could/should be an option parameter to the script and the size should not be part of the file name. The file should also be in the `.gitignore` because it doesn't needs to be under source control. Moreover NEVER put big binary files under source control unless it is absolutely necessary. This will in future especially when the file changes have an impact on git performance !

scripts/multipart_upload_via_API.py
```diff
@@ -0,0 +1,142 @@
...
+        hostname = "sys-data.int.bgdi.ch"
+    else:
+        hostname = "data.geo.admin.ch"
+
+    collection = args.collection
+    item = args.item
+    asset = args.asset
+    asset_path = f'collections/{collection}/items/{item}/assets/{asset}'
+    user = os.environ.get('STAC_USER', 'unknown-user')
+    password = os.environ.get('STAC_PASSWORD', 'unknown-password')
```
ltshb, line 75: Here if no credential is provided either simply let the script crash with `KeyError` or catch this error and stop the script with a correct message. Otherwise you will hide the reason why the script is not working, I'm pretty sure those unknown credential will never be valid.

ltshb (changes requested):
NEVER put useless big binary file under source control, this can have quite a big git performance impact in future, so remove the zip file from source control.
Also use a correct naming convention for the script below which should be snake case (snake case is always lower case with underscore)

scripts/create_20MB_random_asset_file.py
hansmannj: Makes sense, thanks!

scripts/multipart_upload_via_API.py
hansmannj: Good idea, I'll adapt the script here

.gitignore
```diff
@@ -71,6 +71,7 @@ nose2-junit.xml
 .volumes
 **/logs
 /scripts/xxxxxxl_asset_file.zip
+/scripts/dummy_asset_for_multipart_testing.zip
```
ltshb, line 74: Might be a good idea to use wildcard here `/scripts/**/*.zip`

ltshb (approved):
👍🏼

PR #332, opened 2022-01-12
feature_BGDIINF_SB-2135: improved error handling and added new script

hansmannj:
* improved error handling in case of no or wrong credentials are provided for multipart-upload script
* added new script for creating a new asset via the API


ltshb (approved):
Even a better approach for error handling with wrong credential, would be to simply print the warning and do an `exit(1)`. In this case you don't need the callstack to be printed and to raise an exception.

boecklic (approved):

PR #335, opened 2022-02-17
BGDIINF_SB-2196: Order collections by title

rebert:
Order the collection by title instead of its id.

The unittests had to be adapted accordingly


boecklic (approved):
:+1:

PR #340, opened 2022-03-31
BGDIINF_SB-2318: Make sure that Links, Providers and Assets are ordered

ltshb:
Also make sure that the Item Links have unique rel.

End to end pagination tests were failing because the order of the collection
links in the initial call was not the same as in the call of the previous link.

Having links, providers and assets ordered simplify testing.


hansmannj (approved):
LGTM :+1:

ltclm (approved):

PR #348, opened 2022-08-08
BGDIINF_SB-2520: Improved logging

ltshb:
See commit messages.

### DIFF
--- a/app/middleware/logging.py
+++ b/app/middleware/logging.py
@@ -28,7 +28,7 @@ class RequestResponseLoggingMiddleware:
         ] and request.content_type == "application/json" and not request.path.startswith(
             '/api/stac/admin'
         ):
-            extra["request.payload"] = str(request.body[:200])
+            extra["request.payload"] = request.body[:200].decode()
 
         logger.info(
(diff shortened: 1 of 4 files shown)


app/middleware/logging.py
```diff
@@ -53,7 +53,7 @@ def __call__(self, request):
         # HttpResponse and JSONResponse sure have
         # (e.g. WhiteNoiseFileResponse doesn't)
         if isinstance(response, (HttpResponse, JsonResponse)):
-            extra["response"]["payload"] = str(response.content)[:200]
+            extra["response"]["payload"] = response.content[:200].decode()
```
boecklic, line 56: out of interest: is there a difference between using `str()` and `.decode()`?

boecklic (approved):
:+1:

app/middleware/logging.py
ltshb: @boecklic yes, see below

```python
my_byte_string = b'My string'

print(str(my_byte_string))
# "b'My string'"

print(my_byte_string.decode()
# "My string"
```

PR #350, opened 2022-11-21
Added precision in spec for title min, max length

ltshb:
This PR has been opened based on question from a user.


hansmannj (approved):
A bit picky, I know: But probably the order of `minLength` and `maxLength` could be the same everywhere? Now it's sometimes first min then max (which I'd prefer) and sometimes max first and then min.

PR #351, opened 2022-11-23
BGDIINF_SB-2686: Allow removing the items properties_title with patch by setting it to null

rebert:


app/tests/test_items_endpoint.py
```diff
@@ -727,6 +728,23 @@ def test_item_endpoint_patch(self):
...
+        self.assertIn("title", json_data['properties'].keys())
+
+    def test_item_endpoint_remove_properties_title(self):
+        data = {"properties": {"title": None}}
+        path = f'/{STAC_BASE_V}/collections/{self.collection["name"]}/items/{self.item["name"]}'
+        response = self.client.patch(path, data=data, content_type="application/json")
+        json_data = response.json()
+        self.assertStatusCode(200, response)
+        self.assertEqual(self.item['name'], json_data['id'])
+        self.assertNotIn("title", json_data['properties'].keys())
```
ltshb, line 740: Would be good to first check that the item has a title before doing the PATCH.

ltshb (approved):
A unittest with PUT and a null title would be good as well.

hansmannj (approved):
LGTM :+1: Thanks!
Nice opportunity to dig in the STAC code :wink:

app/tests/test_items_endpoint.py
```diff
@@ -682,6 +682,44 @@ def test_item_endpoint_put_update_to_datetime_range(self):
...
+                "end_datetime": "2020-10-19T00:00:00Z",
+            }
+        )
+        response = self.client.put(
+            path, data=sample.get_json('put'), content_type="application/json"
+        )
+        json_data = response.json()
+        self.assertStatusCode(200, response)
+        self.assertEqual(self.item['name'], json_data['id'])
+        self.assertIn("title", json_data['properties'].keys())
```
ltshb, line 721: Should it not be the other way around
1. Create an item with title
2. Update the item with PUT `title: null`

ltshb (approved):

app/tests/test_items_endpoint.py
hansmannj: Or probably both?
Check, that with PUT the title of an existing item with title can be "nulled" and also check, that with PUT an item with title=None can be created from scratch.

PR #352, opened 2022-12-05
BGDIINF_SB-2713: Fixed CORS headers for landing page (STAC Browser)

ltshb:
The STAC browser might tries to access the landing page without trailing slash:

`/api/stac/v0.9`

Which would be automatically redirected to `/api/stac/v0.9/` by the
`django.middleware.common.CommonMiddleware`.

Due to this we need to put the CORS header middleware before the `CommonMiddleware`
otherwise the redirect would not have the correct CORS headers and the STAC
Browser would not be able to access our API from another origin.


ltshb:
> thanks @ltshb . What about adding an e2etest to cover this?

Good idea

PR #353, opened 2022-12-05
BGDIINF_SB-2714: Made Gunicorn worker tmp dir configurable

ltshb:
This is required in order to assure that the tmp dir is on a tempfs file system
to improve performance, see https://docs.gunicorn.org/en/stable/settings.html#worker-tmp-dir.


rebert (approved):
thx

PR #355, opened 2023-01-05
BGDIINF_SB-2693: update to python 3.9 and dependencies

jedef:
- Updated to python 3.9 and all dependencies
- Modified the Dockerfile to use a build stage to increase security. Also rearranged the statements to increase cache hits when building the docker image.
- Reduced the number of errors from the logging utilities shown in the console.

Still to do for this ticket in a later pull request:
-  Fix a log error that it caused by two different object being both stored as request in extra parameters


Dockerfile
```diff
@@ -1,93 +1,131 @@
...
-    && useradd -r -s /bin/false -g geoadmin geoadmin \
-    && mkdir -p /logs && chown geoadmin:geoadmin /logs \
-    # install relevent packages
-    && apt-get update \
-    && apt-get install -y binutils libproj-dev gdal-bin \
-    # the following line contains debug tools that can be removed later
+# Container that contains basic configurations used by all other containers
+# It should only contain variables that don't change or change very infrequently
+# so that the cache is not needlessly invalidated
+FROM python:3.9-slim-bullseye as base
```
ltshb, line 5: We should stick to the same base image: `3.9-slim-buster` or is there any reason using the bullseye base image ?

Dockerfile
```diff
@@ -1,93 +1,131 @@
...
+COPY --chown=${USER}:${GROUP} app/ app/
+
+# Activate virtualenv
+ENV VIRTUAL_ENV=${INSTALL_DIR}/.venv
+ENV PATH="$VIRTUAL_ENV/bin:$PATH"
+
+# this is only used with the docker-compose setup within CI
+# to ensure that the app is only started once the DB container
+# is ready
+COPY ./wait-for-it.sh /app/
```
ltshb, line 53: I would move this after the installation of pip dependencies in line 37. This script is unlikely to change in opposition to the files in `spec` or `app` folders.

Dockerfile
```diff
@@ -1,93 +1,131 @@
...
     curl net-tools iputils-ping postgresql-client-common jq openssh-client \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/* \
     && pip3 install pipenv \
     && pipenv --version
 
-COPY Pipfile* /tmp/
-RUN cd /tmp && \
-    pipenv install --system --deploy --ignore-pipfile
+ENV PIPENV_VENV_IN_PROJECT=1
```
ltshb, line 35: This could also move into the base image.

Dockerfile
```diff
@@ -1,93 +1,131 @@
...
+    # dev dependencies
+    binutils libproj-dev \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/* \
+    && pip3 install pipenv \
+    && pipenv --version
 
-USER geoadmin
+ENV PIPENV_VENV_IN_PROJECT=1
+COPY Pipfile.lock Pipfile ./
```
ltshb, line 85: Pipfiles should not be copied in the installation dir as they should not be part of the final production image.

Dockerfile
```diff
@@ -1,93 +1,131 @@
...
-
 LABEL target=production
 
-# on prod, settings.py just import settings_prod
-RUN echo "from .settings_prod import *" > /app/config/settings.py \
-    && chown geoadmin:geoadmin /app/config/settings.py
+RUN apt-get update -qq \
+    && DEBIAN_FRONTEND=noninteractive apt-get install -y gdal-bin \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
```
ltshb, line 105: This layer could be moved in the base image as it used on all images.

Dockerfile
```diff
@@ -1,93 +1,131 @@
...
+ENV HTTP_PORT=8080
+ENV USER=geoadmin
+ENV GROUP=geoadmin
+ENV INSTALL_DIR=/opt/service-stac
+
+RUN groupadd -r ${GROUP} \
+    && useradd -r -s /bin/false -g ${GROUP} ${USER} \
+    && mkdir -p ${INSTALL_DIR}/logs && chown ${USER}:${GROUP} ${INSTALL_DIR}/logs
+
+WORKDIR ${INSTALL_DIR}
```
ltshb, line 15: I would not change the workdir here in the base image, it is then quite confusing when reading the file. Workdir should change only in the debug and production image at the end of debug and production image. See also comment about the pipfiles. The goal of the builder is to have only what is needed to run the application on the production images while the builder image has everything needed to build/compile/transpile/install the application.

app/config/settings_prod.py
```diff
@@ -188,6 +188,9 @@
 STATIC_HOST = os.environ.get('DJANGO_STATIC_HOST', '')
 STATIC_URL = f'{STATIC_HOST}/api/stac/static/'
 STATIC_SPEC_URL = f'{STATIC_URL}spec/v0.9/'
+# "manage.py collectstatic" will copy all static files to this directory, and
+# whitenoise will serve the static files that are in this directory (unless DEBUG=true in which case
+# it will serve the files from the same directories "manage.py collectstatic" collects data from)
```
ltshb, line 193: 👍🏼

app/tests/test_generic_api.py
```diff
@@ -404,19 +404,19 @@ class ApiCacheHeaderTestCase(StacBaseTestCase):
 
     @classmethod
     @mock_s3_asset_file
-    def setUpTestData(cls):
```
ltshb, line 407: Why changing it from `setUpTestData` to `setUp` ?

NOTE: is this is really required the `@classmethod` needs also to be removed.

app/tests/test_generic_api.py
```diff
@@ -491,15 +491,15 @@ class ApiCORSHeaderTestCase(StacBaseTestCase):
...
```
ltshb, line 494: Why changing it from `setUpTestData` to `setUp` ?

app/tests/test_serializer.py
```diff
@@ -318,15 +318,13 @@ class ItemSerializationTestCase(StacBaseTestCase):
 
     @classmethod
     @mock_s3_asset_file
-    def setUpTestData(cls):  # pylint: disable=invalid-name
```
ltshb, line 321: Why changing it from `setUpTestData` to `setUp` ?

app/tests/test_serializer.py
```diff
@@ -575,15 +573,13 @@ class AssetSerializationTestCase(StacBaseTestCase):
 
     @classmethod
     @mock_s3_asset_file
-    def setUpTestData(cls):
```
ltshb, line 578: Why changing it from `setUpTestData` to `setUp` ?

update_to_latest.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
```
ltshb, line 1: Good idea this script, but I would add a small comment at the top of what it does and how to use it. Also maybe add a small topic in the readme file about this.

Dockerfile
```diff
@@ -1,93 +1,131 @@
...
 
-# Collect static files, uses the .env.default settings to avoid django raising settings error
-RUN APP_ENV=default LOGGING_CFG=0 ./manage.py collectstatic --noinput
+COPY --from=builder ${INSTALL_DIR} ${INSTALL_DIR}
 
-# production container must not run as root
-USER geoadmin
+# Activate virtual environnment
+ENV VIRTUAL_ENV=${INSTALL_DIR}/.venv
+ENV PATH="$VIRTUAL_ENV/bin:$PATH"
```
ltshb, line 111: To activate I would also make sure that the `PYTHONHOME` environment variable is unset, like does the `activate` script.

ltshb (commented):
Don't forget to change the PR title for the release notes.

Dockerfile
jedef: It's just that bullseye is the newest debian release and buster the older (but still supported) debian release. However, the python images seem to support only the last two debian releases and a new debian release should be published this summer, so thats why I thought that it was a good idea to already upgrade now since I also updated the other dependencies.

ltshb: Ok then that's fine using the last debian version if it is a stable version ?

Dockerfile
jedef: I decided to delete the Pipfile after installation as the alternative (put the Pipfile in another dir and manually move the venv after installation) seemed more error prone to me.

Dockerfile
jedef: Because it is impossible to use paths relative to the `${INSTALL_DIR}` if I don't specify the WORKDIR

Dockerfile
jedef: The problem is then that if you change the list of debug tools of dev dependencies to install, you could then land with a mismatch of versions (an old version of gdal-bin and new versions of the other dependencies) thats why I thought that it's better to install everything that comes from apt in a single step.

ltshb: Ok

Dockerfile
ltshb: Yes I think this would make sense and be a bit clearer where stuff are copied.

Dockerfile
ltshb: Or more simply put the pipfiles in a source directory, install dependencies there and copy only the venv over the prod images:

```
# Builder image
# Virtual Env
ENV SRC_DIR=/usr/src
COPY Pipfile.lock Pipfile ${SRC_DIR}/
WORKDIR ${SRC_DIR}
RUN /root/.local/bin/pipenv sync 

# App
COPY --chown=${USER}:${GROUP} .env.default ${INSTALL_DIR}/
COPY --chown=${USER}:${GROUP} spec/ ${INSTALL_DIR}/spec/
COPY --chown=${USER}:${GROUP} app/ ${INSTALL_DIR}/app/

# PROD image
COPY --from=builder --chown=${USER}:${GROUP} ${SRC_DIR}/.venv ${INSTALL_DIR}/* ${INSTALL_DIR}/
```

Something like this (the wildcard in copy need to be checked if it works). 

But I would rather be more explicit on the installation location and what to install and avoid removing files at the end.

ltshb: I mean the builder image is not meant to run the application which means that the venv might be in another directory as the application.

ltshb:
@jedef don't forget to change the PR title

Dockerfile
jedef: Ok perfect then I will change that.

app/tests/test_generic_api.py
jedef: It has to do with the "@mock_s3_asset_file" decoration, if I remember correctly. If I don't change this, the following error appears: `ValueError: I/O operation on closed file.` I couldn't totally understand the root of the problem, but if you think it's important, I can try to dig deeper into that.

Dockerfile
jedef: Ok I did it like this

Dockerfile
jedef: Finally I have still added PYTHONHOME now, tell me if it is good like this.

jedef:
Django-pgtrigger upgrade guide that lists the breaking changes:
https://django-pgtrigger.readthedocs.io/en/4.6.0/upgrading.html

For what I have seen, only two things are relevant for us:
- All triggers now appear in migrations when running python manage.py makemigrations. Triggers from version 2 will appear as new AddTrigger operations. They will succeed when running migrate even if previously installed. Remember, however, that triggers will be deleted if the migrations are reversed.
- Version 2.5 introduced the ability to register triggers on your model’s Meta.triggers list. User can still use pgtrigger.register to register triggers programmatically, but it has been deprecated.

jedef:
Backward incompatible changes of django: https://docs.djangoproject.com/en/4.1/releases/4.0/#backwards-incompatible-4-0
However, I think that nothing is relevant for our app

ltshb (commented):
Looks good to me, just one concern about the pgtrigger, I did not dig into the changes and into the impact on migration. We need to keep in mind that on PROD we have a lot of data (>millions). Maybe @ltclm could have a deeper look with you about this change and its impact on the DB ?
Did you do manual testing ? 

- [ ] Test local using django runserver, gunicorn and docker-compose
- [ ] Admin site, browsing, creating collection, item, asset, editing, deleting
- [ ] Browsing using read endpoints
- [ ] Using the STAC browser https://radiantearth.github.io/stac-browser/#/
- [ ] Test migration with a lots of assets
- [ ]

ltclm:
@jedef or @ltshb  make setup is not working on our shared development instance in the vpc `development.dev.bgdi.ch` with `python 3.9.15`

```bash
[ltclm@development]:[~/service-stac]:[feat-BGDIINF_SB-2693-update-to-python-3.9-and-dependencies]$ make setup
mkdir -p .timestamps
test -e app/config/settings.py || echo "from .settings_dev import *" > app/config/settings.py
test -e /home/ltclm/service-stac/.env.local || (cp /home/ltclm/service-stac/.env.default /home/ltclm/service-stac/.env.local && \
echo -e "\n  \e[91ma local .env.local was created, adapt it to your needs\e[0m")
touch .timestamps/.settins.timestamp
# Create virtual env with all packages for development
pipenv install --dev
Creating a virtualenv for this project…
Using /usr/bin/python3.9 (3.9.15) to create virtualenv…
...
ERROR: Exception:
Traceback (most recent call last):
  File "/home/ltclm/.local/share/virtualenvs/service-stac-zbjZGBHo/lib/python3.9/site-packages/pip/_vendor/packaging/markers.py", line 278, in __init__
    self._markers = _coerce_parse_result(MARKER.parseString(marker))
  File "/home/ltclm/.local/share/virtualenvs/service-stac-zbjZGBHo/lib/python3.9/site-packages/pip/_vendor/pyparsing/core.py", line 1141, in parse_string
    raise exc.with_traceback(None)
pip._vendor.pyparsing.exceptions.ParseException: Expected {Group:({'implementation_version' | 'platform_python_implementation' | 'implementation_name' | 'python_full_version' | 'platform_release' | 'platform_version' | 'platform_machine' | 'platform_system' | 'python_version' | 'sys_platform' | 'os_name' | 'os.name' | 'sys.platform' | 'platform.version' | 'platform.machine' | 'platform.python_implementation' | 'python_implementation' | 'extra'} | {string enclosed in "'" | string enclosed in '"'}} {'===' | '==' | '>=' | '<=' | '!=' | '~=' | '>' | '<' | 'not in' | 'in'} {{'implementation_version' | 'platform_python_implementation' | 'implementation_name' | 'python_full_version' | 'platform_release' | 'platform_version' | 'platform_machine' | 'platform_system' | 'python_version' | 'sys_platform' | 'os_name' | 'os.name' | 'sys.platform' | 'platform.version' | 'platform.machine' | 'platform.python_implementation' | 'python_implementation' | 'extra'} | {string enclosed in "'" | string enclosed in '"'}) | Group:({{Suppress:('(') : ...} Suppress:(')')})}, found 'None'  (at char 0), (line:1, col:1)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/ltclm/.local/share/virtualenvs/service-stac-zbjZGBHo/lib/python3.9/site-packages/pip/_internal/cli/base_command.py", line 160, in exc_logging_wrapper
    status = run_func(*args)
  File "/home/ltclm/.local/share/virtualenvs/service-stac-zbjZGBHo/lib/python3.9/site-packages/pip/_internal/cli/req_command.py", line 247, in wrapper
    return func(self, options, args)
  File "/home/ltclm/.local/share/virtualenvs/service-stac-zbjZGBHo/lib/python3.9/site-packages/pip/_internal/commands/install.py", line 344, in run
    reqs = self.get_requirements(args, options, finder, session)
  File "/home/ltclm/.local/share/virtualenvs/service-stac-zbjZGBHo/lib/python3.9/site-packages/pip/_internal/cli/req_command.py", line 436, in get_requirements
    req_to_add = install_req_from_parsed_requirement(
  File "/home/ltclm/.local/share/virtualenvs/service-stac-zbjZGBHo/lib/python3.9/site-packages/pip/_internal/req/constructors.py", line 471, in install_req_from_parsed_requirement
    req = install_req_from_line(
  File "/home/ltclm/.local/share/virtualenvs/service-stac-zbjZGBHo/lib/python3.9/site-packages/pip/_internal/req/constructors.py", line 393, in install_req_from_line
    parts = parse_req_from_line(name, line_source)
  File "/home/ltclm/.local/share/virtualenvs/service-stac-zbjZGBHo/lib/python3.9/site-packages/pip/_internal/req/constructors.py", line 297, in parse_req_from_line
    markers = Marker(markers_as_string)
  File "/home/ltclm/.local/share/virtualenvs/service-stac-zbjZGBHo/lib/python3.9/site-packages/pip/_vendor/packaging/markers.py", line 280, in __init__
    raise InvalidMarker(
pip._vendor.packaging.markers.InvalidMarker: Invalid marker: 'None', parse error at 'None'

  ☤  ▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉ 0/1 — 00:00:00
make: *** [Makefile:116: setup] Fehler 2
``` 
should this be working with python 3.9.15?
am i doing something wrong?

ltclm:
@ltshb @jedef 

> -[ ] compare pgtrigger result (DDL) of current prod with this pr @ltclm

the migration `0022_asset_update_asset_collection_summaries_trigger_and_more`  on database level is just re-creating the trigger functions. This is applied in a few ms and has no impact on the application. The changes in the trigger functions are only internal, they are transparent for the application:
```sql
-- p.e. 
_pgtrigger_should_ignore(TG_TABLE_NAME, TG_NAME)
-- will be replaced with
"public"._pgtrigger_should_ignore(TG_NAME)
```

i have ran the migration on a copy of the current stac prod database (670MB). it took about 250ms for the migration.

fyi: the sql statements triggered by the django migration can be visualized with:

```bash
./app/manage.py sqlmigrate stac_api 0022 > migrate_0022.sql

```

buildspec.yml
```diff
@@ -14,9 +14,10 @@ env:
 phases:
   install:
     runtime-versions:
-      python: 3.7
+      python: 3.9
     commands:
       - echo "Installing necessary dependencies"
+      - curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add -
```
ltshb, line 20: Add a comment why this has been added, in theorie this should be fixed by AWS Codebuild base image (managed by AWS) and could be removed here in future.

ltshb (commented):
Very good.

Before merging this we need first to prepare the logging configuration changes in https://github.com/geoadmin/infra-kubernetes repository

app/config/logging-cfg-local.yml
```diff
@@ -121,6 +129,7 @@ handlers:
     formatter: standard
     stream: ext://sys.stdout
     filters:
+      - type_validator
       - isotime
       - django
       - console_django_filter_info
```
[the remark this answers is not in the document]
jedef: No I already looked at this, filters directly attached to loggers do not propagate.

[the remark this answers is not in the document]
ltshb: OK

buildspec.yml
jedef: done

PR #356, opened 2023-01-17
BGDIINF_SB-2704: set custom cache-control headers for some temporary

ltclm:
5xx responses

### DIFF
--- a/app/middleware/cache_headers.py
+++ b/app/middleware/cache_headers.py
@@ -28,6 +28,12 @@ class CacheHeadersMiddleware:
 
         response = self.get_response(request)
 
+        public_directives = {"public": True}
+
+        error_directives = {
+            "public": True, "must-revalidate": True, "no-cache": True, "no-store": True
+        }
+
         # Code to be executed for each request/response after
         # the view is called.
         # match /xxx or /api/stac/xxx
@@ -45,6 +51,9 @@ class CacheHeadersMiddleware:
                 extra={"request": request}
             )
             patch_response_headers(response, settings.CACHE_MIDDLEWARE_SECONDS)
-            patch_cache_control(response, public=True)
+            directive = error_directives if response.status_code in (
+                502, 503, 504, 507
+            ) else public_directives
+            patch_cache_control(response, **directive)
 
         return response
--- a/app/middleware/cache_headers.py
+++ b/app/middleware/cache_headers.py
@@ -28,18 +28,20 @@ class CacheHeadersMiddleware:
 
         response = self.get_response(request)
 
-        public_directives = {"public": True}
-
-        error_directives = {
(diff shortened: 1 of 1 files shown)


app/middleware/cache_headers.py
```diff
@@ -28,6 +28,12 @@ def __call__(self, request):
 
         response = self.get_response(request)
 
+        public_directives = {"public": True}
+
+        error_directives = {
+            "public": True, "must-revalidate": True, "no-cache": True, "no-store": True
+        }
```
ltshb, line 35: For error why not using the django `add_never_cache_headers` ?

app/middleware/cache_headers.py
```diff
@@ -45,6 +51,9 @@ def __call__(self, request):
                 extra={"request": request}
             )
             patch_response_headers(response, settings.CACHE_MIDDLEWARE_SECONDS)
-            patch_cache_control(response, public=True)
+            directive = error_directives if response.status_code in (
+                502, 503, 504, 507
+            ) else public_directives
+            patch_cache_control(response, **directive)
```
ltshb, line 57: Patching here would not patch error on STATIC endpoints.

ltshb (commented):
I think it would be simpler and safer to simply enhance the if on line 41. I also noticed that we did not added the cache settings to OPTIONS, but I think we should. this could be re-written like follow to be very 
clear.
 
```python
response = self.get_response(request)

if request.method not in ('GET', 'HEAD', 'OPTIONS'):
    return response

# Patch cache settings only for GET HEAD or OPTIONS

if (
    re.match(fr'^(/{STAC_BASE})?/\w+$', request.path)
    or response.status_code in (502, 503, 504, 507)
):
    add_never_cache_headers(response)
elif response.status_code >= 500:
    patch_cache_control(response, public=True)
    patch_response_headers(response, cache_timeout=10)
elif request.path.startswith(urlparse(settings.STATIC_URL).path):
....
```

app/middleware/cache_headers.py
ltclm: @ltshb how do you want to patch static endpoints? in the original code they are not patched at all or everything but static is patched as follows.
```python
        elif (
            request.method in ('GET', 'HEAD') and
            not request.path.startswith(urlparse(settings.STATIC_URL).path)
        ):
            logger.debug(
                "Patching cache headers for request %s %s",
                request.method,
                request.path,
                extra={"request": request}
            )
            patch_response_headers(response, settings.CACHE_MIDDLEWARE_SECONDS)
            patch_cache_control(response, public=True)
```

ltshb: I meant with your previous code if an error occurred on a static endpoint it would not have set the cache control header falling back to CF default, now with your last commit error are patch regardless of the path.

ltshb (approved):
👍🏼

PR #358, opened 2023-01-27
BGDIINF_SB-2714: Dockerfile now sets worker_tmp_dir

jedef:
Fixed the worker tmp dir to a known directory in Dockerfile so that k8s can then mount a memory file system on that dir. This should increase performance at least by a factor 2.

Already deployed on dev last Friday.

### DIFF
--- a/docker-compose-ci.yml
+++ b/docker-compose-ci.yml
@@ -19,6 +19,9 @@ services:
     command: test --verbosity=2 --parallel=10 --no-input
     depends_on:
       - db
+    volumes:
+      - type: tmpfs
+        target: /tmp/gunicorn-workers
     environment:
       - DEBUG=True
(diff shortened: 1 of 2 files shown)


ltshb (approved):

docker-compose-ci.yml
```diff
@@ -19,6 +19,9 @@ services:
     command: test --verbosity=2 --parallel=10 --no-input
     depends_on:
       - db
+    volumes:
+      - type: tmpfs
+        target: /tmp/gunicorn-workers
```
ltclm, line 24: @jedef shouldn't we have the same folder here as in Dockerfile line 12?
`/tmp/gunicorn_workers`

jedef: Yes good catch I fixed that.

PR #359, opened 2023-02-10
BGDIINF_SB-2707: Now serializer is more permissive with mime types

jedef:
Before, the serializer only accepted exact matches. Now it accepts any
equivalent media types, i.e.:
- Optional spaces between the parameters
- Types, subtypes and parameter names are case insensitive
- Order of parameters not important

The database however still only accepts the "normalized" variant.

### DIFF
--- a/app/stac_api/models.py
+++ b/app/stac_api/models.py
@@ -37,6 +37,7 @@ from stac_api.validators import validate_geoadmin_variant
 from stac_api.validators import validate_geometry
 from stac_api.validators import validate_item_properties_datetimes
 from stac_api.validators import validate_link_rel
+from stac_api.validators import validate_media_type
 from stac_api.validators import validate_name
 
 logger = logging.getLogger(__name__)
@@ -519,7 +520,9 @@ class Asset(models.Model):
     proj_epsg = models.IntegerField(null=True, blank=True)
     # here we need to set blank=True otherwise the field is as required in the admin interface
     title = models.CharField(max_length=255, null=True, blank=True)
-    media_choices = [(x[0], f'{x[1]} ({x[0]})') for x in MEDIA_TYPES]
+    media_choices = [
+        (x.media_type_str, f'{x.description} ({x.media_type_str})') for x in MEDIA_TYPES
+    ]
     media_type = models.CharField(choices=media_choices, max_length=200, blank=False, null=False)
 
     created = models.DateTimeField(auto_now_add=True)
@@ -543,7 +546,11 @@ class Asset(models.Model):
             raise ValidationError(error.args[0]) from None
 
     def clean(self):
-        validate_asset_name_with_media_type(self.name, self.media_type)
+        # Although the media type is already validated, it still needs to be validated a second
+        # time as the clean method is run even if the field validation failed and there is no way
+        # to check what errors were already raised.
+        media_type = validate_media_type(self.media_type)
+        validate_asset_name_with_media_type(self.name, media_type)
 
 
 class AssetUpload(models.Model):
--- a/app/stac_api/serializers.py
+++ b/app/stac_api/serializers.py
@@ -468,8 +468,8 @@ class AssetBaseSerializer(NonNullModelSerializer, UpsertModelSerializerMixin):
         required=False, max_length=255, allow_null=True, allow_blank=False
     )
     description = serializers.CharField(required=False, allow_blank=False, allow_null=True)
-    # Can't be a ChoiceField, as  the validate method normalizes the MIME string only after it
-    # is read. Consistency is nevertheless guaranteed by the validate method.
+    # Can't be a ChoiceField, as the validate method normalizes the MIME string only after it
+    # is read. Consistency is nevertheless guaranteed by the validate() and validate_type() methods.
     type = serializers.CharField(
         source='media_type', required=True, allow_null=False, allow_blank=False
     )
@@ -519,6 +519,8 @@ class AssetBaseSerializer(NonNullModelSerializer, UpsertModelSerializerMixin):
         return asset, created
 
     def validate_type(self, value):
+        ''' Validates the the field "type"
+        '''
         return normalize_and_validate_media_type(value)
 
     def validate(self, attrs):
--- a/app/stac_api/serializers.py
+++ b/app/stac_api/serializers.py
@@ -30,7 +30,7 @@ from stac_api.utils import build_asset_href
 from stac_api.utils import get_browser_url
 from stac_api.utils import get_url
 from stac_api.utils import isoformat
-from stac_api.validators import MEDIA_TYPES_MIMES
+from stac_api.validators import normalize_and_validate_media_type
 from stac_api.validators import validate_asset_name
 from stac_api.validators import validate_asset_name_with_media_type
 from stac_api.validators import validate_checksum_multihash_sha256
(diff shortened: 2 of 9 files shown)


app/stac_api/serializers.py
```diff
@@ -468,12 +468,10 @@ class Meta:
...
-    type = serializers.ChoiceField(
-        source='media_type',
-        required=True,
-        choices=MEDIA_TYPES_MIMES,
-        allow_null=False,
-        allow_blank=False
+    # Can't be a ChoiceField, as  the validate method normalizes the MIME string only after it
+    # is read. Consistency is nevertheless guaranteed by the validate method.
+    type = serializers.CharField(
+        source='media_type', required=True, allow_null=False, allow_blank=False
```
ltshb, line 474: Does this have an impact on the admin interface ? In the admin interface we had a dropdown button, for the admin user this should be kept as dropdown button.

app/stac_api/validators.py
```diff
@@ -18,63 +20,108 @@
...
+    except ValueError:
+        return (mime_string, frozenset())
+
+
+MEDIA_TYPES_EXTENSIONS = frozenset([ext for media_type in MEDIA_TYPES for ext in media_type[2]])
+_MEDIA_TYPES_BY_COMPARABLE = {get_mime_comparable(media[0]): media for media in MEDIA_TYPES}
+_MEDIA_TYPES_BY_MIME_STRING = {media[0]: media for media in MEDIA_TYPES}
+
+
+def get_media_type_or_none(mime_string: str) -> Union[MediaType, None]:
```
ltshb, line 119: Don't use the type notation, to be consistent with the rest of the code.

app/stac_api/validators.py
```diff
@@ -107,27 +154,37 @@ def validate_asset_name(name):
         )
 
 
-def validate_asset_name_with_media_type(name, media_type):
+def validate_and_get_media_type(
+    asset_name, mime_string: str, mime_string_is_normalized=True
+) -> MediaType:
```
ltshb, line 159: Dito don't use the type notation

app/stac_api/validators.py
```diff
@@ -18,63 +20,108 @@
...
+    The rules for equivalence are:
+    - Types, subtypes and parameter names are case insensitive.
+    - Parameter values may be case sensitive (this function interprets this statement as "always
+      case sensitive")
+    - Order of the parameters is not important
+    - no whitespace around "=" allowed. Whitespace only allowed around the ";" parameter separator
+    - Parameter values are allowed to be quoted
+    '''
+    full_type, *arguments = map(lambda x: x.strip(), mime_string.split(';'))
+    full_type = full_type.lower()
```
ltshb, line 102: The `lower()` could/should be integrated in the lambda `lambda x: x.strip().lower()`
`full_type` is misleading, it contains only the type without parameters, should be `media_type`. 
I don't really like the mix of naming of *media type* and *mime type* or *mime string*. MIME type is the former name for media type and is the same thing. I would keep the new name media type everywhere (even though mime is a shorter name).
So instead of `mime_string` could be `media_type_str` or simply `media_type`

app/stac_api/validators.py
```diff
@@ -18,63 +20,108 @@
...
+    try:
+        for i, argument in enumerate(arguments):
+            name, value = argument.split('=')
+            arguments[i] = (
+                name.lower(), value if not quoted_string_pattern.match(value) else value[1:-1]
+            )
+        return (full_type, frozenset(arguments))
+    except ValueError:
+        return (mime_string, frozenset())
```
ltshb, line 111: Is this really needed to strip the quoted argument value ?

jedef: Only the ternary operator on line 107 is needed for that. The rest is needed to accept spaces between the equal signs, to be case insensitive on the mime_string and the parameter name but NOT the parameter value, and to accept the parameters in a random order (thats why I put them in a set, because a set does not have an order).

app/stac_api/validators.py
jedef: If I put lower() in the lambda, it will also lowercase the the parameter values, and according to the specs, if a parameter value is case independent depends on the parameter name, and that's too complicated to implement, that's why I decided to keep all parameter values case dependent.

jedef: The goal was to distinguish the MediaType tuple object and the media type string. So I renamed the "mime_string" to "media_type_str"

app/stac_api/models.py
```diff
@@ -543,7 +546,8 @@ def delete(self, *args, **kwargs):  # pylint: disable=signature-differs
             raise ValidationError(error.args[0]) from None
 
     def clean(self):
-        validate_asset_name_with_media_type(self.name, self.media_type)
+        media_type = validate_and_get_media_type(self.name, self.media_type)
```
ltshb, line 549: I'm not sure anymore, but I think this method `validate_and_get_media_type()` could be simplified and should not be needed here.
The `clean()` method is used for validation (also clean is a bad word for this but comes from django) of fields dependencies. I think we should separate the media type validation from the the media type and asset name dependency validation. I mean leaving here the validation of the asset name with the media type and add a media type validator in the field for checking the media type validity.
This way you would ensure that the media type in DB is always a valid normalized string.

app/stac_api/serializers.py
```diff
@@ -521,13 +519,16 @@ def update_or_create(self, look_up, validated_data):
...
+        name = attrs['name'] if not self.partial else attrs.get('name', self.instance.name)
+        if self.partial and 'media_type' not in attrs:
+            media_type = validate_and_get_media_type(name, self.instance.media_type)
+            validate_asset_name_with_media_type(name, media_type)
+        else:
+            media_type = validate_and_get_media_type(
+                name, attrs['media_type'], media_type_str_is_normalized=False
             )
+            validate_asset_name_with_media_type(name, media_type)
+            attrs['media_type'] = media_type.media_type_str
```
ltshb, line 531: As for the model in the `validate()` method I would try to keep only the field dependencies validations. And try to validate/normalize the media type in a field validator.

ltshb, line 531: I'm not sure it is a good idea to overwrite the media type for the normalization here in the validate method. Is it not possible to do this in the field validation ?

ltshb: After looking again at the django rest framework docu, I think the code could be simplified and made more understandable by using a `CustomField` https://www.django-rest-framework.org/api-guide/fields/#custom-fields. The `to_representation()` method would return the normalized media type while the `to_internal()` method would parse the media type string and raise a ValidationError in case of invalid string and return a `MediaType` object. I would also create a `MediaType` class instead of a named tuple, the class would then have a `__str__()` method that would normalize the media type. The `get_media_type_comparable()` function should then be rename `parse_media_type()` and would return a MediaType object instance.
In theorie in the model we don't need any media type validation as they are coming either from the serializer or from the admin interface choice field, but I would keep/add there a very simple quick validation without normalization (a validate function that checks that the string is in the media type dictionary. The media type dictionary would use the normalize media type as key.

app/stac_api/serializers.py
```diff
@@ -468,12 +468,10 @@ class Meta:
...
     )
     description = serializers.CharField(required=False, allow_blank=False, allow_null=True)
-    type = serializers.ChoiceField(
-        source='media_type',
-        required=True,
-        choices=MEDIA_TYPES_MIMES,
-        allow_null=False,
-        allow_blank=False
+    # Can't be a ChoiceField, as  the validate method normalizes the MIME string only after it
+    # is read. Consistency is nevertheless guaranteed by the validate method.
```
ltshb, line 472: Maybe add the validate method name in the comment `validate_type()`

app/stac_api/serializers.py
```diff
@@ -520,14 +518,15 @@ def update_or_create(self, look_up, validated_data):
         asset, created = Asset.objects.update_or_create(**look_up, defaults=validated_data)
         return asset, created
 
+    def validate_type(self, value):
```
ltshb, line 521: Maybe add a comment that this method validate the field `type`. `validate_type()` is so generic that it could be misunderstood.

ltshb (approved):
Greats thanks, just add some comments and you are ready to merge.

PR #362, opened 2023-02-27
BGDIINF_SB-2799: ADR Cache settings

ltshb:
This ADR will be discussed on 27.02.23 at 13:00 on tech talk call.

### DIFF
--- a/adr/2023_02_27_cache_settings.md
+++ b/adr/2023_02_27_cache_settings.md
@@ -1,10 +1,10 @@
 # Cache Settings
 
-> `Status:`
+> `Status: Accepted`
 
 > `Date: 2023-02-27`
 
-> `Participants: Brice Schaffner, Christoph Böcklin`
+> `Participants: Brice Schaffner, Christoph Böcklin, Jürgen Hansmann`
 
 ## Context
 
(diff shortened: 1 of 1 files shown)


adr/2023_02_27_cache_settings.md
```diff
@@ -0,0 +1,109 @@
...
+- STAC Specification extension: [Timestamps Extension Specification](https://github.com/stac-extensions/timestamps)
+  | Field  | Type   | Description   |
+  |-------------- | -------------- | -------------- |
+  | published    | string     | Date and time the corresponding data (see below) was published the first time, in UTC. |
+  | expires | string | Date and time the corresponding data (see below) expires (is not valid any longer), in UTC. |
+  | unpublished | string | Date and time the corresponding data (see below) was unpublished, in UTC. |
+  
+## Proposal #1
+
+A proposal is to add a new field in the transactional API on the create asset; `update_rate`
```
hansmannj, line 53: A very picky comment follows, sorry for that ;-)

Calling things `rate` would imply something in the form of  `something per seconds`, so e.g. one update every 60 seconds as rate would be `1/60s` so the unit would be `1/s`, not `s`.

If we provide a value in the form of "number of seconds", why don't we call it `livetime` instead of `rate`

[picky mode off ;-)]

hansmannj (commented):
Thanks for the detailed info and the options.

One thought:
What if there are assets which should be updated e.g. twice a day, always at the same time, e.g. at 12:00 and 00:00?
Should we specify this e.g. as `livetime` = 12 * 60 * 60` PLUS specifying a start time, e.g. "00:00"? So it is clear, that the asset lives 12 hours, starting every day at "00:00"?

More general, should there be an (optional?) field that specifies "t0"?

adr/2023_02_27_cache_settings.md
ltshb: @hansmannj what aboute `update_frequence` ?

hansmannj (approved):
Thanks a lot :rocket:

ltshb:
@boecklic  and @hansmannj  I just saw that the asset upload objects are not deleted in the DB once the upload is completed as I thought. So I suggest to add the `update_interval` to both models the `Asset` and `AssetUpload` so we can also track changes of this parameter. The logic would be then the same as for the multihash checksum which is present in both models.

PR #365, opened 2023-03-02
BGDIINF_SB-2799: Added dynamic cache settings

ltshb:
Now the cache setting is dynamically set based on the `update_interval`. `-1` is the default and uses the default cache settings for backward compatibility.
Related to #364 and #366 

- [ ] TODO down DB migrations for fallback in case of issues.
- [ ] Pagination issue when running locally (same with develop branch)


ltclm (approved):
Thank you @ltshb 
Did only review the trigger sql functions. it all looks good to me and it seems to be performant.

I did some tests with a copy of the current prod stac database:
```SQL
		-- item update / trigger
        EXPLAIN ANALYZE SELECT
            COALESCE(MIN(NULLIF(asset.update_interval, -1)), -1) AS min_update_interval
        FROM stac_api_asset AS asset
        WHERE asset.item_id = 208995;
		-- "Execution Time: 0.061 ms"
		
        EXPLAIN ANALYZE UPDATE stac_api_item SET
            update_interval = 999
        WHERE id = 208995;
		-- "Execution Time: 8.466 ms"
		

		-- collection update / trigger 
		-- swissimage dop 10 with 72146 items
        EXPLAIN ANALYZE SELECT
            COALESCE(MIN(NULLIF(item.update_interval, -1)), -1) AS min_update_interval

        FROM stac_api_item AS item
        WHERE item.collection_id = 9
		-- "Execution Time: 47.137 ms"
		
        begin;
		EXPLAIN ANALYZE
		UPDATE stac_api_collection SET
            update_interval = 23
        WHERE id = 72146;
		-- "Execution Time: 0.044 ms"
```

All the indexes are there and they are used by the trigger queries.
Did some tests with the swissimage collection with 72'146 items

app/middleware/cache_headers.py
```diff
@@ -44,10 +45,11 @@ def __call__(self, request):
             patch_response_headers(response, cache_timeout=10)
         elif (
             request.method in ('GET', 'HEAD') and
-            not request.path.startswith(urlparse(settings.STATIC_URL).path)
+            not request.path.startswith(urlparse(settings.STATIC_URL).path) and
+            get_max_age(response) is None  # only set if not already set by the application
```
hansmannj, line 49: So we include 40X intentionally here, right?
Because they should not only be accidentally / temporarily returned, so they can be cached. Do I understand that correctly?

app/stac_api/migrations/0024_added_update_interval_field.py
```diff
@@ -0,0 +1,132 @@
...
+                validators=[django.core.validators.MinValueValidator(-1)]
+            ),
+        ),
+        migrations.AddField(
+            model_name='assetupload',
+            name='update_interval',
+            field=models.IntegerField(
+                default=-1,
+                help_text=
+                'Interval in seconds in which the asset data is updated.-1 means that the data is not on a regular basis updated.This field is can only be set via the API.',
```
hansmannj, line 34: Small typo in the help text: This field `is can`. The `is` can be deleted here.

app/stac_api/migrations/0024_added_update_interval_field.py
```diff
@@ -0,0 +1,132 @@
...
+                validators=[django.core.validators.MinValueValidator(-1)]
+            ),
+        ),
+        migrations.AddField(
+            model_name='collection',
+            name='update_interval',
+            field=models.IntegerField(
+                default=-1,
+                help_text=
+                'Minimal update interval in seconds in which the underlying assets data are updated.',
```
hansmannj, line 44: Consistency: On asset level the text is "Interval in seconds ...", here it is "Minimal update interval". Probably might make sense to have the same description everywhere, on order to avoid confusion?

app/stac_api/migrations/0024_added_update_interval_field.py
```diff
@@ -0,0 +1,132 @@
...
+            model_name='asset',
+            trigger=pgtrigger.compiler.Trigger(
+                name='add_del_asset_item_update_interval_trigger',
+                sql=pgtrigger.compiler.UpsertTriggerSql(
+                    declare=
+                    'DECLARE asset_instance stac_api_asset%ROWTYPE; item_update_interval RECORD;',
+                    func=
+                    "\n        asset_instance = COALESCE(NEW, OLD);\n\n        -- Compute item update_interval (minimum aggregation of asset's update_interval)\n        SELECT\n            COALESCE(MIN(NULLIF(asset.update_interval, -1)), -1) AS min_update_interval\n        INTO item_update_interval\n        FROM stac_api_asset AS asset\n        WHERE asset.item_id = asset_instance.item_id;\n\n        -- Update related item update_interval variables\n        UPDATE stac_api_item SET\n            update_interval = item_update_interval.min_update_interval\n        WHERE id = asset_instance.item_id;\n\n        RAISE INFO 'item.id=% update_interval updated, due to asset.name=% updates.',\n            asset_instance.item_id, asset_instance.name;\n\n        RETURN asset_instance;\n        ",
```
hansmannj, line 70: OK, I guess this answers my question from above. On item level it seems to be the minimum value of all the asses of that collection. And on collection level the minimum of all items.

app/stac_api/migrations/0024_added_update_interval_field.py
```diff
@@ -0,0 +1,132 @@
...
+            trigger=pgtrigger.compiler.Trigger(
+                name='update_asset_item_update_interval_trigger',
+                sql=pgtrigger.compiler.UpsertTriggerSql(
+                    condition='WHEN (OLD.* IS DISTINCT FROM NEW.*)',
+                    declare=
+                    'DECLARE asset_instance stac_api_asset%ROWTYPE; item_update_interval RECORD;',
+                    func=
+                    "\n        asset_instance = COALESCE(NEW, OLD);\n\n        -- Compute item update_interval (minimum aggregation of asset's update_interval)\n        SELECT\n            COALESCE(MIN(NULLIF(asset.update_interval, -1)), -1) AS min_update_interval\n        INTO item_update_interval\n        FROM stac_api_asset AS asset\n        WHERE asset.item_id = asset_instance.item_id;\n\n        -- Update related item update_interval variables\n        UPDATE stac_api_item SET\n            update_interval = item_update_interval.min_update_interval\n        WHERE id = asset_instance.item_id;\n\n        RAISE INFO 'item.id=% update_interval updated, due to asset.name=% updates.',\n            asset_instance.item_id, asset_instance.name;\n\n        RETURN asset_instance;\n        ",
```
hansmannj, line 88: My pgtrigger/SQL foo still needs improvements, but if I understand this correctly, we make sure here, that the minimum will be calculated excluding the -1 values, right? Because otherwise of course we'd have -1 as a minimum, if that value appears somewhere. Which we would not want.

app/stac_api/models.py
```diff
@@ -593,26 +622,41 @@ class Status(models.TextChoices):
...
     etag = models.CharField(blank=False, null=False, max_length=56, default=compute_etag)
 
+    update_interval = models.IntegerField(
+        default=-1,
+        null=False,
+        blank=False,
+        validators=[MinValueValidator(-1)],
+        help_text="Interval in seconds in which the asset data is updated."
+        "-1 means that the data is not on a regular basis updated."
+        "This field is can only be set via the API."
```
hansmannj, line 632: The `is` in `is can only be set ...` can be deleted.

app/stac_api/models.py
```diff
@@ -454,6 +473,7 @@ def upload_asset_to_path_hook(instance, filename=None):
     # set the hash to the storage to use it for upload signing, this temporary attribute is
     # then used by storages.S3Storage to set the MetaData.sha256
     setattr(instance.file.storage, '_tmp_sha256', ctx.hexdigest())
+    setattr(instance.file.storage, '_tmp_update_interval', instance.update_interval)
```
hansmannj, line 476: Does the above comment need to be updated, so that `_tmp_update_interval` appears in the comment, too?

app/stac_api/models.py
```diff
@@ -301,6 +301,15 @@ class Meta:
         blank=False, null=False, editable=False, max_length=56, default=compute_etag
     )
 
+    update_interval = models.IntegerField(
+        default=-1,
+        null=False,
+        blank=False,
+        validators=[MinValueValidator(-1)],
+        help_text="Minimal update interval in seconds "
+        "in which the underlying assets data are updated."
```
hansmannj, line 310: should it be "underlying items" here? (Even if the value of the item's is that of the asset's, and hence it will be the same here.)

app/stac_api/pgtriggers.py
```diff
@@ -119,12 +127,32 @@ class UpdateCollectionSummariesTrigger(pgtrigger.Trigger):
...
+        declare = [
+            ('asset_instance', 'stac_api_asset%ROWTYPE'),
+            ('item_update_interval', 'RECORD'),
+        ]
+        func = '''
+        asset_instance = COALESCE(NEW, OLD);
+
+        -- Compute item update_interval (minimum aggregation of asset's update_interval)
+        SELECT
+            COALESCE(MIN(NULLIF(asset.update_interval, -1)), -1) AS min_update_interval
```
hansmannj, line 141: Not sure if I understand this correctly, but won't this return -1 as min, in case -1 is present? Do we want this? -1 means the data is not updated on a regular basis. Don't we want to have the minimum value of those items, that ARE updated on a regular basis?

app/stac_api/utils.py
```diff
@@ -388,3 +389,42 @@ def is_valid_b64(value):
...
+        raise ValueError('update_interval should not be less than 1h')
+
+    return -1  # means use default cache settings
+
+
+def get_s3_cache_control_value(update_interval):
+    max_age = get_dynamic_max_age_value(update_interval)
+
+    if max_age == 0:
+        return 'max-age=0, no-cache, no-store, must-revalidate, private'
```
boecklic, line 425: why is this marked as `private`?

app/stac_api/views_mixins.py
```diff
@@ -167,3 +171,28 @@ def perform_destroy(self, instance):
             raise serializers.ValidationError(
                 _(f'Deleting {instance.__class__.__name__} with {child_name} not allowed')
             ) from None
+
+
+def patch_cache_settings_by_update_interval(response, update_interval):
+    max_age = get_dynamic_max_age_value(update_interval)
```
boecklic, line 177: this can be `-1` as well right? what happens in this case?

boecklic (approved):
I can't comment in detail on the pgtriggers, rest looks good.

app/middleware/cache_headers.py
ltshb: 40x errors are cached the same way as 20x calls, 40x are error due to client wrong usage opposed to 50x which is a server error that might be temporar (e.g. 502 or 504 due to server overloaded).
Here in the line 48 and 49 we don't overwrite cache settings if the view already set it.

app/stac_api/utils.py
```diff
@@ -388,3 +389,42 @@ def is_valid_b64(value):
...
+    threshold_no_cache = 10
+    max_threshold = 60 * 60  # 1h
+    if 0 <= update_interval < threshold_no_cache:
+        return 0  # means never cache
+
+    if threshold_no_cache <= update_interval < max_threshold:
+        return int(math.log(update_interval, 10) * math.log(update_interval, 9))
+
+    if max_threshold <= update_interval:
+        raise ValueError('update_interval should not be less than 1h')
```
hansmannj, line 416: Should this not be "update_interval should not be more than 1 h" here? Or "should be less than 1 h"?

app/stac_api/utils.py
```diff
@@ -388,3 +389,42 @@ def is_valid_b64(value):
...
+    - 10 <= update_interval then log10(update_interval)*log9(update_interval)
+    Args:
+        update_interval: int
+            Value of the DB field `update_interval`
+
+    Returns:
+        int - Cache control max_age, or -1 to use default cache settings
+    '''
+    threshold_no_cache = 10
+    max_threshold = 60 * 60  # 1h
```
hansmannj, line 408: Would it make sense to have a global setting or an ENV var for the max cache duration?

app/stac_api/migrations/0024_added_update_interval_field.py
ltshb: @hansmannj `update_interval` can only be set on asset level (via the asset upload object).

On item level:
it is and will always be read only. The value is set by pgtriggers. The value is the minimal non negative aggregation of all item's assets or -1 if all assets have -1.

On collection level:
it is and will always be read only. The value is set by pgtriggers. The value is the minimal non negative aggregation of all collection's items or -1 if all items have -1.

Therefore I used different help text because this value is different on asset level as on  item and collection level. This help text is only display on the admin interface and currently the `update_interval` is read only in admin interface.

app/stac_api/migrations/0024_added_update_interval_field.py
ltshb: Correct

app/stac_api/models.py
ltshb: Nope the `update_interval` is always for the asset data/file update interval. On item/collection it is the minimal value of all their assets but still refer to the asset data update interval.

app/stac_api/pgtriggers.py
ltshb: `MIN(NULLIF(asset.update_interval, -1))` => returns either a positive value (including `0`) or `null`. It returns `null` if all `update_interval` are `-1`
`COALESCE(.., -1)` return `-1` if the first part is `null`.
We want the minimal positive value (including 0) or -1 if there is no minimal value.

app/tests/base_test.py
```diff
@@ -69,7 +69,27 @@ def get_error_msg(code):
...
+                Check that the header is `max-age=0, no-cache, no-store, must-revalidate, private`
+        '''
+        self.assertTrue(response.has_header('Cache-Control'), msg="Cache-Control header missing")
+        if no_cache:
+            self.assertEqual(
+                response['Cache-Control'],
+                'max-age=0, no-cache, no-store, must-revalidate, private'
+            )
+        elif max_age is not None:
+            self.assertEqual(response['Cache-Control'], f'max-age={max_age}, public')
```
hansmannj, line 90: Probably add a check, that `max_age` is less-equal the max allowed cache duration (1h)?

app/stac_api/utils.py
ltshb: To avoid browser cache as well. This string is the same that django helper `add_never_cache_headers` generates.

app/stac_api/views_mixins.py
ltshb: @boecklic In this case there is no cache-control header patching here and it is done via the cache middleware, see https://github.com/geoadmin/service-stac/blob/3b7c4d1ed106679a129a840433ba961246fca87d/app/middleware/cache_headers.py#L49

app/stac_api/utils.py
ltshb: good catch

app/stac_api/utils.py
ltshb: I don't think, for a bigger value we need to re-think the logic with the logarithm value anyway and will very likely do some changes in the code anyway.

app/tests/base_test.py
ltshb: @hansmannj See comment above, the maximum value or 1h is just because the current algorithm is not optimal for bigger value, but in theory and in future we should not have any limitation. Therefore I don't it worth the effort to add a check here need to be modified/removed in future.

app/stac_api/pgtriggers.py
hansmannj: OK, thanks, now I got it :wink:

app/tests/test_generic_api.py
```diff
@@ -452,6 +454,11 @@ def test_get_cache_header(self):
                                        timedelta(seconds=3600).total_seconds(),
                                        delta=2)
 
+    def test_get_asset_object_cache_header(self):
+        key = get_asset_path(self.item.model, self.asset["name"])
+        self.assertS3ObjectExists(key)
+        self.assertS3ObjectCacheControl(key, max_age=7200)
```
hansmannj, line 460: Not sure I understand, why max_age is 7200 here..

app/stac_api/utils.py
hansmannj: Can live with the way it currently is. I just thought, that if we changed this value here, probably other places will be forgotten to be changed, e.g. the hardcoded value of 1h in the error message in line 416. So having it as a global setting, we would only need to change it in one place.

app/tests/base_test.py
hansmannj: See my comment above. But we can leave it as it is.

hansmannj (approved):
Looks good to me, thanks for the effort :rocket:

app/tests/test_generic_api.py
ltshb: Default cache settings from the settings.

PR #366, opened 2023-03-02
BGDIINF_SB-2799: Updated ADR for cache settings

ltshb:
The linear approach discussed originally and later updated to use 10% between 10s and 60s
and then 1% above 60s is not the best.

Here is another approach using logarithmic calculation.


hansmannj (approved):
Good idea :+1:

PR #368, opened 2023-03-06
Fixed gunicorn tmp dir worker

ltshb:
The gunicorn temporary dir for worker should not be created and set by the
docker image but left to the deployment (aka k8s) which needs to ensure to have
it on tmpfs.

See also https://github.com/geoadmin/infra-kubernetes/pull/315


hansmannj (approved):
:+1: Thanks!

PR #369, opened 2023-03-06
BGDIINF_SB-2799: Fixed static files serving and clean up environment variables

ltshb:
The latest develop.latest image could only be used with the DEBUG=1 otherwise
the static files (spec, admin page templates,...) could not be served. This
is probably due to the environment variable being added to docker and the
collect static was done with DEBUG=1. It is not clear why this work before on
int and prod staging.

Anyway now the .env file is not imported in docker image to avoid any mixing of
environment variables. The .env.local file is now only used for local development.

Updated also the readme. Also make sure to use the same port if using make serve
or make dockerrun.

See also commits for some other minor improvements

---
Deploying `v1.16.0-beta.7` on INT had the issue that the static assets were missing. On the pods the folder /opt/service-stac/var/www is empty !


boecklic (approved):

hansmannj:
Sorry, I was too late. Thanks :+1:

PR #370, opened 2023-03-08
Rename the migration file for django upgrade to 4.0.0

ltshb:
This allow to better understand the migrations files.


ltshb:
Best to avoid to rename productive migrations.

PR #373, opened 2023-03-09
BGDIINF_SB-2799: Fixed performance issue with the search endpoint due to update_interval

ltshb:
In the search endpoint an aggregation over the whole queryset was done for
the update interval. This can potentially be very slow as a GET /search return
a queryset with all items in the DB ! Also the aggregation was done even
in a POST /search where we don't need to add Cache-Control header !

Now the cache-control header is only in GET /search geadded and the
update_interval aggregation is also only done for GET request.

The aggregation is done on the paginated queryset, meaning up to maximum of
100 items. We need to set the cache control header for the response of the
request and not based on all matching items.

Has we can see below the performance tests are passing again:
BASE:
```text
environment: alb-int
service-version: ltshb.v1.16.0-beta.10-1-g1d1e5e2
test-version: f34b8db
pods: 3
```
TARGET:
```text
ENV=alb-int
PODs=3
service-version=v1.6.0-beta.4
```
image

### DIFF
--- a/app/stac_api/views.py
+++ b/app/stac_api/views.py
@@ -215,14 +215,20 @@ class SearchList(generics.GenericAPIView, mixins.ListModelMixin):
         validate_search_request.validate(request)  # validate the search request
         queryset = self.filter_queryset(self.get_queryset())
 
-        update_interval = self.get_min_update_interval(queryset)
-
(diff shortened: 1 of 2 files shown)


boecklic (approved):
:1st_place_medal: great!

PR #374, opened 2023-03-09
BGDIINF_SB-2813: Fixed collection pagination due to missing title

ltshb:
Because the collections are ordered by title (was required for the STAC Browser)
the title is used by the cursor paginator. If the title is missing (null or empty
string) it breaks the pagination !

Therefore made the title mandatory. Up to now all production collection had a
title. This commit will add a default title as 'unknown' to all collection
without title, but this has no impact on productive data as they all do have
a title.

The title has to be unique as well if we use it as ordering otherwise the pagination won't work !


hansmannj (approved):
Merci :+1:

boecklic (commented):
why do we have such massive changes in the spec? I assume it's just reordering/reformatting the same content but it's hard to review...

ltshb:
@boecklic I think the issue is more that we the latest changes of spec done by #371

ltshb:
Convert it to draft to avoid intempestive merge. This can only be merged once the productive data has been update (duplicate title on prod) otherwise we won't be able to deploy !

ltshb:
It has been decided to revert to collection ID sorting, see comment in ticket. Therefore this PR is closed and replaced by #383

PR #375, opened 2023-03-13
BGDIINF_SB-2806: Added more checks to be performed by CI

ltshb:
Added more check in makefile to be performed by the CI. Also remove the CI
docker compose file, now the CI will use the conventional `make test`.

Require https://github.com/geoadmin/infra-terraform-bgdi-builder/pull/102


rebert (approved):
Thanks @ltshb

PR #376, opened 2023-03-14
BGDIINF_SB-2827: Removed auto abort asset upload on duplicate upload to fix integrity crash

ltshb:
Only one asset upload is allowed at a time, before it used to automatically
abort ongoing upload when creating a new upload. The implementation had an
issue when multiple create upload happened on parallel, it crashed the service
due to race condition.

Making parallel or multiple create upload doesn't make sense as only one call
will be valid and the user will not know which one. E.g. if a user do two
requests in a row, both would have return 201 but only one of it would have
been valid and we could not ensure which one, could be the first or the second !

To avoid such random cases, we removed the auto abort mechanism. This means
that if multiple requests are sent at a time it is guarantied that only one
would be successful with 201 and valid. The others will returns 400 Bad Request
- Asset upload already in progress.

Improved a bit the test code.

### DIFF
--- a/app/stac_api/s3_multipart_upload.py
+++ b/app/stac_api/s3_multipart_upload.py
@@ -29,10 +29,6 @@ class MultipartUpload:
     def __init__(self):
         self.s3 = get_s3_client()
 
-    @property
-    def client(self):
-        return self.s3
-
     def list_multipart_uploads(self, key=None, limit=100, start=None):
         '''List all in progress multipart uploads
 
--- a/app/stac_api/s3_multipart_upload.py
+++ b/app/stac_api/s3_multipart_upload.py
@@ -29,6 +29,10 @@ class MultipartUpload:
     def __init__(self):
         self.s3 = get_s3_client()
 
+    @property
+    def client(self):
+        return self.s3
+
     def list_multipart_uploads(self, key=None, limit=100, start=None):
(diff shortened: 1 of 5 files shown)


app/stac_api/s3_multipart_upload.py
```diff
@@ -29,6 +29,10 @@ class MultipartUpload:
     def __init__(self):
         self.s3 = get_s3_client()
 
+    @property
+    def client(self):
+        return self.s3
+
```
boecklic, line 35: is this property used somewhere? If not I would remove it

app/tests/base_test.py
```diff
@@ -54,11 +54,12 @@ def get_error_msg(code):
                 return msg
             return 'Wrong status code'
 
-        self.assertEqual(code, response.status_code, msg=get_error_msg(response.status_code))
-        if code in [412, 304]:
+        code = code if isinstance(code, list) else [code]
+        self.assertIn(response.status_code, code, msg=get_error_msg(response.status_code))
```
boecklic, line 58: `codes = ` (plural) would be slightly more descriptive and indicate that it's a list

boecklic (approved):

app/tests/base_test.py
ltshb: Done

PR #377, opened 2023-03-14
BGDIINF_SB-2827: Fixed asset upload integrity crash

ltshb:
Allow to have at least two consecutive/parallel call to create asset upload.

This solution works but one of the call will be invalid although it returns
HTTP 201 ! This is not very user friendly.


ltshb:
We can guarantee the service to not crash in case of parallel call to create asset upload, but only one call can be valid ! This mean that a client might get an HTTP 201 with an invalid upload. Not very useful, therefore this is closed and #376 is implemented instead.

PR #378, opened 2023-03-15
BGDIINF_SB-2799: Fixed compression issue with Cloudfront

ltshb:
Cloudfront requires at least a TTL of 1s in order to compress data.
So only set no cache when the update_interval is 0.

### DIFF
--- a/app/stac_api/utils.py
+++ b/app/stac_api/utils.py
@@ -395,7 +395,8 @@ def get_dynamic_max_age_value(update_interval):
     '''Get the max_age value for dynamic cache based on `update_interval` DB field
 
     -       update_interval < 0  then use default cache settings
-    -  0 <= update_interval < 10 then no cache
+    -  0 == update_interval then no cache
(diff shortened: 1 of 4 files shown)


app/stac_api/utils.py
```diff
@@ -395,7 +395,8 @@ def get_dynamic_max_age_value(update_interval):
     '''Get the max_age value for dynamic cache based on `update_interval` DB field
 
     -       update_interval < 0  then use default cache settings
-    -  0 <= update_interval < 10 then no cache
+    -  0 == update_interval then no cache
+    -  0 < update_interval < 10 then cache of 1s
```
boecklic, line 399: maybe a comment here about the relation to compression behaviour of CF could be useful to understand why this was done that way in the future.

boecklic (approved):

ltshb:
This was not the issue, the problem was that CF cannot compress data bigger equal than 10MB !

PR #380, opened 2023-03-16
BGDIINF_SB-2824: Add new Mime Type

ltkum:
Issue : We needed to accept a new Mime Type for some files

Fix : We added the Mime Type to the authorized Media Types in the validator


ltshb:
Wrong ticket number in commit (should be BGDIINF_SB-2824) in branch you leave it, the branch is anyway deleted afterwards.

app/stac_api/validators.py
```diff
@@ -49,6 +49,7 @@
     MediaType('application/vnd.laszip', 'Zipped LIDAR', ['.laz', '.zip']),
     MediaType('application/x.shapefile+zip', 'Zipped Shapefile', ['.zip']),
     MediaType('application/x.filegdb+zip', 'Zipped File Geodatabase', ['.zip']),
+    MediaType('application/x.filegdbp+zip', 'Zipped File Geodatabase', ['.zip']),
```
ltshb, line 52: I would use another description, I've no idea what is the difference between x.filegdb+zip and x.filegdbp+zip, but maybe ask @LucaSchneller or @phanuel for better description.

phanuel: Zipped File Geodatabase (ArcGIS Pro)

phanuel (approved):

PR #382, opened 2023-03-22
BGDIINF_SB-2835: Added content_encoding param to API Asset Upload

ltshb:
NOTE currently this parameter is not supported by the admin interface.


rebert (approved):
At least I haven't noticed anything negative. Thanks for adding the encoding param with coresponding unittests

app/tests/test_validators.py
```diff
@@ -58,6 +59,30 @@ def test_validate_function_invalid_datetime_string(self):
...
+    def test_validate_valid_content_encoding(self):
+        for value in [
+            'gzip',
+            'gzip, br',
+            'br, gzip',
+            'compress',
+            'compress, br, gzip, deflate',
+            'gzip,br',
+            'gzip,     br'
+        ]:
```
boecklic, line 79: shouldn't only a single compression algorithm be valid at once, i.e. 
```
for value in ['gzip', 'br', 'compress', 'deflate']:
```
?

app/stac_api/models.py
```diff
@@ -634,6 +634,8 @@ class Status(models.TextChoices):
         "This field can only be set via the API."
     )
 
+    content_encoding = models.CharField(blank=True, null=False, max_length=32, default='')
```
boecklic, line 637: we could provide the valid choices here as well, this would make it easier to integrate into the admin afterwards

app/stac_api/validators.py
```diff
@@ -526,3 +526,37 @@ def validate_md5_parts(md5_parts, number_parts):
...
+    Args:
+        value: str
+            value to validate
+
+     Raises:
+        ValidationError in case of invalid content_encoding value
+    '''
+    # See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding for
+    # supported encoding
+    supported_encoding = set(['gzip', 'br', 'deflate', 'compress'])
```
boecklic, line 543: since `deflate` and `compress` are hardly used and e.g. not supported by all browsers: should we restrict it to `gzip` and `br`?

boecklic (approved):
Thanks!

app/stac_api/validators.py
ltshb: As discussed I'll be more restrictive and only allow gzip or br

ltshb: Done

PR #388, opened 2023-04-05
BGDIINF_SB-2852: Fixed asset file content-type for asset uploaded via admin page

ltshb:
When uploading an asset via the admin page, the content-type of the asset object
was guessed and not taken from the asset metadata. For some case it works
but for other the guess type was wrong.

Due to the spec changes require first #389


hansmannj (approved):
Thanks, LGTM :+1:

PR #389, opened 2023-04-05
Fixed build spec dependencies issues.

ltshb:
During the PR #371, the last commit of the PR; ad3f1e1da79659d9fb8aee9cf8382f6be61dae7e
removed some commented part, but the spec has not been built.

Because the build output is under source control and that the makefile used
the file as dependency, once this PR has been merged, running `make build-specs`
did nothing. However if you forced a build by removing the static
static/spec/v0.9/openapi.yaml and static/spec/v0.9/openapitransactional.yaml
files, it trigger a build with the modification of the last commit of #371.

Although on this case the spec did not really changed as it was just a code comment
clean up this can be dangerous as if user forget to build the spec and it gets
merged, the CI won't even notice it (the `make build-specs` of the CI would
do nothing). And we would notice only the next time the spec is modified and
built.

To avoid such circumstance, I made the build targets as PHONY and the build
will be always triggered even if nothing has changed.

NOTE: Another solution would have been to remove the static files from source
control and rely on the built tool.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -210,13 +210,16 @@ lint-specs:
 
 
 .PHONY: ci-build-check-specs
-ci-build-check-specs: build-specs
-	@if [[ -n `git status --porcelain --untracked-files=no` ]]; then \
-	 	>&2 echo "ERROR: the following files changed after building the spec"; \
-	 	>&2 echo "'git status --porcelain' reported changes in those files after a 'build-specs' :"; \
-		>&2 git status --porcelain --untracked-files=no; \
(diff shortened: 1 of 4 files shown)


hansmannj (approved):
LGTM, merci! :+1:

ltshb:
Because the `yq` merged output differ on the CI from the output generated from my ubuntu 20.04 laptop I disabled the build spec check in CI. We need to talk in a tech talk we we want to do.

PR #395, opened 2023-06-21
BGDIINF_SB-2924: Added docu about asset upload error handling.

ltshb:


ltshb:
@hansmannj I used your png from slack, but it has poor pixel quality, I don't know what tool you used but would be nice if you can create an `svg` with transparent background file. I would add the 3 steps number in the diagram as well. Feel free to correct my poor english and change the diagram file directly in this PR.

hansmannj (approved):
Thanks! LGTM :+1:
Hope the figure is OK now.
Tried with transparent background in SVG format.
SVG would be nicer, not so pixely. But when using dark-mode on github, the lines become almost invisible.

boecklic (approved):
Thanks a lot! 
@hansmannj : what tool did you use to create the graphics? We should probably agree on one tool to use for diagrams etc., and commit the raw file to the repo as well. In the worst case the next one needing to edit the graphics has to start over again. Not urgent but a thing to consider when rethinking documentation in general.

ltshb:
@hansmannj on the graphic there is a line from step 3 to step 1. in case of 200 which should be removed. Would be good to put the raw graphic as mentioned by Chris here as well.

hansmannj:
@ltshb @boecklic: managed to have a .svg file with white background, which is much nicer. Also added raw format file and a link to the Dia tool in the .md file

doc/stac-upload-error-handling.md
```diff
@@ -0,0 +1,84 @@
...
+
+        Abort upload using the current `upload_id` and contact service administrator.
+
+        ```text
+        POST https://data.geo.admin.ch/api/stac/v0.9/collections/{collection}/items/{item}/assets/{asset}/uploads/{upload_id}/abort
+        ```
+
+    - `502 Bad Gateway`, `503 Service Unavailable`, `504 Gateway Timeout` => Retry
+
+        Retry step 2. with a short wait time (min 100ms).
```
hansmannj, line 63: In the figure I used 200 ms. We should either adapt the text or the figure.
I'd opt for the longer period

PR #396, opened 2023-06-21
BGDIINF_SB-2924: Fixed duplicate asset upload complete issue

ltshb:
When retrying a complete operation, the service crashed with a KeyError. The
retry should be allow and return a proper 400 Bad Request. A user might want
to retry a complete operation if it received a 50x message not knowing
the state of the complete.


hansmannj (approved):
:+1: Thanks! LGTM

PR #415, opened 2024-05-30
PB-512: Update Django 5

benschs:
Update Django version from 4 to 5.
Update python version from 3.9 to 3.12, as Django 5 requires at least python 3.10.

Changes for Django 5:
Remove deprecated features from code, specifically `django.contrib.gis.admin.GeoModelAdmin` and `USE_L10N` setting.
See https://docs.djangoproject.com/en/5.0/releases/4.0/#deprecated-features-4-0 for all deprecation details.

Changes for python 3.12:
Remove deprecated `distutils`. Ignore some linter issues.


app/stac_api/admin.py
```diff
@@ -159,7 +159,7 @@ def clean_text_geometry(self):
 
 
 @admin.register(Item)
-class ItemAdmin(admin.GeoModelAdmin):
+class ItemAdmin(admin.ModelAdmin):
```
schtibe, line 162: I'm just wondering why this is necessary and make sure that it's an intended change and not an accident ;)

boecklic: same question: maybe a short comment or hint to changelog would be helpful for all changes of this kind

boecklic (changes requested):

app/stac_api/admin.py
benschs: Thanks for the input, yes it was intended, as `GeoModelAdmin` is deprecated in version 5 and is replaceable with `ModelAdmin`.
I added mode information to the PR description regarding deprecation changes.

schtibe (approved):
Looks good! 👍

app/stac_api/management/commands/list_asset_uploads.py
```diff
@@ -63,7 +63,7 @@ def are_uploads_equal(s3_upload, db_upload):
                 return False
 
             # Add all db uploads
-            for db_upload in db_uploads_qs:
+            for db_upload in db_uploads_qs:  # pylint: disable=possibly-used-before-assignment
```
ltshb, line 66: I would not disable this pylint but fix the code, you can set `db_uploads_qs` to an empty list at the top of the function instead.

app/stac_api/management/commands/list_asset_uploads.py
```diff
@@ -96,6 +96,7 @@ def are_uploads_equal(s3_upload, db_upload):
             only_s3_uploads = s3_uploads
 
         print(
+            # pylint: disable=possibly-used-before-assignment
```
ltshb, line 99: Dito this is not safe to disable this pylint might be a better idea to fix the code

app/stac_api/models.py
```diff
@@ -525,7 +525,7 @@ def filename(self):
     description = models.TextField(blank=True, null=True, default=None)
     eo_gsd = models.FloatField(null=True, blank=True, validators=[validate_eo_gsd])
 
-    class Language(models.TextChoices):
+    class Language(models.TextChoices):  # pylint: disable=too-many-ancestors
```
ltshb, line 528: Would be better to update the pylint configuration to allow more ancestors instead of repeating this pylint disable statement

app/tests/test_asset_upload_model.py
```diff
@@ -78,7 +77,7 @@ def test_create_asset_upload_default(self):
         self.assertEqual(asset_upload.ended, None, msg="Wrong default value")
         self.assertAlmostEqual(
             utc_aware(datetime.utcnow()).timestamp(),
-            asset_upload.created.timestamp(),
+            asset_upload.created.timestamp(),  # pylint: disable=no-member
```
ltshb, line 80: Did you verify that `asset_upload.created` really have a `timestamp()` method ?

ltshb:
@benschs do you know if django 5 solve the logging issue where we don't have the request context like in flask ?

app/tests/test_asset_upload_model.py
benschs: Yes, timestamp seems to still exist. Testcase also passes without any issues.

benschs:
> @benschs do you know if django 5 solve the logging issue where we don't have the request context like in flask ?

As far as i can tell there is nothing like this in the Release notes.

ltshb (approved):

boecklic (approved):
:+1:

PR #416, opened 2024-06-11
PB-610: Store different geometries

benschs:
Item geometry only supported polygons, but according to STAC spec should support any geometry type.

Use Django GeometryField type for item geometry. Also use GeometryField for collection extent, as it no longer need be a polygon, bounding box stays unchanged.


app/stac_api/models.py
```diff
@@ -269,7 +269,7 @@ class Meta:
     # only for the initial value.
     updated = models.DateTimeField(auto_now_add=True)
     description = models.TextField()
-    extent_geometry = models.PolygonField(
+    extent_geometry = models.GeometryField(
```
ltshb, line 272: I'm  not sure this is a good idea here, the extent is anyway always a geometry like a bbox, having 4 points/corner. @boecklic or did I missunderstand the extent definition and we could have a single point as extent ?

benschs: The collection extent is computed (by DB trigger) from all the items. In edge cases, this is not a polygon. AFAICS this is only stored in the database, the API of a collection only reveals the bbox, which is still calculated correctly.

PR #417, opened 2024-06-18
PB-612: stac v1 setup

benschs:
Setup service to handle both versions in the request path. For now both
versions have the same functionality.

- Add path namespaces per version and add all stac paths to each namespace.
- Extend get URL to handle both versions when serializing links.
- Remove version information from config.
- Split spec for both versions, makefile generates spec for v1, v0.9 must be
maintained manually.
- Duplicate unit tests for each version.


boecklic (approved):
:rocket: 
(no comments anymore, all discussion have been resolved previously)

PR #419, opened 2024-06-19
PB-613: adapt spec for item geometry

benschs:
Adapt spec of item geometry to be one of point, lineString, polygon, multiPoint, multiLineString or multiPolygon.
This change is only in spec of v1, although the implementation is actually available in both v1 and v0.9.

### DIFF
--- a/spec/components/schemas.yaml
+++ b/spec/components/schemas.yaml
@@ -841,13 +841,13 @@ components:
       example: smr200-200-4-2019
       type: string
     itemGeometry:
-      # oneOf:
-        # - $ref: "#/components/schemas/geoJsonPoint"
-        # - $ref: "#/components/schemas/geoJsonLineString"
-      $ref: "#/components/schemas/geoJsonPolygon"
-        # - $ref: "#/components/schemas/geoJsonMultiPoint"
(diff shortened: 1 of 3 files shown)


spec/components/schemas.yaml
```diff
@@ -841,13 +841,13 @@ components:
...
-        # - $ref: "#/components/schemas/geoJsonMultiPoint"
-        # - $ref: "#/components/schemas/geoJsonMultiLineString"
-        # - $ref: "#/components/schemas/geoJsonMultiPolygon"
+      oneOf:
+        - $ref: "#/components/schemas/geoJsonPoint"
+        - $ref: "#/components/schemas/geoJsonLineString"
+        - $ref: "#/components/schemas/geoJsonPolygon"
+        - $ref: "#/components/schemas/geoJsonMultiPoint"
+        - $ref: "#/components/schemas/geoJsonMultiLineString"
+        - $ref: "#/components/schemas/geoJsonMultiPolygon"
```
boecklic, line 850: :+1: actually we can extend it to any geoJson geometry, makes sense

boecklic (approved):

PR #420, opened 2024-06-19
Add missing dockerlogin build target.

adk-swisstopo:
The "dockerlogin" build target is documented in the help but not defined. This adds the definition based on how it exists in service-print. This also abstracts the name of the Docker registry into a variable.


adk-swisstopo:
Actually, dockerlogin is in the Makefile in the parent directory. I don't like how this works but duplicating it in this file isn't going in the right direction either. I might come up with something else later.

PR #421, opened 2024-06-19
Clean up how we build the OpenAPI spec.

adk-swisstopo:
This should make it easier to inject things we need to use AWS API Gateway. It should also be a bit more robust and easier to use for external developers.

### DIFF
--- a/spec/Makefile
+++ b/spec/Makefile
@@ -28,7 +28,6 @@ help:
 	@echo
 	@echo "Possible targets:"
 	@echo -e " \033[1mLOCAL DOCKER TARGETS\033[0m "
-	@echo "- dockerlogin              Login to the AWS ECR registery for pulling/pushing docker images"
 	@echo -e " \033[1mTREAT THE SPECS TARGETS\033[0m "
 	@echo "- build-specs              Build the specs"
 	@echo "- lint-specs               lint the specs"
--- a/spec/Makefile
+++ b/spec/Makefile
@@ -1,4 +1,4 @@
-SHELL = /bin/bash
+SHELL = /bin/bash -o pipefail
 .DEFAULT_GOAL := help
 
 # List of files that should be merged.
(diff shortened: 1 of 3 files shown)


adk-swisstopo:
I think the missing tag descriptions are because they are missing from https://github.com/geoadmin/service-stac/blob/develop/spec/transaction/tags.yaml
And the missing "Swissregio" value might be because this "title" field is duplicated: https://github.com/geoadmin/service-stac/blob/develop/spec/components/schemas.yaml#L1433
Unclear whether we can make yq 4 behave the same than 3 or if it's preferable to "fix" the source.

adk-swisstopo:
Reading https://yaml.org/spec/1.2.2/#mapping I believe the following is a bug:
https://github.com/geoadmin/service-stac/blob/27eef5cbd44a827f017a68d293aa1694a2b37db6/spec/components/schemas.yaml#L1433
```
      example:
        title:
          eq: "Swissregio"
        title:
          contains: "Swiss"
```
It defines a "Mapping" with two keys "title". Within a mapping, each key needs to be unique. So, yq 4 keeping just one of the keys is entirely reasonable.

spec/Makefile
```diff
@@ -28,33 +31,38 @@ help:
...
-
+	$(OPENAPI_VALIDATOR) -e $(OPENAPI)
+	$(OPENAPI_VALIDATOR) -e $(OPENAPI_TRANSACTIONAL)
+
+# This can be useful to distinguish style and semantic differences when we
+# update the tools. For example if you update yq and see lot of whitespace
+# changes, you can "make static/spec/v1/openapi.props" and diff that to check
+# for change in the properties themselves.
+%.props: %.yaml
+	$(YQ) -P 'sort_keys(..)' -o=props $(^) > $(@)
```
boecklic, line 65: :+1: very useful!

boecklic (approved):
Thanks for the cleanup and improvement!

PR #423, opened 2024-06-20
PB-617 Introduce parallel bucket setup

schtibe:
Dual bucket setup

These changes allow for having two buckets in service-stac. Some of the uploads will go to the "managed" bucket based on the file patterns.
The changes concern the multipart upload. Based on the collection, different presigned urls will be generated from the buckets.

Other things included in this PR:
* Fixed a deprecation warning for the minio container
* Fixed the pylint configuration 
* Removed various code that I think is not needed anymore (it's not called anywhere in the code)
* Switch to docker compose plugin (instead of docker-compose)


app/config/settings_dev.py
```diff
@@ -53,3 +53,6 @@
     )
 
 SHELL_PLUS_POST_IMPORTS = ['from tests.data_factory import Factory']
+
+# Regex patterns of collections that should go to the managed bucket
+MANAGED_BUCKET_COLLECTION_PATTERNS = [r"^ch\.meteoschweiz\.ogd.*"]
```
schtibe, line 58: I'm unsure if we should have these in the environment. Or somewhere else.

app/config/settings_prod.py
```diff
@@ -295,3 +314,6 @@ def get_logging_config():
     'STAC_BROWSER_HOST', None
 )  # if None, the host is taken from the request url
 STAC_BROWSER_BASE_PATH = os.getenv('STAC_BROWSER_BASE_PATH', 'browser/index.html')
+
+# Regex patterns of collections that should go to the managed bucket
+MANAGED_BUCKET_COLLECTION_PATTERNS = [r"^ch\.meteoschweiz\.ogd.*"]
```
schtibe, line 319: See comment above for `settings_dev.py`

app/stac_api/validators_serializer.py
```diff
@@ -57,152 +57,6 @@ def validate_json_payload(serializer):
         raise serializers.ValidationError(code='payload', detail=errors)
 
 
-def validate_asset_href_path(item, asset_name, path):
```
schtibe, line 60: I noticed that these weren't called anywhere in the code (except from the tests). I guess uncalled code can be removed then?!

app/tests/utils.py
```diff
@@ -173,35 +181,6 @@ def upload_file_on_s3(file_path, file, params=None):
     obj.upload_fileobj(file, ExtraArgs=extra_args)
 
 
-def mock_requests_asset_file(mocker, asset, **kwargs):
```
schtibe, line 176: This is never called.

app/tests/test_multiple_bucket_setup.py
```diff
@@ -0,0 +1,170 @@
+from unittest.mock import patch
```
schtibe, line 1: I intentionally put the tests here instead of in one of the subfolder, as they don't really interact with either API.

app/config/settings_dev.py
boecklic: Environment makes sense to me, this way we can just change config without creating a new image

app/config/settings_prod.py
```diff
@@ -195,22 +195,41 @@
...
+    AWS_MANAGED = {
+        # STORAGE_BUCKET_NAME is the only required additional var
+        # The others are inferred by the ones above if they're not specified
+        # So if we have two buckets on the same server with the same access key, we needn't
+        # specify all the variables. But this still leaves the possibility to
+        # have it in a somewhere entirely different location
+        "STORAGE_BUCKET_NAME": os.environ['AWS_STORAGE_BUCKET_NAME_MANAGED'],
+        "ACCESS_KEY_ID": os.environ.get('AWS_ACCESS_KEY_ID_MANAGED', AWS_LEGACY['ACCESS_KEY_ID']),
+        "SECRET_ACCESS_KEY":
+            os.environ.get('AWS_SECRET_ACCESS_KEY_MANAGED', AWS_LEGACY['SECRET_ACCESS_KEY']),
```
boecklic, line 225: the goal would be to control access for the `*managed` bucket with IAM policies instead of technical users and access keys. We'll need to see how we can make that work for minio though.

app/stac_api/utils.py
```diff
@@ -109,27 +113,34 @@ def get_s3_resource():
...
+    client = boto3.client(
         's3',
-        endpoint_url=settings.AWS_S3_ENDPOINT_URL,
-        region_name=settings.AWS_S3_REGION_NAME,
-        config=Config(signature_version='s3v4'),
+        endpoint_url=conf['S3_ENDPOINT_URL'],
+        region_name=conf['S3_REGION_NAME'],
+        config=Config(signature_version=conf['S3_SIGNATURE_VERSION']),
+        aws_access_key_id=conf['ACCESS_KEY_ID'],
+        aws_secret_access_key=conf['SECRET_ACCESS_KEY']
```
boecklic, line 139: see comment above regarding access_keys

app/stac_api/utils.py
```diff
@@ -444,3 +455,27 @@ def get_s3_cache_control_value(update_interval):
...
+
+    return AVAILABLE_S3_BUCKETS.LEGACY
+
+
+def get_aws_settings(s3_bucket: AVAILABLE_S3_BUCKETS = AVAILABLE_S3_BUCKETS.LEGACY):
+
+    if s3_bucket == AVAILABLE_S3_BUCKETS.LEGACY:
+        return settings.AWS_LEGACY
+
+    return settings.AWS_MANAGED
```
boecklic, line 481: we could also just use a dict with bucket-name as keys in the settings, then we wouldn't need an additional function. what do you think?

app/stac_api/validators_serializer.py
boecklic: Surprising. But yes, if not used anywhere move it to the tests. 
File name suggests it is(was?) used in serializers at some point, but just to be sure: guess you checked the admin as well right?
Just one thing: currently we derive the asset path from collection_name/item_name/asset_filename. With meteoschweiz we'll have the case that we'll have to accept asset paths that are not on our platform. So this validation function might come handy then. Maybe it's an idea to mark it for removal but leave it for the moment. Or we re/move it now and add checks again once we know what exactly we have to check

app/config/settings_prod.py
boecklic, line 319: will be `ch.meteoschweiz.ogd-*`

app/tests/test_multiple_bucket_setup.py
```diff
@@ -0,0 +1,170 @@
...
+        patterns = [r'ch\.meteoschweiz\.ogd.*', r'ch\.swisstopo\.bgdi.*']
+        with self.settings(MANAGED_BUCKET_COLLECTION_PATTERNS=patterns):
+            bucket_name = select_s3_bucket(collection_name)
+
+        self.assertEqual(bucket_name, expected_bucket)
+
+    @parameterized.expand([
+        'ch.meteoschweiz.ogd',
+        'ch.meteoschweiz.ogd.heatwave',
+        'ch.meteoschweiz.ogd.snowstorms',
```
boecklic, line 162: see comment above: `ch.meteoschweiz.ogd-*`

app/tests/tests_10/test_validators.py
```diff
@@ -1,55 +1,15 @@
...
-            serializers.ValidationError,
-            msg="Invalid Asset href path did not raises serializers.ValidationError"
-        ):
-            validate_asset_href_path(item, 'asset-test', 'asset-test')
-            validate_asset_href_path(item, 'asset-test', 'item-test/asset-test')
-            validate_asset_href_path(item, 'asset-test', 'collection-test/item-test/asset-test')
-            validate_asset_href_path(
-                item, 'asset-test', '/service-stac-local/collection-test/item-test/asset-test'
-            )
-
```
boecklic, line 52: Hmm, are you sure this check aren't useful at all? (together with the corresponding validate functions)

boecklic (commented):
Nice!

app/stac_api/validators_serializer.py
schtibe: No it didn't occur to me to check the admin. Also it could be that it somehow is used implicitly with getattr or something. Will do further tests!

app/stac_api/utils.py
schtibe: Yes good idea. Will change that. This would also be more elegant in the future when we potentially remove the old bucket.

app/tests/tests_10/test_validators.py
schtibe: I'll check again. I had the impression that there were functions that were only called in the tests, but there might be some other magic going on, see comment above.

app/tests/test_multiple_bucket_setup.py
schtibe: This test becomes obsolete when we have the patterns in the environment.

app/stac_api/validators_serializer.py
schtibe: Maybe @ltshb knows more. He seems to have made this.

app/tests/tests_10/test_validators.py
schtibe: cc @ltshb

app/config/settings_prod.py
```diff
@@ -192,40 +195,76 @@
...
+        'managed': {
+            # STORAGE_BUCKET_NAME is the only required additional var
+            # The others are inferred by the ones above if they're not specified
+            # So if we have two buckets on the same server with the same access key, we needn't
+            # specify all the variables. But this still leaves the possibility to
+            # have it in a somewhere entirely different location
+            # This probably won't be for long, as we'll need some other means to access
+            # the new bucket than the access_key
+            "STORAGE_BUCKET_NAME": env('AWS_STORAGE_BUCKET_NAME_MANAGED'),
+            "ROLE_ARN": env('AWS_ROLE_ARN_MANAGED', default=None),
```
boecklic, line 242: see pm

app/stac_api/utils.py
```diff
@@ -109,26 +113,65 @@ def get_s3_resource():
...
+        # From the response that contains the assumed role, get the temporary
+        # credentials that can be used to make subsequent API calls
+        credentials = assumed_role_object['Credentials']
+
+        client = boto3.client('s3',)
+        client_access_kwargs = {
+            "aws_access_key_id": credentials['AccessKeyId'],
+            "aws_secret_access_key": credentials['SecretAccessKey'],
+            "aws_session_token": credentials['SessionToken']
+        }
```
boecklic, line 160: see pm

app/config/settings_prod.py
```diff
@@ -192,40 +195,76 @@
...
+            "S3_REGION_NAME":
+                env('AWS_S3_REGION_NAME_MANAGED', default=env('AWS_S3_REGION_NAME', default=None)),
```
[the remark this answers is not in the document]
schtibe: Hmm, not sure I understand. This variable would be for the `managed` bucket and is derived from the env `AWS_S3_REGION_NAME_MANAGED` whereas the other is from `S3_REGION_NAME`

app/stac_api/models.py
```diff
@@ -494,6 +492,37 @@ def upload_asset_to_path_hook(instance, filename=None):
...
+        """
+        collection = model_instance.item.collection
+
+        bucket = select_s3_bucket(collection.name).name
+
+        parts = settings.STORAGES[bucket]['BACKEND'].split(".")
+        storage_module_name = ".".join(parts[:-1])
+        storage_cls_name = parts[-1:]
+        storage_module = __import__(storage_module_name, fromlist=[parts[-2:-1]])
+        storage_cls = getattr(storage_module, storage_cls_name[0])
```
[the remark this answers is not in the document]
schtibe: Oh yes 😂

app/stac_api/storages.py
```diff
@@ -60,3 +67,33 @@ def get_object_parameters(self, name):
...
+
+class ManagedS3Storage(BaseS3Storage):
+    # pylint: disable=abstract-method
+
+    def __init__(self):
+        # specifying the configuration as we're not using the
+        # environment variables / globals settings (due two having two
+        # buckets to access)
+        self.access_key = settings.AWS_SETTINGS['managed']['ACCESS_KEY_ID']
+        self.secret_key = settings.AWS_SETTINGS['managed']['SECRET_ACCESS_KEY']
```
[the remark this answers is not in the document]
schtibe: Yes indeed, thanks.

app/stac_api/models.py
```diff
@@ -494,6 +492,37 @@ def upload_asset_to_path_hook(instance, filename=None):
...
+
+        self.storage = storage_cls()
+
+        instance = model_instance
+
+        # we need to specify the storage for the actual
+        # file as well
+        instance.file.storage = self.storage
```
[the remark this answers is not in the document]
schtibe: Oh this is somewhat a leftover from changing the code all the time. This would make more sense as
`model_instance.file.storage = self.storage`
What this does is set the correct file storage on the actual file, so everything that uses the storage on the file instance directly will have the correct one.
Also see `upload_asset_to_path_hook` in this file and `BaseS3Storage.get_object_parameters` for how they interact across multiple code locations

app/config/settings_dev.py
```diff
@@ -45,11 +45,14 @@
...
-    DEBUG_PROPAGATE_API_EXCEPTIONS = bool(
-        strtobool(os.getenv('DEBUG_PROPAGATE_API_EXCEPTIONS', 'False'))
-    )
+    DEBUG_PROPAGATE_API_EXCEPTIONS = env.bool('DEBUG_PROPAGATE_API_EXCEPTIONS', 'False')
 
 SHELL_PLUS_POST_IMPORTS = ['from tests.data_factory import Factory']
+
+# Regex patterns of collections that should go to the managed bucket
+MANAGED_BUCKET_COLLECTION_PATTERNS = env.list(
+    'MANAGED_BUCKET_COLLECTION_PATTERNS', default=[r"^ch\.meteoschweiz\.ogd.*"]
```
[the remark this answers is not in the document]
schtibe: For the n-th time 🤦

app/stac_api/validators_serializer.py
schtibe: So as I see it, the usage of this function was removed with this PR: https://github.com/geoadmin/service-stac/pull/194

PR #425, opened 2024-07-01
PB-615: Spec multipart upload process

benschs:
Describe multipart upload process in more details. Added a quick description of the process. After talk with @boecklic we will add more detailed documentation (probably in another tool) later.


spec/static/spec/v1/openapitransactional.yaml
```diff
@@ -92,6 +92,34 @@ tags:
...
+
+      (2) Use the urls to [upload asset file part](#tag/Asset-Upload-Management/operation/uploadAssetFilePart). Do this for each file part.
+
+      (3) Once all parts have been uploaded, execute the [complete the upload](#tag/Asset-Upload-Management/operation/completeMultipartUpload) request.
+      The new asset file be available to users after you have successfully completed the upload.
+
+      If any errors happen during the upload process, you can [abort the upload](#tag/Asset-Upload-Management/operation/abortMultipartUpload)
+      and restart the process.
+
+      There can only be 1 upload in progress per asset. If you wish to start a new upload you must first abort any existing upload that is in progress.
```
adk-swisstopo, line 123: I think this is a bit unclear. From my understanding it's OK to upload multiple parts concurrently but you cannot "overwrite" (start a new upload of) an asset for which upload has not completed (or been aborted).

adk-swisstopo (approved):
Thanks. I think that's exactly the kind of high-level description we are missing.

spec/static/spec/v1/openapitransactional.yaml
```diff
@@ -92,6 +95,34 @@ tags:
...
+      To upload an assets file there are generally three steps involved:
+
+      <ol>
+        <li>Create new upload process</li>
+        <li>Upload parts/files</li>
+        <li>Complete upload process</li>
+      </ol>
+
+      Any file that is larger than 5 GB must be split into multiple parts. A file part must be at
+      least 5 MB except for the last one and at most 5 GB, otherwise the complete operation will fail.
```
hansmannj, line 109: How about "except for the last (or only) part" ?

spec/static/spec/v1/openapitransactional.yaml
hansmannj: Esp. the flow chart could be of interest to some users

hansmannj (approved):
Thanks, LGTM :+1:

spec/static/spec/v1/openapitransactional.yaml
adk-swisstopo: If we want to call it out, I would add a separate sentence instead of parenthesis. "If there is only one part, it can be smaller than 5 MB since it is also the last part."

spec/static/spec/v1/openapitransactional.yaml
adk-swisstopo: I wasn't aware of that doc. Maybe it should be pulled into the API doc itself instead of being a separate document? If it needs to remain separate, I strongly support linking to it.

spec/static/spec/v1/openapitransactional.yaml
```diff
@@ -599,7 +602,9 @@ paths:
           $ref: "#/components/responses/ServerError"
     delete:
       summary: Delete an existing feature by Id
-      description: Use this method to delete an existing feature/item.
+      description: |
+        Use this method to delete an existing feature/item. The feature can only be deleted if all
+        its assets have been deleted first. Also see [Delete an existing asset by id](#tag/Data-Management/operation/deleteAsset).
```
adk-swisstopo, line 607: "See also" sounds less weird than "Also see" I think.

adk-swisstopo (approved):

spec/static/spec/v1/openapitransactional.yaml
benschs: Adapted the concurrent upload part sentence. I was also not aware of this document. I added the detailed description of the upload to the spec, as well as the diagram asset to the static files.

adk-swisstopo: I suggest replacing the document with a redirect to the API doc that has the content. Less copypasta to maintain.

benschs: Yes, I removed the existing file as all the content is integrated in the spec. Also moved both diagram files to the static assets.

spec/static/spec/v1/openapitransactional.yaml
benschs: Following this sentence there is the following sentence:
"If the file is less that 5 GB, you will only upload a single part, but must still start and complete the process as with multiple parts."
Do you think it needs more than this?

PR #426, opened 2024-07-03
PB-769: Required fields to comply with stac v1

benschs:
Following changes are done to v1 (v0.9 stays unchanged)

**Landing Page**
- Add field `type`
- Add field `conformsTo`

**Collection**
- Add field `type`

**Item**
- Rename `checksum:multihash` to `file:checksum`
- Rename `eo:gsd` to `gsd`

These changes will remove the ConformancePage table from the database. The information is moved to the LandingPage.


boecklic (approved):
looks good. I left a comment on the test PR, there's a project that claims to be able to check STAC APIs for validity, maybe it could be useful to validate these changes and make sure we catched all changes

Pipfile
```diff
@@ -39,6 +39,7 @@ django-prometheus = "~=2.3.1"
 django-admin-autocomplete-filter = "~=0.7.1"
 django-pgtrigger = "~=4.11.1"
 logging-utilities = "~=4.4.1"
+stac-api-validator = "*"
```
ltshb, line 42: Should not this be in the dev-package ? In the package section we should have fix major, minor version.

benschs: :+1: Yes that makes sense. I added to dev-deps and fixed the version to the current latest.

app/stac_api/admin.py
```diff
@@ -15,14 +15,11 @@
 from django.http import HttpResponseRedirect
 from django.urls import reverse
 
-from solo.admin import SingletonModelAdmin
```
boecklic, line 18: was this the only we used that `SingletonModelAdmin`? If yes we could remove the package from the dependancy list.

app/stac_api/migrations/0033_auto_20240704_1157.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('stac_api', '0032_delete_conformancepage_landingpage_conformsto_and_more'),
+    ]
+
+    operations = [
+        migrations.RunPython(add_landing_page_version, reverse_landing_page_version),
```
boecklic, line 50: nice!

app/stac_api/migrations/0034_merge_0032_alter_asset_file_0033_auto_20240704_1157.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('stac_api', '0032_alter_asset_file'),
+        ('stac_api', '0033_auto_20240704_1157'),
+    ]
+
+    operations = []
```
boecklic, line 13: was this migration done automatically? Merging migrations has not always worked well in the past, but this is already a while ago, so it's probably reliable now.

app/stac_api/models.py
```diff
@@ -565,11 +557,15 @@ class Meta:
     def filename(self):
         return os.path.basename(self.file.name)
 
+    # From v1 on the json representation of this field changed from "checksum:multihash" to
+    # "file:checksum". The two names may be used interchangeably for a now.
     checksum_multihash = models.CharField(
```
boecklic, line 562: cosmetics, but we could rename the fields to relate to the naming in v1...

boecklic (approved):
very nice! thanks.

app/stac_api/admin.py
benschs: Thanks for the hint, I removed the dependency. :+1:

app/stac_api/migrations/0034_merge_0032_alter_asset_file_0033_auto_20240704_1157.py
benschs: Yes this was created automatically because of the conflicts with the changes for the new s3 bucket. I am not aware of any migration issues, it seems to work so far.

PR #430, opened 2024-07-10
Switch to a supported OpenAPI validator.

adk-swisstopo:
This requires some changes to the spec and I still disabled some checks that would be too disruptive. See commits for details.


boecklic (approved):
thanks for looking into this and updating validation!

PR #431, opened 2024-07-11
PB-769: STAC v1 optional asset features

benschs:
These changes are not _required_ to conform with STAC v1, but are possible and nice to have for our use-cases.

- Add `roles` field to assets to describe the purpose of the asset.
- Add assets to collections directly that do not belong to an item.


app/stac_api/admin.py
```diff
@@ -271,6 +272,106 @@ def save_model(self, request, obj, form, change):
...
+        (
+            'File', {
+                'fields': ('file', 'media_type', 'href', 'checksum_multihash', 'update_interval')
+            }
+        ),
+        ('Description', {
+            'fields': ('title', 'description', 'roles')
+        }),
+        ('Attributes', {
+            'fields': ('eo_gsd', 'proj_epsg', 'geoadmin_variant', 'geoadmin_lang')
```
boecklic, line 307: I think only `proj_epsg` is relevant on collection asset level, I'd use the other ones only on item asset level

app/stac_api/models.py
```diff
@@ -640,6 +634,50 @@ def clean(self):
...
+        triggers = generates_asset_triggers()
+
+    item = models.ForeignKey(
+        Item,
+        related_name='assets',
+        related_query_name='asset',
+        on_delete=models.PROTECT,
+        help_text=_(SEARCH_TEXT_HELP_ITEM)
+    )
+
```
boecklic, line 651: add the item asset specific parameters (`eo_gsd`, `geoadmin...` only here)

app/stac_api/pgtriggers.py
```diff
@@ -189,6 +196,113 @@ class UpdateItemUpdateIntervalTrigger(pgtrigger.Trigger):
...
+        func = '''
+        asset_instance = COALESCE(NEW, OLD);
+
+        -- Compute collection summaries
+        SELECT
+            a.collection_id,
+            array_remove(array_agg(DISTINCT(a.proj_epsg)), null) AS proj_epsg,
+            array_remove(array_agg(DISTINCT(a.geoadmin_variant)), null) AS geoadmin_variant,
+            array_remove(array_agg(DISTINCT(a.geoadmin_lang)), null) AS geoadmin_lang,
+            array_remove(array_agg(DISTINCT(a.eo_gsd)), null) AS eo_gsd
```
boecklic, line 225: not necessary here

app/stac_api/admin.py
```diff
@@ -271,6 +272,106 @@ def save_model(self, request, obj, form, change):
...
+        'update_interval'
+    ]
+    list_display = ['name', 'collection_name', 'collection_published']
+    fieldsets = (
+        (None, {
+            'fields': ('name', 'collection', 'created', 'updated', 'etag')
+        }),
+        (
+            'File', {
+                'fields': ('file', 'media_type', 'href', 'checksum_multihash', 'update_interval')
```
boecklic, line 300: I'm not sure we need the `update_interval` here... would it be considered at all? If I remember correctly we had set a fixed cache duration for assets uploaded via admin...

app/stac_api/pgtriggers.py
```diff
@@ -189,6 +196,113 @@ class UpdateItemUpdateIntervalTrigger(pgtrigger.Trigger):
...
+        GROUP BY a.collection_id;
+
+        -- Update related collection (auto variables + summaries)
+        UPDATE stac_api_collection SET
+            updated = now(),
+            etag = gen_random_uuid(),
+            summaries_proj_epsg = collection_summaries.proj_epsg,
+            summaries_geoadmin_variant = collection_summaries.geoadmin_variant,
+            summaries_geoadmin_lang = collection_summaries.geoadmin_lang,
+            summaries_eo_gsd = collection_summaries.eo_gsd
```
boecklic, line 246: remove

app/stac_api/pgtriggers.py
```diff
@@ -189,6 +196,113 @@ class UpdateItemUpdateIntervalTrigger(pgtrigger.Trigger):
...
+        asset_instance = COALESCE(NEW, OLD);
+        -- Update related collection update_interval variables
+        -- if new value is lower than existing one.
+        UPDATE stac_api_collection SET
+            update_interval = COALESCE(LEAST(NULLIF(asset_instance.update_interval, -1), update_interval), -1)
+        WHERE id = asset_instance.collection_id;
+        RAISE INFO 'collection.id=% update_interval updated, due to collectionasset.name=% updates.',
+            asset_instance.collection_id, asset_instance.name;
+        RETURN asset_instance;
+        '''
```
boecklic, line 269: can probably be removed completely, see comment above

boecklic (commented):
Nice!

app/stac_api/admin.py
boecklic: From discussion: cachecontrol header is set to a fixed value for assets uploaded via admin ui, so `update_interval` has no effect. We decided to keep it mainly to keep the two asset models and logic similar

app/stac_api/pgtriggers.py
benschs: As we keep the update_interval in the model we should also keep the trigger to avoid issues down the road.

boecklic (approved):

PR #433, opened 2024-07-16
PB-35 Implemented external asset urls

schtibe:
- Added a flag on the collection to allow for external assets and a char field to provide a pattern
- Use the href field on asset creation request to ingest external URLs if that's allowed from the collection
- Disallow multipart upload creation for the external assets
- Validate the external url for
    - General URL validity
    - If it matches the collection pattern
    - If the provided URL is reachable
- Adjust Asset admin page so that an asset can only be external if the collection allows it

I don't know if this is everything that's needed, but nevertheless I think a review would be good.


app/tests/tests_10/test_external_assets_endpoint.py
```diff
@@ -0,0 +1,206 @@
...
+    def test_create_asset_with_external_url(self):
+        collection = self.collection
+        item = self.item
+
+        asset_data = {
+            'id': 'clouds.jpg',
+            'description': 'High in the sky',
+            'type': 'image/jpeg',  # specify the file explicitly
+            'href':
+                'https://sys-data.int.bgdi.ch/ch.meteoschweiz.ogd-swisstopo-test/ch.meteoschweiz.ogd-swisstopo-testitem/clouds.jpg'
```
schtibe, line 34: Exposed URL

app/tests/tests_10/test_external_assets_endpoint.py
```diff
@@ -0,0 +1,206 @@
...
+        collection = self.collection
+        item = self.item
+
+        collection.allow_external_assets = True
+        collection.save()
+
+        asset = self.factory.create_asset_sample(item=self.item, sample='asset-1', db_create=True)
+
+        asset_data = asset.get_json('put')
+        asset_data['href'] = "https://sys-data.dev.bgdi.ch/ch.bgdi.test/test.jpg"
```
schtibe, line 74: Exposed URL

app/stac_api/serializers.py
```diff
@@ -338,13 +350,61 @@ def validate_type(self, value):
...
+
+        # get the collection regex. if there's none, allow all urls
+        collection_regex = self.collection.external_asset_pattern or r".*"
+
+        if not re.search(collection_regex, url):
+            errors = {'href': _("Invalid URL provided. It doesn't match the collection pattern")}
+            raise serializers.ValidationError(code='payload', detail=errors)
+
+        try:
+            response = requests.head(url, timeout=settings.EXTERNAL_URL_REACHABLE_TIMEOUT)
```
boecklic, line 373: an alternative for this check would be to just request a small range of the response (similar to e.g. `curl <URL> -i -H "Range: bytes=0-511"`, but this would require the external server to support `range`s, dunno if it makes sense to have that requirement (for the MCH case it probably makes sense, since they have big files as well where support of `range`s would allow for pause/resume download etc. 
I'll communicate to MCH that we'll do the check with `head`, mentioning the alternative using `range`.

app/tests/tests_10/sample_data/asset_samples.py
```diff
@@ -126,6 +126,18 @@
...
+    'external-asset': {
+        'name': 'test.jpg',
+        'title': 'An external asset',
+        'description': 'This asset is hosted externally',
+        'eo_gsd': 3.3,
+        'geoadmin_lang': 'fr',
+        'geoadmin_variant': 'kgrs',
+        'proj_epsg': 2056,
+        'media_type': 'image/jpeg',
+        'file': 'https://sys-data.dev.bgdi.ch/ch.bgdi.test/test.jpg',
```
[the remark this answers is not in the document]
boecklic: Not very nice, but not a huge problem either, can you easily avoid exposing it?

app/tests/tests_10/test_external_assets_endpoint.py
boecklic: For this to work we rely on the presence of that file (which was uploaded manually) right? Could we upload that file during the setup? To really make sure it's there? (Unlikely that it will be deleted, but still...)

boecklic (approved):
Very nice!

benschs:
Looks good. Maybe also merge all the db migration files into a single file?

app/stac_api/serializers.py
```diff
@@ -338,13 +350,61 @@ def validate_type(self, value):
...
+            errors = {'href': _('Invalid URL provided')}
+            raise serializers.ValidationError(code='payload', detail=errors)
+
+        # get the collection regex. if there's none, allow all urls
+        collection_regex = self.collection.external_asset_pattern or r".*"
+
+        if not re.search(collection_regex, url):
+            errors = {'href': _("Invalid URL provided. It doesn't match the collection pattern")}
+            raise serializers.ValidationError(code='payload', detail=errors)
+
```
boecklic, line 371: Regarding also @benschs comment: It would probably be safer to have prefixes and check with `startswith`, instead of just searching in the whole string.

app/tests/tests_10/test_external_assets_endpoint.py
schtibe: I'm gonna try!! :+1: :)

schtibe:
> Looks good. Maybe also merge all the db migration files into a single file?

Gonna do that, or at least partially. The changes to the models have separate commits, thus I'm gonna keep their migrations separate as well!

app/stac_api/serializers.py
schtibe: According to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Range the servers that don't support the header *may* ignore it. Don't know though what the common case is.

What would be a benefit of the Range over the Head?

app/stac_api/serializers.py
```diff
@@ -389,6 +403,61 @@ def to_representation(self, instance):
...
+            if not collection.allow_external_assets:
+                errors = {'href': _("Found read-only property in payload")}
+                raise serializers.ValidationError(code="payload", detail=errors)
+
+            self._validate_href_url(attrs['file'])
+
+    def validate(self, attrs):
+        self._validate_href_field(attrs)
+        return super().validate(attrs)
```
boecklic, line 459: One more thing that I just realized: I dunno what will be logged in this part by default, but it's probably a section that we want to have a bit more verbosity regarding logging to be able to debug issues

app/tests/tests_10/test_external_assets_endpoint.py
```diff
@@ -0,0 +1,238 @@
...
+
+    def test_create_asset_with_external_url(self):
+        collection = self.collection
+        item = self.item
+
+        asset_data = {
+            'id': 'clouds.jpg',
+            'description': 'High in the sky',
+            'type': 'image/jpeg',  # specify the file explicitly
+            'href': settings.EXTERNAL_TEST_ASSET_URL
```
boecklic, line 37: better than hardcoded :+1:

app/tests/tests_10/test_external_assets_endpoint.py
```diff
@@ -0,0 +1,238 @@
...
+        # create the asset with an existing one
+        response = self.client.put(
+            reverse_version('asset-detail', args=[collection.name, item.name, asset_data['id']]),
+            data=asset_data,
+            content_type="application/json"
+        )
+
+        self.assertStatusCode(201, response)
+
+        asset_data['href'] = 'https://bgdi.ch/notexist.jpg'
```
boecklic, line 161: maybe also use `test-domain.test` here (bgdi.ch actually exists and is forwarded to www.geo.admin.ch)

schtibe: The thing here is that I need something that can establish a connection but then returns 404, because the behaviour is different compared to the next test. But It should mention that, and I could easily use geo.admin.ch :+1:

app/stac_api/serializers.py
schtibe: Ok, good to know! I'm gonna leave it for now then, and we now have much better logging, too!

app/tests/tests_10/test_external_assets_endpoint.py
schtibe: Since it's now available from the env, I don't see how to do this reliably from the code. It's now up to who or whatever triggers the test to make sure that the file is available.
Could maybe be done in the builder pipeline, for that matter.

app/stac_api/admin.py
```diff
@@ -464,33 +470,117 @@ def save_model(self, request, obj, form, change):
...
     def get_fieldsets(self, request, obj=None):
-        fields = super().get_fieldsets(request, obj)
+        """Build the different field sets for the admin page
+
+        The create page takes less fields than the edit page. This is because
+        at creation time we don't know yet if the collection allows for external
+        assets and thus can't determine whether to show the flag or not
+        """
+
+        # Safe the request for use in the href field
```
benschs, line 510: small typo Sa**f**e -> Sa**v**e

benschs (approved):
Instead of needing to save the admin form to set a file or external asset, would it be possible to allow to set external or upload a file to begin with, and then return an error on save if the collection does not allow it?

schtibe:
> Instead of needing to save the admin form to set a file or external asset, would it be possible to allow to set external or upload a file to begin with, and then return an error on save if the collection does not allow it?

I'm not against it per se, but I would like to hear @boecklic's opinion, especially regarding as to how much effort we want to put into the admin or if that won't even be used that much.
BTW. the current workflow is the same as with the creation of Users (default Django), so it's at least a common workflow.

boecklic:
> > Instead of needing to save the admin form to set a file or external asset, would it be possible to allow to set external or upload a file to begin with, and then return an error on save if the collection does not allow it?
> 
> I'm not against it per se, but I would like to hear @boecklic's opinion, especially regarding as to how much effort we want to put into the admin or if that won't even be used that much. BTW. the current workflow is the same as with the creation of Users (default Django), so it's at least a common workflow.

Currently it's unlikely that external assets will be created through admin, except for testing purposes maybe. So the most effortless approach seems good to me.

boecklic (approved):

PR #434, opened 2024-07-24
PB-754: Improve summary calculation triggers

benschs:
**This is a first version of how the _summary calculations_ could be improved. I will also check on how we could run the calculations asynchronously, especially as we still need a solution for the _extent calculation_.**

With many assets per collection, the triggers to calculate the collection summaries is slow and CPU intensive.

Introduce counter tables to keep track of the different summary values per collection. Changes to the assets update the counter tables and changes to the counter tables update the collection summary. This works especially well if the number of distinct values is low. This solution essentially leverages more space for less computation.

Other refactoring:
- Triggers that set the `etag` and `updated` fields on the table and parent tables.
- Make trigger run conditions more strict so they only run if the relevant fields have changed.


app/stac_api/models.py
```diff
@@ -771,3 +773,77 @@ def update_asset_from_upload(self):
...
+    class Meta:
+        abstract = True
+
+    id = models.BigAutoField(primary_key=True)
+    collection = models.ForeignKey(
+        Collection,
+        related_name='+',
+        on_delete=models.CASCADE,
+    )
+    count = models.IntegerField(null=True, blank=True)
```
ltclm, line 797: what would be the use-case to have a value with null or blank count?
wouldn't it be better to not accept such situations and only store values having a count of 1 or bigger?

app/stac_api/models.py
```diff
@@ -771,3 +773,77 @@ def update_asset_from_upload(self):
...
+
+
+class GSDCount(CountBase):
+    # Update by asset triggers.
+
+    class Meta:
+        unique_together = (('collection', 'value'),)
```
ltclm, line 804: also for the index that will be created with unique_together i think it would make sense to not accept null/blank in the `count` column

app/stac_api/pgtriggers.py
```diff
@@ -61,79 +80,100 @@ def generate_child_triggers(parent_name, child_name):
...
             WHERE id = asset_instance.item_id
         );
 
-        -- Update related item auto variables
-        UPDATE stac_api_item SET
-            updated = now(),
-            etag = gen_random_uuid()
-        WHERE id = asset_instance.item_id;
+        UPDATE stac_api_{count_table}
+        SET count = count-1
```
ltclm, line 106: if count-1 will be 0 the entry could be removed from the `stac_api_{count_table}`

app/stac_api/pgtriggers.py
```diff
@@ -61,79 +80,100 @@ def generate_child_triggers(parent_name, child_name):
...
 
-        RAISE INFO 'collection.id=% summaries updated, due to asset.name=% update.',
-            related_collection_id, asset_instance.name;
+        related_collection_id = (
+            SELECT collection_id FROM stac_api_item
+            WHERE id = asset_instance.item_id
+        );
+
+        INSERT INTO stac_api_{count_table} (collection_id, value, count)
+        VALUES (related_collection_id, asset_instance.{value_field}, 1)
```
ltclm, line 129: nice solution with the ON CONFLICT

app/stac_api/pgtriggers.py
```diff
@@ -262,28 +249,68 @@ class UpdateCollectionUpdateIntervalTrigger(pgtrigger.Trigger):
...
+        when = pgtrigger.After
+        declare = [('asset_instance', 'stac_api_collectionasset%ROWTYPE'),
+                   ('related_collection_id', 'INT')]
+        func = '''
+        asset_instance = OLD;
+
+        related_collection_id = asset_instance.collection_id;
+
+        UPDATE stac_api_projepsgcount
+        SET count = count-1
```
ltclm, line 262: same here, why not remove when `count=0`?

app/stac_api/pgtriggers.py
```diff
@@ -369,23 +394,18 @@ class UpdateCollectionUpdateIntervalTrigger(pgtrigger.Trigger):
...
-            func=AUTO_VARIABLES_FUNC
-        ),
+        *auto_variables_triggers('item'),
+        *generate_child_triggers('collection', 'Item'),
         UpdateCollectionExtentTrigger(
             name='update_item_collection_extent_trigger',
             operation=pgtrigger.Update,
-            condition=pgtrigger.Condition('OLD.* IS DISTINCT FROM NEW.*')
+            # condition=pgtrigger.Condition('OLD.* IS DISTINCT FROM NEW.*')
+            condition=pgtrigger.Condition(
```
ltclm, line 403: yes good improvement here thank you, that should prevent us from useless extent aggregation in the future :+1:

app/stac_api/pgtriggers.py
```diff
@@ -461,3 +470,54 @@ def generates_asset_upload_triggers():
...
+        RAISE INFO 'collection.id=% summaries updated',
+            count_instance.collection_id;
+        RETURN count_instance;
+        '''
+
+    return [
+        UpdateCollectionSummaryTrigger(
+            name=f'update_collection_{count_table}_trigger',
+            operation=pgtrigger.Update,
+            condition=pgtrigger.Condition('OLD.count < 2 OR NEW.count < 2')
```
ltclm, line 517: not sure if i understand the purpose of this condition here but the trigger will only be fired if the old or the new count is smaller than 2? why is that necessary?

ltclm (changes requested):
thank you very much for that work :+1: 
I only have a few general remarks and questions, see comments.

since you are improving quite a lot and introducing new logic into the aggregation, it would be great to have a few numbers of even postgres query plans comparing the old and the new code? so we would know the impact on the performance when updating assets on larger collections (INSERT/UPDATE/DELETE).

app/stac_api/pgtriggers.py
benschs: Yes, I forgot to look at that. I changed it so now the entry should be removed if count is 0. Thanks!

app/stac_api/pgtriggers.py
benschs: Yes, I forgot to look at that. I changed it so now the entry should be removed if count is 0. Thanks!

app/stac_api/pgtriggers.py
benschs: The idea is that we don't need to take the aggregate list again if the count changed from like 3 to 4, as there is no new entry. we are only really interested in the case where the count changes from 0 to 1 and vice versa.
I slightly changed the Condition and added a comment to try and make it clearer.

app/stac_api/pgtriggers.py
```diff
@@ -61,79 +80,114 @@ def generate_child_triggers(parent_name, child_name):
...
+         value_field: summary field name on the asset
 
-    Returns: tuple
-        tuple for all needed triggers
+    Returns:
+        List of triggers
     '''
 
-    class UpdateCollectionSummariesTrigger(pgtrigger.Trigger):
+    class UpdateDeleteCounterTrigger(pgtrigger.Trigger):
```
schtibe, line 94: In all honesty, the naming of these classes confuse me a bit (and yes I know, you didn't start it).
I think it is supposed to say "update the delete-counter-trigger", right? Yet classes usually don't contain verbs I would argue? And it's especially confusing since "Update" is a common word in SQL :D

app/stac_api/pgtriggers.py
```diff
@@ -61,79 +80,114 @@ def generate_child_triggers(parent_name, child_name):
...
+            Condition(f'''OLD.{value_field} IS DISTINCT FROM NEW.{value_field}''')
+        ),
+        UpdateInsertCounterTrigger(
+            name=f'add_{value_field}_trigger',
+            operation=pgtrigger.Insert,
+        ),
+    ]
+
+
+def generates_asset_triggers():
```
schtibe, line 178: Could this function be merged with `asset_counter_trigger` like it is with `generate_collection_asset_triggers`?

schtibe (commented):
Wicked stuff! 🚀

app/stac_api/models.py
```diff
@@ -771,3 +773,77 @@ def update_asset_from_upload(self):
...
+    collection also stays performant.
+    '''
+
+    class Meta:
+        abstract = True
+
+    id = models.BigAutoField(primary_key=True)
+    collection = models.ForeignKey(
+        Collection,
+        related_name='+',
```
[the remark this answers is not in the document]
benschs: According to the [doc](https://docs.djangoproject.com/en/5.0/ref/models/fields/#django.db.models.ForeignKey.related_name) using `+` does not add a backward relation, so `collection_instance.+.objects.all()` will not work. My idea was that these counter tables are just a kind of _memory_ for the triggers and should not be used otherwise within the model.

app/stac_api/pgtriggers.py
```diff
@@ -61,79 +80,114 @@ def generate_child_triggers(parent_name, child_name):
...
+def asset_counter_trigger(count_table, value_field):
+    '''Creates triggers to adjust the asset summary on a counter table.
 
-    Those triggers act on `insert`, `update` and `delete` Asset event and do the followings:
-      - Update the `updated` and `etag` fields of the assets and their parents
-      - Update the parent collection summaries.
-      - Update the parent item `update_interval` by using the minimal aggregation of all of its
-        assets
+    Args:
+         count_table: tha table name to be updated (without prefix stac_api_)
```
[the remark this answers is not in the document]
benschs: Definitely a typo, Thanks :)

app/stac_api/pgtriggers.py
```diff
@@ -461,3 +497,56 @@ def generates_asset_upload_triggers():
             func=etag_func
         ),
     ]
+
+
+class SummaryFields(Enum):
+    # pylint: disable=invalid-name
+    GSD = 'summaries_eo_gsd', 'gsdcount'
```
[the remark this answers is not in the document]
benschs: I tried this, but could not get it to work without ending in circular dependency errors. Would be happy to try if you have more details on how to achieve this.

app/stac_api/pgtriggers.py
benschs: My initial idea in this case was that the trigger is used in SQL Update and SQL Delete command, But I see how this is confusing. I refactored the naming of the functions and classes used in the triggers to try and make it more clear. Now there are functions that create "generic triggers" and functions that create triggers for a specific model (these start with `generates_`).

app/stac_api/pgtriggers.py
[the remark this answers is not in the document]
schtibe: I see. Maybe it could be done with https://docs.djangoproject.com/en/5.0/ref/applications/#django.apps.AppConfig.get_model

boecklic (approved):
:+1: (comments on PR https://github.com/geoadmin/service-stac/pull/435)

ltclm (approved):

PR #435, opened 2024-07-24
PB-754: Calculate extent management command

benschs:
Draft to create a management command to calculate the collection extent asynchronously. This command would need to be scheduled by a cron (k8s cron?) job.


app/stac_api/management/commands/calculate_extent.py
```diff
@@ -0,0 +1,108 @@
...
+            '-f',
+            '--force',
+            type=bool,
+            nargs='?',
+            default=False,
+            help='Run all without confirmation'
+        )
+
+    def handle(self, *args, **options):
+        # if flag is passed without argument value is set to none
```
schtibe, line 103: I think you might be able to simplify this by adding `action='store_true'` to the add_argument of these fields. See https://docs.python.org/3/library/argparse.html#action

app/stac_api/management/commands/calculate_extent.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def run(self):
+        self.print_success('Runing command to update collection extents')
+        with connection.cursor() as cursor:
+            # print(self.options)
+            if self.options['all']:
+                cursor.execute("SELECT id FROM stac_api_collection")
+            else:
+                cursor.execute("SELECT id FROM stac_api_collection WHERE extent_out_of_sync = TRUE")
+            collections = cursor.fetchall()
```
schtibe, line 27: Why using the cursor and not django orm?

app/stac_api/management/commands/calculate_extent.py
```diff
@@ -0,0 +1,108 @@
...
+                    "Are you sure you want to continue? (y/n): ",
+                    False
+                )
+                if not cont:
+                    self.print_warning("Aborted")
+                    self.print_success("No collections updated")
+                    return
+
+            start = time.monotonic()
+            for collection_id in collections:
```
schtibe, line 41: Do we need some error handling with proper logging here?

app/stac_api/management/commands/calculate_extent.py
```diff
@@ -0,0 +1,108 @@
...
+                        extent_out_of_sync = FALSE,
+                        extent_geometry = collection_extent.extent_geometry,
+                        extent_start_datetime = collection_extent.extent_start_datetime,
+                        extent_end_datetime = collection_extent.extent_end_datetime
+                    FROM collection_extent
+                    WHERE id = collection_extent.collection_id;
+                    """, [collection_id[0], collection_id[0]]
+                )
+                self.print_success(f"collection.id={collection_id[0]} extent updated.")
+            self.print_success(
```
schtibe, line 72: Maybe log the success too. Unless k8s cronjob stdout is already logged properly by our setup, I don't know?

app/stac_api/management/commands/calculate_extent.py
benschs: Thanks, this is exactly what I was looking for.

app/stac_api/management/commands/calculate_extent.py
benschs: I started with the update command and then also put the select into the cursor without much thought.
Changed it to have the select commands done by orm as I feel it looks nicer. But kept the actual update command in cursor as I don't know how that could be done using the orm.

app/stac_api/migrations/0047_prefill_counter_tables.py
```diff
@@ -0,0 +1,62 @@
+# Generated by Django 5.0.7 on 2024-07-23 15:14
+
+from django.db import migrations
+
+
+# The migration SQL in this file was created manually.
+class Migration(migrations.Migration):
+    migrate_sql = '''
```
boecklic, line 8: From offline discussion: It's unlikely that these count tables will get out of sync, cause all triggers are atomic within a transaction. However, it's useful to have a mgmt command ready to reset/recalculate these values in case things get out of sync for some reason

app/stac_api/management/commands/calculate_extent.py
```diff
@@ -0,0 +1,94 @@
...
+                    UPDATE stac_api_collection SET
+                        extent_out_of_sync = FALSE,
+                        extent_geometry = collection_extent.extent_geometry,
+                        extent_start_datetime = collection_extent.extent_start_datetime,
+                        extent_end_datetime = collection_extent.extent_end_datetime
+                    FROM collection_extent
+                    WHERE id = collection_extent.collection_id;
+                    """, [collection_id, collection_id]
+                )
+                self.print_success(f"collection.id={collection_id} extent updated.")
```
boecklic, line 72: `print_success` writes to `stdout`, which makes sense since these commands are mostly executed in an interactive shell. In case they are executed as cron job we need to have the output in a log message to ensure proper json-formatting (otherwise it cannot be correctly parsed by the log pipeline)

boecklic (approved):
Great improvement! :rocket:

ltclm (approved):
nothing to add from my side. thank you :+1:

PR #436, opened 2024-07-29
PB-851 Inline the collection assets to collection in admin

schtibe:
Inline the collection assets to the collection's admin page.

The simplest, most straight forward implementation. Don't know whether the use case demands a better way. How many collection assets are usually in a collection?
I guess I could easily spend more time on this and create a custom template for this model, but on the other hand this will make it less dynamic in case of future changes..


boecklic (approved):
Perfect to start :+1: 
The current usecase has max only a handful of assets

PR #437, opened 2024-07-30
PB-286: Remove unique constraint from rel-collection pair

LukasJoss:
Remove unique constraint between collection and collection links to allow e.g. multiple license owners


boecklic (approved):
Nice :+1: 
(a very small hint: the reason _why_ the constraint was removed would be useful to have in the commit message as well. The fact that it was removed can be seen from the diff, but not the reason, so that info is useful to have in the commit as well)

PR #438, opened 2024-08-07
Actually test that login works correctly.

adk-swisstopo:
/api/stac/admin always redirects so we don't actually know whether login succeeds when testing that endpoint. In this change we modify the test to point to /api/stac/admin/ which does return 200 when valid credentials are used and 302 when failing to log in. We also add a test to verify invalid credentials are correctly rejected/redirected.


boecklic (approved):
:+1:

PR #439, opened 2024-08-07
Test the CORS headers for the actual landing page.

adk-swisstopo:
This test function assertion was added in https://github.com/geoadmin/service-stac/commit/54dca401bfcac33baf9cbf9fd8319bca253d6124 It only tests the redirection to the landing page but not the actual landing page.

This change ensures the landing page is also covered by this test.

It may make sense to test all the pages, which could be done in a separate change.


ltshb (approved):

adk-swisstopo:
Build failure seems unrelated and caused by PB-694.

PR #442, opened 2024-08-08
PB-754: Endpoint to immediately calculate extent

benschs:
Add POST endpoint to run the management command to calculate the collection extents.
This is especially useful for e2e testing.

Should this be added to the spec and "officially" available?


boecklic (approved):
I wouldn't make this neither official nor (once we have access control) accessible by default users. I see this (currently) as a pure admin feature.

PR #443, opened 2024-08-12
PB-569 Switch from docker-compose to docker compose

schtibe:
docker-compose is obsolete. Instead we use the `docker compose` plugin
henceforth


boecklic (approved):
:+1:

ltshb (approved):

PR #444, opened 2024-08-12
PB-200: Add geometry checks to items

LukasJoss:


benschs (approved):
:+1: Code looks good.

PR #446, opened 2024-08-13
PB-569: Remove obsolete "version" from docker-compose.yml.

adk-swisstopo:
This removes the following warning when running `make setup`:

    WARN[0000] docker-compose.yml: the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion


schtibe (approved):
Oh right, I guess that could be removed from almost all the repos :+1:

PR #448, opened 2024-08-13
PB-612: actually test links start with http(s)://

adk-swisstopo:
Since "https" does start with "http", `'https'.startswith('http')` is not a useful test. This change ensures we test for "https://" and "http://" respectively.


adk-swisstopo:
Test failure is unrelated PB-694 again.

PR #449, opened 2024-08-14
Deflake serializer tests.

adk-swisstopo:
The serializer test involve comparing the "created" and "updated" fields of collections. These fields are populated with whatever Django thinks is the current time. Until this PR, the tests assumed that comparing to "test start time" +/- 2 seconds was good enough. As per PB-694 this has proved wrong several times.

This change updates the tests such that we only verify the field values are between the start time and end time of each test. As long as time goes forward this should be fine. In any case, it should already make the test much more reliable.

We verified this works by adding a 5 seconds delay after the start of the test (which was removed in the second to last change in this PR).


benschs (approved):
Thanks

PR #450, opened 2024-08-14
PB-847: Add item field 'expires'

benschs:
Add optional field 'expires' to item property.
New field should conform to stac-extensions/timestamps (https://github.com/stac-extensions/timestamps).
Refactor item properties validation function.


app/stac_api/validators.py
```diff
@@ -364,8 +364,20 @@ def validate_geometry(geometry):
     return geometry
 
 
-def validate_item_properties_datetimes_dependencies(
-    properties_datetime, properties_start_datetime, properties_end_datetime
+def validate_datetime_format(date_string):
+    try:
+        if not isinstance(date_string, datetime) and date_string is not None:
```
schtibe, line 369: Just some nitpick: wouldn't it make sense to first check for None and then for the instance?

app/stac_api/validators.py
```diff
@@ -364,8 +364,20 @@ def validate_geometry(geometry):
...
+        return date_string
+    except ValueError as error:
+        logger.error("Invalid datetime string %s", error)
+        raise ValidationError(
+            _('Invalid datetime string %(error)s'), params={'error': error}, code='invalid'
+        ) from error
+
+
+def validate_item_properties_datetimes(
+    properties_datetime, properties_start_datetime, properties_end_datetime, properties_expires
```
schtibe, line 380: Just some higher-level thinking here, (it doesn't have to do with your change directly):
I think it might be better if the callers of this function made sure that what they pass in is either `datetime` or `None`. Instead, the above function, as a side-effect, parses the dates in case they are strings, even though this here is about validation... (I guess that's a result of not having strict typing 😁)

spec/components/schemas.yaml
```diff
@@ -1072,6 +1072,8 @@ components:
           $ref: "#/components/schemas/datetime"
         end_datetime:
           $ref: "#/components/schemas/datetime"
```
schtibe, line 1074: Is it meant to be on both versions of our API (v0.9 and v1)

app/stac_api/validators.py
benschs: I played around with different options, but nothing that is really nicer.

spec/components/schemas.yaml
benschs: AFAIK the v0.9 spec should not change anymore. The new feature should only be in the v1 spec.
To keep things easier (and as it is not a breaking change), I decided to add the field to response of both versions. We could also check for v0.9 and remove the 'expires' field from the response again.

app/stac_api/validators.py
```diff
@@ -408,25 +413,15 @@ def validate_item_properties_datetimes_dependencies(
...
-    '''
-    Validate datetime values in the properties Item attributes
-    '''
-    validate_item_properties_datetimes_dependencies(
-        properties_datetime,
-        properties_start_datetime,
-        properties_end_datetime,
-    )
+        if properties_expires is not None:
+            if properties_expires < properties_end_datetime:
```
boecklic, line 421: maybe we should even enforce a minimal lifetime of an object, e.g. 24/72/... hours?

boecklic (approved):
Nice!
If we want to be perfect, the standard suggest to add extensions to the corresponding `stac_extensions` list, e.g. for items: https://github.com/radiantearth/stac-spec/blob/master/item-spec/item-spec.md#stac_extensions
We haven't done this so far though

app/stac_api/validators.py
benschs: As discussed offline, we are not sure about the different use cases, so for now we keep it simple and don't add more restrictions.

PR #451, opened 2024-08-15
PB-848: Delete expired items

benschs:
New management command to delete expired items and their assets.
Remove expired items from result in get & list items/assets endpoints.

### DIFF
--- a/app/stac_api/management/commands/remove_expired_items.py
+++ b/app/stac_api/management/commands/remove_expired_items.py
@@ -10,13 +10,6 @@ from stac_api.utils import CommandHandler
 from stac_api.utils import CustomBaseCommand
 
 
-def boolean_input(question, default=None):
-    result = input(f"{question}")
-    if not result and default is not None:
-        return default
-    return len(result) > 0 and result[0].lower() == "y"
-
-
 class Handler(CommandHandler):
 
     def delete(self, instance, object_type):
--- a/app/stac_api/management/commands/remove_expired_items.py
+++ b/app/stac_api/management/commands/remove_expired_items.py
@@ -4,7 +4,6 @@ from django.conf import settings
 from django.core.management.base import CommandParser
 from django.utils import timezone
 
-from stac_api.models import Asset
 from stac_api.models import Item
 from stac_api.utils import CommandHandler
 from stac_api.utils import CustomBaseCommand
@@ -19,22 +18,21 @@ class Handler(CommandHandler):
             instance.delete()
 
     def run(self):
-        # print(self.options)
         self.print_success('running command to remove expired items')
         min_age_hours = self.options['min_age_hours']
         self.print_warning(f"deleting all items expired longer than {min_age_hours} hours")
         items = Item.objects.filter(
             properties_expires__lte=timezone.now() - timedelta(hours=min_age_hours)
         ).all()
-        for i in items:
-            assets = Asset.objects.filter(item_id=i.id).all()
+        for item in items:
+            assets = item.assets.all()
             assets_length = len(assets)
             self.delete(assets, 'assets')
-            self.delete(i, 'item')
+            self.delete(item, 'item')
             if not self.options['dry_run']:
                 self.print_success(
-                    f"deleted item {i.name} and {assets_length}" + " assets belonging to it.",
-                    extra={"item": i.name}
+                    f"deleted item {item.name} and {assets_length}" + " assets belonging to it.",
+                    extra={"item": item.name}
                 )
 
         if self.options['dry_run']:
--- a/app/stac_api/management/commands/remove_expired_items.py
+++ b/app/stac_api/management/commands/remove_expired_items.py
@@ -1,0 +1,75 @@
+from datetime import timedelta
+
+from django.conf import settings
+from django.core.management.base import CommandParser
+from django.utils import timezone
+
+from stac_api.models import Asset
+from stac_api.models import Item
+from stac_api.utils import CommandHandler
+from stac_api.utils import CustomBaseCommand
+
+
+def boolean_input(question, default=None):
+    result = input(f"{question}")
+    if not result and default is not None:
+        return default
+    return len(result) > 0 and result[0].lower() == "y"
+
+
+class Handler(CommandHandler):
+
+    def delete(self, instance, object_type):
+        if self.options['dry_run']:
+            self.print_success(f'skipping deletion of {object_type} {instance}')
+        else:
+            instance.delete()
+
+    def run(self):
+        # print(self.options)
+        self.print_success('running command to remove expired items')
+        min_age_hours = self.options['min_age_hours']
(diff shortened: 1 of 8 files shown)


app/stac_api/management/commands/remove_expired_items.py
```diff
@@ -0,0 +1,68 @@
...
+class Handler(CommandHandler):
+
+    def delete(self, instance, object_type):
+        if self.options['dry_run']:
+            self.print_success(f'skipping deletion of {object_type} {instance}')
+        else:
+            instance.delete()
+
+    def run(self):
+        # print(self.options)
```
schtibe, line 22: Little leftover?

app/stac_api/management/commands/remove_expired_items.py
```diff
@@ -0,0 +1,68 @@
...
+        self.print_success('running command to remove expired items')
+        min_age_hours = self.options['min_age_hours']
+        self.print_warning(f"deleting all items expired longer than {min_age_hours} hours")
+        items = Item.objects.filter(
+            properties_expires__lte=timezone.now() - timedelta(hours=min_age_hours)
+        ).all()
+        for i in items:
+            assets = Asset.objects.filter(item_id=i.id).all()
```
schtibe, line 30: JFYI I think you could use reverse relations here: `i.assets.all()`

app/stac_api/management/commands/remove_expired_items.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def run(self):
+        # print(self.options)
+        self.print_success('running command to remove expired items')
+        min_age_hours = self.options['min_age_hours']
+        self.print_warning(f"deleting all items expired longer than {min_age_hours} hours")
+        items = Item.objects.filter(
+            properties_expires__lte=timezone.now() - timedelta(hours=min_age_hours)
+        ).all()
+        for i in items:
```
schtibe, line 29: Just personal taste: I find `i` as an iterator suboptimal (except maybe for simple numeric for loops). But I'm very curious about other opinions!

app/stac_api/management/commands/remove_expired_items.py
```diff
@@ -0,0 +1,68 @@
...
+
+from stac_api.models import Asset
+from stac_api.models import Item
+from stac_api.utils import CommandHandler
+from stac_api.utils import CustomBaseCommand
+
+
+class Handler(CommandHandler):
+
+    def delete(self, instance, object_type):
```
schtibe, line 15: Looking at the code below, you're actually not passing an `instance` here, but rather a set of instances. (Actually, from a type POV I believe it's a queryset).
Update: OK on the second call it is *just* an instance :)

app/stac_api/management/commands/remove_expired_items.py
```diff
@@ -0,0 +1,68 @@
...
+class Command(CustomBaseCommand):
+    help = """Remove items and their assets that have expired more than
+    DELETE_EXPIRED_ITEMS_OLDER_THAN_HOURS hours ago.
+    This command is thought to be scheduled as cron job.
+    """
+
+    def add_arguments(self, parser: CommandParser) -> None:
+        super().add_arguments(parser)
+        parser.add_argument(
+            '--dry-run',
```
schtibe, line 55: Good idea!

schtibe (commented):
Very nice. I just have some minor comments on some details in the code.

Otherwise, I'm just pondering about our solution to *actually* delete the files on the bucket. I don't quite remember what we ultimately decided there.

schtibe (approved):

app/stac_api/management/commands/remove_expired_items.py
benschs: Thanks. I changed it as I think it's nicer. :+1:

app/stac_api/management/commands/remove_expired_items.py
benschs: I usually choose names based on the scope the variable is used.
Usually very few letters if it is only used in the next few lines, more letters/descriptive names if it used used "further away". In this case I see that "i" is also often used as index, so I am also fine with 4 letters "item".

app/stac_api/validators_view.py
```diff
@@ -38,7 +40,9 @@ def validate_item(kwargs):
         Http404: when the item doesn't exists
     '''
     if not Item.objects.filter(
-        name=kwargs['item_name'], collection__name=kwargs['collection_name']
+        Q(properties_expires=None) | Q(properties_expires__gte=timezone.now()),
+        name=kwargs['item_name'],
+        collection__name=kwargs['collection_name']
     ).exists():
         logger.error(
             "The item %s is not part of the collection %s",
```
LukasJoss, line 48: It might make sense to have a different error message in case that the item expired (though it is probably fine if the delete delay is only 24 hours)

LukasJoss (approved):
A validation check would be useful when creating an item to check that the expire date is after the current time instead of just after the item properties time.
Currently I can add an item on the admin page with an expiry date in the past.

app/stac_api/validators_view.py
benschs: Could possibly do this. In my understanding from the user perspective the items are 'deleted' as soon as they are expired. In the admin page the expired items are still available (for 24 hours) and even editable to extend the expiration.

app/stac_api/management/commands/remove_expired_items.py
```diff
@@ -0,0 +1,66 @@
...
+        ).all()
+        for item in items:
+            assets = item.assets.all()
+            assets_length = len(assets)
+            self.delete(assets, 'assets')
+            self.delete(item, 'item')
+            if not self.options['dry_run']:
+                self.print_success(
+                    f"deleted item {item.name} and {assets_length}" + " assets belonging to it.",
+                    extra={"item": item.name}
```
boecklic, line 35: I'm not sure this will automatically be parsed correctly unless the log config isn't adapted. But this is probably sth we should look at separately

app/stac_api/management/commands/remove_expired_items.py
```diff
@@ -0,0 +1,66 @@
...
+                )
+
+        if self.options['dry_run']:
+            self.print_success(f'[dry run] would have removed {len(items)} expired items')
```
boecklic, line 39: If I understand correctly this will end up as a normal log message as well?

boecklic (approved):
Nice, thanks!

app/stac_api/management/commands/remove_expired_items.py
benschs: As discussed offline we will look at this separately.

app/stac_api/management/commands/remove_expired_items.py
benschs: As discussed offline we will look at this separately.

PR #454, opened 2024-09-02
PB-511: Add request context to all logs

benschs:
Add middleware and logging filter to add the django request to all log records.
Also see: https://github.com/geoadmin/lib-py-logging-utilities/pull/62


app/config/settings_prod.py
```diff
@@ -81,6 +81,7 @@
 # last, put everything else in between
 MIDDLEWARE = [
     'django_prometheus.middleware.PrometheusBeforeMiddleware',
+    'logging_utilities.django_middlewares.add_request_context.AddToThreadContextMiddleware',
```
ltshb, line 84: It would be wise to add a comment why this middleware should be on top.

ltshb (approved):

PR #455, opened 2024-09-04
PB-755: Spec for collection assets

benschs:
Add CRUD endpoints for collection level assets.
Add endpoints for collection asset upload management. As process for upload is the same as for feature assets, only link to description.

### DIFF
--- a/spec/static/spec/v1/openapitransactional.yaml
+++ b/spec/static/spec/v1/openapitransactional.yaml
@@ -112,7 +112,7 @@ tags:
       (1) Use the [create new upload](#tag/Asset-Upload-Management/operation/createAssetUpload) request to start a new upload.
       It will return a list of urls.
 
-      (2) Use the urls to [upload asset file part](#tag/Asset-Upload-Management/operation/uploadAssetFilePart). Do this for each file part.
+      (2) Use the urls to [upload asset file part](#tag/Asset-Upload-Management/operation/uploadAssetFilePart). Do this for each file part. You may also upload multiple parts in parallel.
 
       (3) Once all parts have been uploaded, execute the [complete the upload](#tag/Asset-Upload-Management/operation/completeMultipartUpload) request.
       The new asset file be available to users after you have successfully completed the upload.
@@ -296,6 +296,16 @@ tags:
       ```
 
       See https://aws.amazon.com/premiumsupport/knowledge-center/data-integrity-s3/ for other examples on how to compute the base64 MD5 of a part.
+  - name: Collection Asset Upload Management
+    description: |
+      Collection Asset files are uploaded via the STAC API using the API requests described in this chapter.
+
+      The flow of the requests is the same as for assets that belong to features, which is described under [Asset Upload Management](#tag/Asset-Upload-Management)
+      - [Basic steps to upload assets](#section/Basic-Steps-to-upload-assets)
+      - [Detailed example of asset upload](#section/Detailed-example-of-asset-upload)
+      - [Authentication](#section/Authentication)
+      - [Compression](#section/Compression)
+      - [Example](#section/Example)
   - name: Authentication
     description: |
       All write requests require authentication. There is currently three type of supported authentications:
@@ -416,7 +426,7 @@ paths:
         - Data Management
       summary: Update or create a collection
       description: >-
-        Update or create a collection with Id `collectionId` with a complete collection definition. If the collection doesn't exists it is then created.
+        Update or create a collection with Id `collectionId` with a complete collection definition. If the collection doesn't exist it is created.
       operationId: updateCollection
       parameters:
         - $ref: "#/components/parameters/collectionId"
@@ -574,7 +584,7 @@ paths:
     put:
       summary: Update or create a feature
       description: >-
-        Update or create a feature with Id `featureId` with a complete feature definition. If the feature doesn't exists it is then created.
+        Update or create a feature with Id `featureId` with a complete feature definition. If the feature doesn't exist it is then created.
 
 
         *NOTE: Optional fields that are not part of the PUT payload, will be erased in the resource. For example if the resource has a properties.title and the PUT payload doesn't, then the resource's properties.title will be removed.*
@@ -641,12 +651,12 @@ paths:
           $ref: "#/components/responses/NotFound"
         "412":
           $ref: "#/components/responses/PreconditionFailed"
-        5XX:
+        "500":
           $ref: "#/components/responses/ServerError"
     patch:
       summary: Update an existing feature by Id with a partial item definition
       description: >-
-        Use this method to update an existing feature. Requires a JSON fragment (containing the fields to be updated) be submitted.
+        Use this method to update an existing feature. Requires a JSON fragment (containing the fields to be updated) to be submitted.
       operationId: patchFeature
       tags:
         - Data Management
@@ -686,7 +696,7 @@ paths:
           $ref: "#/components/responses/NotFound"
         "412":
           $ref: "#/components/responses/PreconditionFailed"
-        5XX:
+        "500":
           $ref: "#/components/responses/ServerError"
     delete:
       summary: Delete an existing feature by Id
@@ -709,7 +719,7 @@ paths:
           $ref: "#/components/responses/NotFound"
         "412":
           $ref: "#/components/responses/PreconditionFailed"
-        5XX:
+        "500":
           $ref: "#/components/responses/ServerError"
   /{assetObjectHref}:
     servers:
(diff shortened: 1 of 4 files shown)


boecklic (approved):
:+1:

spec/transaction/components/responses.yaml
```diff
@@ -18,6 +25,16 @@ components:
         application/json:
           schema:
             $ref: "./schemas.yaml#/components/schemas/assetWrite"
+    CollectionAsset:
+      description: >-
+        The response is a document consisting of one asset of the collection.
+      headers:
+        ETag:
+          $ref: "../../components/headers.yaml#/components/headers/ETag"
```
asteiner-swisstopo, line 33: For my understanding: Why does the `CollectionAssets` have no `headers` while `CollectionAsset` does?

spec/transaction/paths.yaml
```diff
@@ -789,6 +942,249 @@ paths:
...
+      summary: Get a collection assets multipart upload
+      description: |
+        Return a collection assets multipart upload.
+      operationId: getCollectionAssetUpload
+      parameters:
+        - $ref: "../components/parameters.yaml#/components/parameters/IfMatch"
+        - $ref: "../components/parameters.yaml#/components/parameters/IfNoneMatch"
+      responses:
+        "200":
+          description: Collection asset multipart upload description.
```
asteiner-swisstopo, line 1062: Looks like a stub description that has not been changed yet.

spec/transaction/paths.yaml
```diff
@@ -789,6 +942,249 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: "./components/schemas.yaml#/components/schemas/assetUploadCreate"
+        "400":
+          $ref: "../components/responses.yaml#/components/responses/BadRequest"
+        "404":
+          $ref: "../components/responses.yaml#/components/responses/NotFound"
+        "409":
+          description: Another upload is already in progress, you need first to complete or abort it before creating a new upload.
```
asteiner-swisstopo, line 1038: Shouldn't this be "you first need to complete" instead of "you need first to complete"?

spec/transaction/paths.yaml
```diff
@@ -789,6 +942,249 @@ paths:
...
+      tags:
+        - Collection Asset Upload Management
+      summary: Create a new collection asset multipart upload
+      description: |
+        Create collection asset multipart upload.
+
+        A file part must be at least 5 MB except for the last one and at most 5 GB, otherwise the complete operation will fail.
+
+        *Note: in order to provide integrity check during the upload, the base64-encoded 128-bit MD5 digest of each part must be
+        computed and passed in the create endpoint. Then this digest must also be passed as `Content-MD5` header during the upload.*
```
asteiner-swisstopo, line 1013: Typo: "passed in" --> "passed to"

asteiner-swisstopo, line 1013: As mentioned by @boecklic, we might want make it explicit that uploading parts can happen in parallel. Would this be the right place?

spec/transaction/paths.yaml
```diff
@@ -562,6 +714,7 @@ paths:
     put:
       tags:
         - Asset Upload Management
+        - Collection Asset Upload Management
```
asteiner-swisstopo, line 717: For my understanding: Is this tagged with "Collection Asset Upload Management" because you can also directly upload a part to a collection now, instead of only to an item?

asteiner-swisstopo (commented):
Hi @benschs,

As discussed, I have had a look and found mostly typos or unclear writing.

The only more important remarks are
- Should we mention the parallel upload capabilities?
- Is the response "500" intentional or should it be "5XX"?

Note to myself: `openapitransactional.yaml` is generated from the other files (`responses.yaml`, `paths.yaml`, `tags.yaml`).

spec/transaction/components/responses.yaml
benschs: AFAIK the ETag header is only used if a single entity is returned, and does not really make sense if lists are returned.

spec/static/spec/v1/openapitransactional.yaml
```diff
@@ -296,6 +296,16 @@ tags:
       ```
 
       See https://aws.amazon.com/premiumsupport/knowledge-center/data-integrity-s3/ for other examples on how to compute the base64 MD5 of a part.
+  - name: Collection Asset Upload Management
+    description: |
+      Collection Asset files are uploaded via the STAC API using the API requests described in this chapter.
+
+      The flow of the requests is the same as for assets that belong to features, which is described under [Asset Upload Management](#tag/Asset-Upload-Management)
```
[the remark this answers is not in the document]
benschs: I am not sure on the difference, it seems that feature comes from OGC, but STAC uses items. I use them more or less interchangeably.

benschs:
> Hi @benschs,
> 
> As discussed, I have had a look and found mostly typos or unclear writing.
> 
> The only more important remarks are
> 
>     * Should we mention the parallel upload capabilities?
> 
>     * Is the response "500" intentional or should it be "5XX"?
> 
> 
> Note to myself: `openapitransactional.yaml` is generated from the other files (`responses.yaml`, `paths.yaml`, `tags.yaml`).

Thanks for all the feedback, I think I fixed most of the typos.

I added a short sentence about parallel capabilities in the tag description for the multipart upload process.
I also unified the responses to only include "500", although we could even think about removing the 5XX completely from the spec, because AFAICT the user does not gain any information they could act upon.

spec/transaction/paths.yaml
```diff
@@ -107,6 +107,158 @@ paths:
...
+      parameters:
+        - $ref: "../components/parameters.yaml#/components/parameters/collectionId"
+      responses:
+        "200":
+          $ref: "./components/responses.yaml#/components/responses/CollectionAssets"
+        "400":
+          $ref: "../components/responses.yaml#/components/responses/InvalidParameter"
+        "404":
+          $ref: "../components/responses.yaml#/components/responses/NotFound"
+        "500":
```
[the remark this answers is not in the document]
asteiner-swisstopo: Changed to "500" everywhere for consistency.

spec/transaction/paths.yaml
```diff
@@ -789,6 +942,249 @@ paths:
...
+      - $ref: "./components/parameters.yaml#/components/parameters/uploadId"
+    post:
+      tags:
+        - Collection Asset Upload Management
+      summary: Complete multipart upload
+      operationId: completeCollectionAssetMultipartUpload
+      description: >-
+        Complete the multipart upload process. After completion, the collection asset metadata are updated
+        with the new `file:checksum` from the upload and the parts are automatically deleted.
+        The Asset `href` field is also set if it was the first upload.
```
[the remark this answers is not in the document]
asteiner-swisstopo: Typo: "assets" --> "asset's". Same on line 854.

spec/transaction/paths.yaml
asteiner-swisstopo: Now it is changed to

> Specified collection asset multipart upload.

I am not sure I understand what this means. Shouldn't it be something like "Did X successfully"? For example, `completeCollectionAssetMultipartUpload` has description "Asset multipart upload completed successfully."

spec/transaction/paths.yaml
asteiner-swisstopo: Fixed here, thanks. Would you mind correcting also on line 660 where this is probably copied from?

spec/transaction/paths.yaml
asteiner-swisstopo: Ok thanks. Now mentioned in tags.yaml under tag "Asset Upload Management":
> You may also upload multiple parts in parallel.

spec/transaction/paths.yaml
asteiner-swisstopo: Still open, would you mind clarifying please?

asteiner-swisstopo:
> Thanks for all the feedback, I think I fixed most of the typos.
> 
> I added a short sentence about parallel capabilities in the tag description for the multipart upload process. I also unified the responses to only include "500", although we could even think about removing the 5XX completely from the spec, because AFAICT the user does not gain any information they could act upon.

Had another look. Most are fixed, some still open. Would you mind having another look please? :innocent:

spec/transaction/paths.yaml
benschs: Yes. This path is where the parts are uploaded to s3. I decided to add the tag so it shows up in both tag sections.
Depending on the use-case you can either upload a part of an asset, or of a collection asset.

spec/transaction/paths.yaml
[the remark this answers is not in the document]
benschs: Thanks, I missed this. Should be done now. :)

spec/transaction/paths.yaml
asteiner-swisstopo: Ok, fine with me :+1:

asteiner-swisstopo (approved):

PR #456, opened 2024-09-09
PB-756: CRUD API for collection assets

benschs:
Add collection asset endpoints for stac v1. Endpoint to _get list_, _get single,_ _create_, _update_ and _delete_ assets.

Basically duplicated the Asset Views and Serializers for Collection Assets and adapted. The main difference is that there is obviously no item for collection assets, as well as no external url allowed. They also do not have the fields `eo_gsd`, `geoadmin_lang`, `geoadmin_variant`.

The files for views and serializers are now each over 1000 lines of code. Is there a guideline how we might split this into smaller files?

### DIFF
--- a/app/config/urls.py
+++ b/app/config/urls.py
@@ -39,6 +39,7 @@ if settings.DEBUG:
     import debug_toolbar
 
     from stac_api.views_test import TestAssetUpsertHttp500
+    from stac_api.views_test import TestCollectionAssetUpsertHttp500
     from stac_api.views_test import TestCollectionUpsertHttp500
     from stac_api.views_test import TestHttp500
     from stac_api.views_test import TestItemUpsertHttp500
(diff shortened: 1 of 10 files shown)


schtibe (commented):
Looking good, just some minor comments. Maybe it would make sense to also have a e2e test or two

schtibe (approved):

schtibe:
> The files for views and serializers are now each over 1000 lines of code. Is there a guideline how we might split this into smaller files?

In django it would be possible to split those into multiple apps. But I'm not sure if or how this makes sense in this case here. From a URL perspective, everything kinda belongs together.
Otherwise it could maybe make sense to split away the upload part of the API. But also not sure if that's really helpful in terms of file sizes?

boecklic (approved):
Very nice! :ok_hand:

PR #458, opened 2024-09-11
Refactor views and serializers

benschs:
Split views and serializers into multiple files.

Does this split make sense to you?
In the future we could also split the models in the same way if needed.


ltshb (commented):
I think it is a good idea as it will be easier to navigate in the code. However I have just some concerns, see below:
- Doing that we loose the git history blame (`git blame`) of a line of code, of course you can check out before the split and do the blame on the original file, but quite cumbersome
- Somewhere deed in my memory, I have something telling me that we already wanted to do this when we started the project, but did not for a reason that I cannot remember, did you test locally your changes ?

Of course if we do this we need to do it for all serializers and views, as I'm not a fan of having module like this `serializers.serializers`. So for some serializers that don't make sense to have them in single files we should put them in something like this `serializers.general`

benschs:
Thanks for the feedback

> Doing that we loose the git history blame (`git blame`) of a line of code, of course you can check out before the split and do the blame on the original file, but quite cumbersome

I am not sure how/if we can solve the git blame issue completely. There are options to use `git blame` with `-C` and/or `-M` to try and take line copy/movement into consideration.

> Somewhere deed in my memory, I have something telling me that we already wanted to do this when we started the project, but did not for a reason that I cannot remember, did you test locally your changes ?

I tested the changes locally, also using the e2e tests. Everything seems to work as expected. Maybe @hansmannj was also involved in previous discussions and might know about issues with this approach?

> Of course if we do this we need to do it for all serializers and views, as I'm not a fan of having module like this serializers.serializers. So for some serializers that don't make sense to have them in single files we should put them in something like this serializers.general

I agree about the "double" name issue and will adapt it.

app/stac_api/urls.py
```diff
@@ -4,28 +4,28 @@
...
-from stac_api.views import recalculate_extent
+from stac_api.views.collection import CollectionAssetDetail
+from stac_api.views.collection import CollectionAssetsList
+from stac_api.views.collection import CollectionDetail
+from stac_api.views.collection import CollectionList
+from stac_api.views.item import AssetDetail
+from stac_api.views.item import AssetsList
+from stac_api.views.item import ItemDetail
+from stac_api.views.item import ItemsList
+from stac_api.views.views import AssetUploadAbort
```
boecklic, line 15: Same comment as @ltshb regarding serializers: use `views.assetupload` or sth instead of `views.views`

boecklic (approved):
Very good idea! Can we keep somehow the git history if we `git-cp` the files and then remove the not needed parts? Or have you done it that way anyway @benschs ?

ltshb:
@benschs if there is no issues and if you move all view/serializers in their own file (some might be kept in a single called `general` or the like) then you are all good to go :-) of course the git blame is not so nice but that should not block us to do improvements 😉 maybe try the @boecklic  suggestions to keep the git blame.

benschs:
I did not use git-cp. After testing though it does not work as nice. The initial copy of the file is fine, but then after deleting hundres of lines it overwrites everything again to myself.
The best option seems to be to use the [-C](https://git-scm.com/docs/git-blame#Documentation/git-blame.txt--Cltnumgt) option with git blame. If needed this can also be added to vs code gitlens by adding `"gitlens.advanced.blame.customArguments": ["-C"]` to the `settings.json`.

hansmannj:
> Thanks for the feedback
> 
> > Doing that we loose the git history blame (`git blame`) of a line of code, of course you can check out before the split and do the blame on the original file, but quite cumbersome
> 
> I am not sure how/if we can solve the git blame issue completely. There are options to use `git blame` with `-C` and/or `-M` to try and take line copy/movement into consideration.
> 
> > Somewhere deed in my memory, I have something telling me that we already wanted to do this when we started the project, but did not for a reason that I cannot remember, did you test locally your changes ?
> 
> I tested the changes locally, also using the e2e tests. Everything seems to work as expected. Maybe @hansmannj was also involved in previous discussions and might know about issues with this approach?
> 
> > Of course if we do this we need to do it for all serializers and views, as I'm not a fan of having module like this serializers.serializers. So for some serializers that don't make sense to have them in single files we should put them in something like this serializers.general
> 
> I agree about the "double" name issue and will adapt it.

Yes, I also slightly remember that we discussed it, but unfortunately, same as @ltshb, I cannot remember the reasoning behind not having done it :see_no_evil: 
I'll try to check, if we wrote it down somewhere in an ADR. But that might not be the case.
Anyways, we should go ahead and will surely notice if there are issues and in worst case could still revert, if really needed.

PR #459, opened 2024-09-12
PB-932: Avoid UTF-8 truncation in logging

ltshb:
The request body is in utf-8 which needs either 1 byte or 2 bytes per character
depending on the character, so we need to decode first before truncating otherwise
we might truncate in the middle of a utf-8 character and it will break the decode.


ltshb:
@boecklic should I do this fix on develop or on master and deploy direclty ?

boecklic:
> @boecklic should I do this fix on develop or on master and deploy direclty ?

If you don't mind I'd create a normal minor release and deploy all the way to prod

benschs (approved):

PR #462, opened 2024-09-24
PB-796: remove CommonMiddlewareWithInternalRedirect.

adk-swisstopo:
In https://github.com/geoadmin/service-stac/pull/440 we added CommonMiddlewareWithInternalRedirect to respond to URLs that don't end with a trailing slash without user-visible redirection. This was primarily done to handle a limitation of AWS API Gateway v1. We eventually decided to use API Gateway v2 which does not have this limitation.

Hence, we remove this code that is not useful and update the tests accordingly. This is basically a revert of https://github.com/geoadmin/service-stac/pull/440 although we keep some of the test refactoring/improvements.


boecklic (approved):
:+1: thanks for cleaning up again!

PR #463, opened 2024-10-01
PB-950 Switch to using ssm params - #minor

schtibe:
Use the params from the SSM store instead of gopass for instantiating the database.

(Since we already have the databases set up this is done rather for completeness sake than actual usage. We probably won't actually use these params in the near future. Also, the Database will be [managed by TF](https://jira.swisstopo.ch/browse/PB-986))

See https://github.com/geoadmin/infra-terraform-bgdi/pull/410 for more context.


boecklic (approved):
Thanks! I guess these credentials won't be necessary anymore once https://jira.swisstopo.ch/browse/PB-986 is tackled right?

Pls remove the secrets from gopass once this is merged.

PR #464, opened 2024-10-03
PB-917: Save file size on asset object

LukasJoss:


app/tests/tests_10/test_asset_upload_endpoint.py
```diff
@@ -497,6 +501,8 @@ def test_asset_upload_gzip(self):
         self.assertS3ObjectExists(key)
         obj = self.get_s3_object(key)
         self.assertS3ObjectContentEncoding(obj, key, encoding='gzip')
+        self.asset.refresh_from_db()
+        self.assertEqual(size, self.asset.file_size)
```
LukasJoss, line 505: There seems to be a slight discrepancy in the file size in this case. I'm not sure if this is expected behavior or there is an error in the code.

boecklic (commented):
Very nice! Can you comment on performance of the pgtriggers? We could verywell think about an asynchronous calculation here as well. The aggregated information is currently purely informational for ourselves, so no need for realtime calculation.

benschs (approved):
In case the triggers are slow, maybe it would also be an option to just subtract the old and add the the new file size to the parent? Then you don't need aggregate each time. The aggregation would only be required on the initial migration when adding the new fields.

LukasJoss:
> In case the triggers are slow, maybe it would also be an option to just subtract the old and add the the new file size to the parent? Then you don't need aggregate each time. The aggregation would only be required on the initial migration when adding the new fields.

@benschs  @boecklic  I have updated the triggers now so that it works as suggested, the only potential problem is that errors would not be detected if someone alters the data directly in the database.

benschs (approved):
Looks good, thanks.
Are there already plans how to migrate the initial file size values for the existing files?

app/stac_api/management/commands/update_asset_file_size.py
```diff
@@ -0,0 +1,59 @@
...
+
+        self.print_success('update file size for assets')
+        assets = Asset.objects.filter(file_size=0).all()
+        for asset in assets:
+            key = SharedAssetUploadBase.get_path(None, asset)
+            try:
+                file_size = s3.head_object(Bucket=bucket, Key=key)['ContentLength']
+                asset.file_size = file_size
+                asset.save()
+            except ClientError:
```
boecklic, line 35: this will fail for all assets that are stored on the managed bucket. I'd prepare the command to have two clients, legacy and managed and then check the correct bucket based on the info in the asset object

app/stac_api/admin.py
```diff
@@ -213,6 +238,12 @@ class Media:
     list_display = ['name', 'collection', 'collection_published']
     list_filter = [CollectionFilterForItems]
 
+    #helper function which displays the bytes in human-readable format
+    def displayed_total_data_size(self, instance):
+        return filesizeformat(instance.total_data_size)
+
+    displayed_total_data_size.short_description = 'Total data size'
```
boecklic, line 245: pls add a unit here (`B`, `KB`?)

boecklic (approved):

app/stac_api/admin.py
LukasJoss: I dont quite understand why that is necessary, doesn't the unit from the filesizeformat suffice?
Screenshot from 2024-10-22 11-41-21

PR #469, opened 2024-11-04
PB-1071: add new mime types for Zipped LAS and Cloud Optimized Point

ltclm:
Cloud (COPC)


ltshb (approved):
Looks good as far as I can tell

PR #472, opened 2024-11-20
PB-224: Added a query parameter on the /collections route to filter by provider

sami-nouidri-swisstopo:
The response from the /collections route can now be filtered by provider. The get method of COllectionList has also been re-factored to reduce code repetition.

In order to achieve this, filter_queryset has been overwritten in CollectionList to read the provider argument and filter the dataset accordingly.


benschs (approved):
:+1:

PR #476, opened 2024-11-26
PB-1210: Set default file path for assets - #patch

benschs:
When assets are created via admin GUI there is no file path set. If the file is then uploaded via API using the multipart upload, the file path will never be set on the asset.

This aligns the admin GUI behaviour with that of creating an asset via API PUT.

### DIFF
--- a/app/stac_api/models.py
+++ b/app/stac_api/models.py
@@ -628,6 +628,19 @@ class AssetBase(models.Model):
     def __str__(self):
         return self.name
 
+    def get_asset_path(self):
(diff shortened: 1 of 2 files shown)


hansmannj (commented):
Does that mean, if an asset is created via admin GUI, but no file is yet uploaded, there will already be a link to a non-existing file?

ltshb:
@benschs @hansmannj @boecklic This has been meanwhile solved (probably with the big asset upload) but now with v1.33.0-beta.14 the href is always set even if no asset has been uploaded yet.

PR #477, opened 2024-11-27
PB-1073 Implement forecast extension

asteiner-swisstopo:
This adds fields of the [forecast extension](https://github.com/stac-extensions/forecast) to the "properties" of an Item in [our STAC API](https://sys-data.dev.bgdi.ch/api/stac/static/spec/v1/apitransactional.html):
- Fields that are already part of [the existing forecast extension](https://github.com/stac-extensions/forecast)
   - forecast:reference_datetime
   :information_source: A regular datetime
   - forecast:horizon
   :information_source: A duration in ISO 8601 format (e.g. "P3DT6H")
   - forecast:duration
   :information_source: A duration in ISO 8601 format (e.g. "P3DT6H")
- Fields that are proposed to be added in [the PR](https://github.com/stac-extensions/forecast/pull/12):
   - forecast:param
   :information_source: Free text
   - forecast:mode
   :information_source: Allowed values: "ctrl" and "perturb"

An example response for `getFeature`:

```json
{
  "id": "smr200-200-4-2019",
  "links": [...],
  "assets": {...},
  "bbox": [...],
  "geometry": {...},
  "properties": {
    "start_datetime": "2018-02-12T23:20:50Z",
    "end_datetime": "2018-02-12T23:20:50Z",
    "expires": "2018-02-12T23:20:50Z",
    "title": "Feature title",
    "forecast:reference_datetime": "2018-02-12T23:20:50Z",
    "forecast:horizon": "P3DT6H",
    "forecast:duration": "P3DT6H",
    "forecast:param": "T",
    "forecast:mode": "ctrl"
  },
  "stac_version": "0.9.0",
  "type": "Feature"
}
```

Notes on the implementation:
1. Django:
    1. The help texts are copy-pasted from [the forecast stac-extension README](https://github.com/stac-extensions/forecast).
    2. Fields `forecast:horizon` and `forecast:duration` are special because they are encoded in a [ISO 8601 format](https://en.wikipedia.org/wiki/ISO_8601#Durations), for example "P3DT6H" for "3 days, 6 hours".
        1. In the Django model, they are represented as a [django.db.models.DurationField](https://docs.djangoproject.com/en/5.1/ref/models/fields/#durationfield). Using a pure text (CharField) would have been easier but a DurationField is going to facilitate filtering later (see [PB-1169](https://jira.swisstopo.ch/browse/PB-1169)).
        3. For validation, serialization and deserialization, the fields are represented by a custom DRF Serializer (`IsoDurationField`). This translates the ISO 8601 format to a Python datetime and vice versa by using [django.utils.dateparse.parse_duration](https://docs.djangoproject.com/en/5.1/ref/utils/#django.utils.dateparse.parse_duration) and [django.utils.duration.duration_iso_string](https://github.com/django/django/blob/main/django/utils/duration.py#L31). We considered using the dedicated Python package [isodate](https://github.com/gweis/isodate) but considered the dependency too much for the limited gain.
        :question: `parse_duration` also accepts other formats than ISO 8601: Something like "DD HH:MM:SS.uuuuuu", "DD HH:MM:SS,uuuuuu" or "3 days 04:05:06" (PostgreSQL). So when validating the input, not only the ISO 8601 format would pass. I assume that this is acceptable? 
        :question: `duration_iso_string` is not mentioned in [the django.utils docs](https://docs.djangoproject.com/en/5.1/ref/utils/). Not sure why... I assume it's still safe to use this?
   3.  To map the model field names to their counterparts in the payload ("forecast_param" --> "forecast:param"), the [`to_internal_value` and `to_representation` methods](https://www.django-rest-framework.org/api-guide/relations/#custom-relational-fields) were overwritten in the ItemsPropertiesSerializer.
   5. Admin UI: The forecast fields are grouped in a "Forecast" section.
2. OpenAPI specs: Ignore the files in `spec/static`, they are stitched together from `spec/components/schemas.yaml` and `spec/transaction/paths.yaml` by `make build-specs` (using yq).
   1. Both the [getFeature](https://sys-data.dev.bgdi.ch/api/stac/static/spec/v1/apitransactional.html#tag/Data/operation/getFeature) (Data --> "Fetch a single feature") and the [putFeature](https://sys-data.dev.bgdi.ch/api/stac/static/spec/v1/apitransactional.html#tag/Data-Management/operation/putFeature) (Data Management  --> "Update or create a feature") docs should be up-to-date both regarding the specs (response/request schema) and the examples (right pane).
   2. A bug was discovered in the course of expanding these docs: The "properties" have disappeared in the docs for the request body schema for `putFeature` of v1.0. Compare the request bodies in production: [v0.9](https://sys-data.dev.bgdi.ch/api/stac/static/spec/v0.9/apitransactional.html#tag/Data-Management/operation/putFeature) vs [v1.0](https://sys-data.dev.bgdi.ch/api/stac/static/spec/v1/apitransactional.html#tag/Data-Management/operation/putFeature). This was introduced by accident when updating the OpenAPI validator, see [PB-687](https://jira.swisstopo.ch/browse/PB-687) and the exact commit [613f084](https://github.com/geoadmin/service-stac/commit/613f0842de6e478ff52f8e2ce8cfff2ef5cf32dc): By setting all fields to `readOnly: true` they disappeared from the request body of `putFeature`. For more details on the fix see commit [dd31133](https://github.com/geoadmin/service-stac/commit/dd311331d1233a513d1cfffc98527050a37604f6) in this PR.
   :point_up: CC @adk-swisstopo, @ltshb
   :question: With this fix, the following fields in "properties" are still read-only, so hidden in `updateFeature`: `created`, `updated` and `expires`. This is in line with what we see in [v0.9](https://sys-data.dev.bgdi.ch/api/stac/static/spec/v0.9/apitransactional.html#tag/Data-Management/operation/putFeature). For `created` and `updated`, this sounds reasonable: They are set automatically and should not be defined in a PUT request. But for `expires` (description: "Date time when this presigned URL expires and is not valid") this might be a bug as well: In the DRF validator [the field is examined](https://github.com/geoadmin/service-stac/blob/develop/app/stac_api/validators.py#L428), so I suppose it is writeable. Should the OpenAPI specs be fixed here too?

## Screenshots

Admin UI
pb-1073-forecast-extension

OpenAPI getFeature
pb-1073-openapi-getFeature-forecast-fields

OpenAPI updateFeature
pb-1073-openapi-putFeature-forecast-fields


app/stac_api/serializers/utils.py
```diff
@@ -339,3 +343,21 @@ def to_representation(self, value):
...
+class IsoDurationField(serializers.Field):
+    '''Handles duration in the ISO 8601 format like "P3DT6H"'''
+
+    def to_internal_value(self, data):
+        '''Convert from ISO 8601 (e.g. "P3DT1H") to Python's timedelta'''
+        internal = parse_duration(data)
+        if internal is None:
+            raise serializers.ValidationError(
+                code="payload",
+                detail={'href': _("Duration doesn't match ISO 8601 format")}
```
benschs, line 357: I guess is should be `'duration'` not `'href'`?

spec/components/schemas.yaml
```diff
@@ -1073,7 +1077,9 @@ components:
         end_datetime:
           $ref: "#/components/schemas/datetime"
         expires:
-          $ref: "#/components/schemas/datetime"
+          readOnly: true
```
benschs, line 1080: `expires` is should actually be writable. It can optionally be set to delete the data in the future.

https://jira.swisstopo.ch/browse/PB-847

benschs (approved):
Cool. Thanks for the work.

Small side note for the future: The README mentions to try and make only 1 Migration per PR (if possible) https://github.com/geoadmin/service-stac?tab=readme-ov-file#migrate-db-with-django

app/stac_api/serializers/utils.py
asteiner-swisstopo: Yes, will fix. Lack of understanding from my part.

While the actual field name would be better, we agreed that only having the error message is good enough. Having `duration` would only be confusing because there is no such field.

spec/components/schemas.yaml
asteiner-swisstopo: Alright, will adapt. Thanks!

This answers my question from the PR description:
> ❓ With this fix, the following fields in "properties" are still read-only, so hidden in updateFeature: created, updated and expires. This is in line with what we see in [v0.9](https://sys-data.dev.bgdi.ch/api/stac/static/spec/v0.9/apitransactional.html#tag/Data-Management/operation/putFeature). For created and updated, this sounds reasonable: They are set automatically and should not be defined in a PUT request. But for expires (description: "Date time when this presigned URL expires and is not valid") this might be a bug as well: In the DRF validator [the field is examined](https://github.com/geoadmin/service-stac/blob/develop/app/stac_api/validators.py#L428), so I suppose it is writeable. Should the OpenAPI specs be fixed here too?

boecklic (approved):
Nice!

PR #481, opened 2024-12-04
Make session tracking work with API Gateway.

adk-swisstopo:
As per commits.

The main change in this PR is to `ApiGatewayMiddleware` which is updated to correctly handle the headers that API Gateway sends. In particular in the case where the user is already authenticated via a Django session cookie but where the API Gateway `Geoadmin-Username` header is invalid.

This also refactors and adds tests as well as tweaking session cookie parameters.

This only applies to the debug build. I have verified this works in the dev environment.


benschs (approved):

boecklic (approved):
Very nice! And thanks for the proposal to enhance Django.

PR #482, opened 2024-12-05
PB-1009: simplify ApiGatewayMiddleware.

adk-swisstopo:
As per discussion in https://code.djangoproject.com/ticket/35971

Instead of copy/pasting `process_request` from the parent class, we can just drop the `Geoadmin-Username` header from `request.META` if `Geoadmin-Authenticated` is not true before calling `process_request` from the parent class.


benschs (approved):
Cool

PR #483, opened 2024-12-05
PB-1157 Add hreflang to Links

schtibe:
Adding the extension field `hreflang` to the fields of the Links.


app/stac_api/models.py
```diff
@@ -153,13 +154,23 @@ class Link(models.Model):
...
 
     def __str__(self):
         return f'{self.rel}: {self.href}'
 
+    def save(self, *args, **kwargs) -> None:
+        """Validate the hreflang"""
+        self.full_clean()
+
+        if self.hreflang is not None and self.hreflang != '' and not tags.check(self.hreflang):
+            raise ValidationError(_("Invalid value for hreflang"))
```
asteiner-swisstopo, line 170: Apparently the `language-tags` package [gives quite detailed error messages](https://github.com/OnroerendErfgoed/language-tags/blob/develop/language_tags/Tag.py#L347) (for example "The private-use subtag 'abcdefhijk' is too long."). Is it on purpose that these are not returned here? Looks good enough for me as it is, just to double check.

spec/components/schemas.yaml
```diff
@@ -1237,6 +1237,10 @@ components:
             - GET
             - POST
           type: string
+        hreflang:
+          description: The language of the link target
+          example: de-CH
+          type: string
```
asteiner-swisstopo, line 1243: `hreflang` is part of [STAC extension "language"](https://github.com/stac-extensions/language?tab=readme-ov-file#fields-for-links-and-assets). Do we need to point that out somewhere in the documentation?

app/stac_api/admin.py
```diff
@@ -66,7 +66,17 @@ class ProviderInline(admin.TabularInline):
     }
 
 
-class CollectionLinkInline(admin.TabularInline):
+class LinkInline(admin.TabularInline):
+
+    def formfield_for_dbfield(self, db_field, request, **kwargs):
+        # make the hreflang field a bit shorter
```
asteiner-swisstopo, line 72: Maybe explain why it is useful to have this attribute name smaller/short? Not clear to me what this would change and why without seeing it.

asteiner-swisstopo (approved):
Nice! I like how test coverage was improved with this. A few side comments mainly to possibly improve maintainability.

app/tests/tests_10/test_collections_endpoint.py
```diff
@@ -596,3 +598,85 @@ def test_unauthorized_collection_delete(self):
...
+
+        path = f'/{STAC_BASE_V}/collections/{self.collection.name}'
+        response = self.client.put(path, data=data, content_type="application/json")
+
+        self.assertEqual(response.status_code, 200)
+
+        link = CollectionLink.objects.last()
+        # for the type checker here. But it's of course also valuable for the test
+        # since pytest should be the future, maybe I can afford this?
+        assert link is not None
```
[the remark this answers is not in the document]
schtibe: Through the lens of typing, `link` is possibly None. I need to tell the type checker that it shouldn't be, otherwise it'll say that "hreflang" is not a known attribute of "None". `self.assertNotEqual` can't do that. Now elegantly, `assert` both satisfies the type checker as well as it provides an assertion in the context of testing (at least if pytest is used).

Now we probably don't want to mix the way we assert things in testing, that's why I'm a bit hesitant here. On the other hand, it annoys me that I can't use this elegant solution 😄

Of course, static type checking isn't a requirement in service-stac but I thought we might as well start with that...

spec/components/schemas.yaml
schtibe: Do you mean for the consumer or rather here for the coder?

app/stac_api/models.py
schtibe: Hmm no actually I haven't thought about that

schtibe: Included the error messages :+1:

app/tests/tests_10/test_collections_endpoint.py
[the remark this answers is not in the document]
asteiner-swisstopo: Ok thanks.

No strong opinion, either works for me. I would just suggest to adapt the comment to something like
```py
# Check for None with `assert` because `self.assertNotEqual` is not understood by the type checker.
```

spec/components/schemas.yaml
asteiner-swisstopo: Possibly both. I wondered whether it needs to be made transparent that it's not vanilla STAC but includes such extensions. For the forecast extension, it was at least a bit visible through the "forecast:" in the field label (could be better though...).

Maybe the reference to the "language" extension is not needed in the docs visible to the consumer, not sure how this is handled. Having a reference to the extension in the code, however, might simplify looking up what that is in the future for a coder.

What do you think?

boecklic: @asteiner-swisstopo @schtibe as noted here https://github.com/radiantearth/stac-spec/blob/master/extensions/README.md#using-extensions stac extenstions should be added to the `stac_extensions` list in principle. I'm unsure for the `hreflang` since we just implement one field and not the whole extension if we should add it or not, I'd rather not. For forecast we can think about it if it gets accepted.

app/tests/tests_10/test_collections_endpoint.py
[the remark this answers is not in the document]
schtibe: Agreed, thanks for the suggestion!!

PR #484, opened 2024-12-09
PB-1091 Add count to size updater, exclude is_external

schtibe:
Make it possible to only update a chunk of assets at once, so the command will hopefully stop timing out. I'm attempting to solve the problem by re-running the command a couple of times, maybe in a temporary cronjob.

Also excluded the external assets and improved the output a bit.

Running this command every 60 seconds with 1000 assets at once should result in a completion after 30h. I will have to figure out a sensible chunk size.

Of course it's 1000 assets and 1000 collection assets, but there are only 2 CollectionAssets on prod 😂

### DIFF
--- a/app/stac_api/management/commands/update_asset_file_size.py
+++ b/app/stac_api/management/commands/update_asset_file_size.py
@@ -15,6 +15,10 @@ from stac_api.views.upload import SharedAssetUploadBase
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


boecklic (approved):
nice, thanks!

schtibe:
See https://github.com/geoadmin/infra-kubernetes/pull/681. I *think* it works, but can't say for sure.

PR #487, opened 2024-12-16
PB-1293 Add docs for endpoint bulkUpsertItems

asteiner-swisstopo:
This adds a section for a new endpoint in the OpenAPI docs:
- endpoint: `POST collections/{collectionId}`
- docs on local web server: http://0.0.0.0:8090/apitransactional.html#tag/Data-Management/operation/bulkUpsertItems

Some notes:
1. The term "bulk" was used instead of "batch" after looking up the difference in [a blog post](https://www.codementor.io/blog/batch-endpoints-6olbjay1hd):
   > It should be noted that some people, such as [John Apostolidis](http://apostolidis.me/bulk-operations/), make a hard distinction between bulk processing (applying the same operation to multiple entries) and batch processing (applying potentially different operations to different entries). While it makes sense to have this distinction, in reality, the two ideas are often conflated and used interchangeably. "Batch" is often regarded as the more general term (processing batches of requests or batches of data), and "bulk" as a subset of batch (batching data, but not operations).
3. Like the other endpoints, the public text uses the term "features", not "items". Internally, we still call it "items" (e.g. `bulkUpsertItems`).
4. The non-success response codes (403, 404, 500) are copied from a similar endpoint. Not sure how explicitly every possible case that is ultimately covered by DRF has to be listed here. Also, not sure if 403 should already by there if there is no auth yet?
5. `Idempotency-Key`: We won't check for this in a first round but we agreed to already place it there. A link to [the IETF draft for the parameter](https://datatracker.ietf.org/doc/draft-ietf-httpapi-idempotency-key-header/) was added. Note that the draft is marked as "expired", so not sure if this is really the best source to cite. Or if this is really the right param ID to use?
6. The generated file `openapi.yaml` also has changes even though the files/parts it derives from have no changes. The changes are only reshuffling content, the displayed text should be the same. Possibly a missing update of the file in a different PR?

To explore the updated docs in your browser:
1. Build the specs:
   ```bash
   make build-specs
   ```
7. Run the web server:
    ```
    make serve-specs
   ```
8. Open the page in your web browser:
    http://0.0.0.0:8090/apitransactional.html#tag/Data-Management/operation/bulkUpsertItems

In a screenshot:
pb-1293-openapi-bulkupsertitems-v1

CC @benschs @schtibe

### DIFF
--- a/spec/transaction/components/schemas.yaml
+++ b/spec/transaction/components/schemas.yaml
@@ -57,6 +57,13 @@ components:
                   title: Licence for the free geodata of the Federal Office of Topography swisstopo
                 - href: https://www.geocat.ch/geonetwork/srv/eng/catalog.search#/metadata/4c8c7c58-61c7-4a61-8e7a-6ffb95d183df
                   rel: describedby
+    collectionUpsertItems:
+      type: object
+      properties:
+        items:
+          type: array
+          items:
+            $ref: "../../components/schemas.yaml#/components/schemas/item"
     assetWrite:
       title: Asset
       description: The `property name` defines the ID of the Asset.
--- a/spec/transaction/components/schemas.yaml
+++ b/spec/transaction/components/schemas.yaml
@@ -60,7 +60,7 @@ components:
     collectionUpsertItems:
       type: object
       properties:
-        items:
+        features:
           type: array
           items:
             $ref: "../../components/schemas.yaml#/components/schemas/item"
--- a/spec/static/spec/v1/openapi.yaml
+++ b/spec/static/spec/v1/openapi.yaml
@@ -1798,6 +1798,121 @@ components:
       type: string
       format: date-time
       readOnly: true
+  parameters:
+    assetQuery:
+      description: >-
+        Query for properties in assets (e.g. mediatype). Use the JSON form of the assetQueryFilter used in POST.
+      in: query
+      name: assetQuery
+      required: false
+      schema:
+        type: string
+    bbox:
+      explode: false
+      in: query
+      name: bbox
+      required: false
+      schema:
+        $ref: "#/components/schemas/bbox"
+      style: form
+    collectionId:
+      description: Local identifier of a collection
+      in: path
+      name: collectionId
+      required: true
+      schema:
+        type: string
+    collectionsArray:
+      explode: false
+      in: query
+      name: collections
+      required: false
+      schema:
+        $ref: "#/components/schemas/collectionsArray"
+    datetime:
+      explode: false
+      in: query
+      name: datetime
+      required: false
+      schema:
+        $ref: "#/components/schemas/datetimeQuery"
+      example: 2018-02-12T00%3A00%3A00Z%2F2018-03-18T12%3A31%3A12Z
+      style: form
+    featureId:
+      description: Local identifier of a feature
+      in: path
+      name: featureId
+      required: true
+      schema:
+        type: string
+    assetObjectHref:
+      name: assetObjectHref
+      in: path
+      description: Full URL to asset object including protocol, host and path
+      required: true
+      schema:
+        type: string
+    ids:
+      description: >-
+        Array of Item ids to return. All other filter parameters that further restrict the number of search results are ignored
+      explode: false
+      in: query
+      name: ids
+      required: false
+      schema:
+        $ref: "#/components/schemas/ids"
+    limit:
+      explode: false
+      in: query
+      name: limit
+      required: false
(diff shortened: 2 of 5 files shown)


benschs:
For me `POST /collections/{collectionId}` sounds like we create a collection. Maybe `POST /collections/{collectionId}/bulkitems` is better?

Regarding idempotency, if we use POST i think it is fine to use the idempotency-key header as I haven't found an alternative. But as we upsert anyway, can we not just use PUT and be idempotent by design?

asteiner-swisstopo:
Discussed in a call with @benschs:
1. :white_check_mark:  Name of the endpoint: `POST /collections/{collectionId}` is ok.
   --> To create a collection, you would actually do a `POST /collections`, not `POST /collections/{collectionId}`. `/bulkitems` is a bit more self-explaining on what it adds. Ultimately no big difference, `POST /collections/{collectionId}` can be good enough.
2. :x: Idempotency: Better use PUT. Consider an extra header parameter `Request-Id`.
   --> PUT would indeed provide idempotency without adding additional fields. Most importantly, it would be more consistent with other existing operations, for example ["Update or create a feature"](https://sys-data.dev.bgdi.ch/api/stac/static/spec/v1/apitransactional.html#tag/Data-Management/operation/putFeature) which uses PUT. To enable identification of request for debugging, we may want to use a custom header parameter like "Request-Id". Using `idempotency-key` for this only creates confusion and it would not allow to distiguish retries of the same operation.

How do you see it, @boecklic?

boecklic:
> Discussed in a call with @benschs:
> 
> 1. ✅  Name of the endpoint: `POST /collections/{collectionId}` is ok.
>    --> To create a collection, you would actually do a `POST /collections`, not `POST /collections/{collectionId}`. `/bulkitems` is a bit more self-explaining on what it adds. Ultimately no big difference, `POST /collections/{collectionId}` can be good enough.
> 2. ❌ Idempotency: Better use PUT. Consider an extra header parameter `Request-Id`.
>    --> PUT would indeed provide idempotency without adding additional fields. Most importantly, it would be more consistent with other existing operations, for example ["Update or create a feature"](https://sys-data.dev.bgdi.ch/api/stac/static/spec/v1/apitransactional.html#tag/Data-Management/operation/putFeature) which uses PUT. To enable identification of request for debugging, we may want to use a custom header parameter like "Request-Id". Using `idempotency-key` for this only creates confusion and it would not allow to distiguish retries of the same operation.
> 
> How do you see it, @boecklic?

Hej @asteiner-swisstopo @benschs : regarding idempotency: it's not so much the http method that makes a call idempotent, I think just changing `POST` to `PUT` doesn't change anything in our case. Idempotency states that you can issue the exact same call twice without changing resources, you can achieve this if you can uniquely relate a call to resources. If you have a `PUT` on e.g. `/orders/{order_id}` and the payload contains all fields, that call is idempotent since issuing the call twice doesn't change anything (except for maybe timestamps, if that is relevant, then you would have to add other measures to that PUT call as well). In our case where we `POST` a list, objects might be created or updated, it's not that clear. You could argue that all objects have a clear identifier provided by the client, and that therefore it doesn't matter if the POST call is issued twice (except for as mentioned above timestamps). However, it might still be useful to be able to identify multiple calls which are basically the same from client perspective. I think a request-id is mostly issued by the API, not the client, to be able to group related calls/logs in a microservice architecture (we'll look at this soon with Berni, buzzword "tracing"), and we already have it: Cloudfront creates a unique key for each call and adds it as header to the origin (`x-amz-cf-id`). So two calls with the same idempotency key should have two different request-ids, since it's two different calls from service perspective, but just one from client perspective.
I think for calls like the one we're discussing where the payload is a list of resources it's useful to have an idempotency-key header for debugging, since we don't have any information on the payload and have no other means to see whether the same call is issued several times.

spec/transaction/components/schemas.yaml
```diff
@@ -57,6 +57,13 @@ components:
                   title: Licence for the free geodata of the Federal Office of Topography swisstopo
                 - href: https://www.geocat.ch/geonetwork/srv/eng/catalog.search#/metadata/4c8c7c58-61c7-4a61-8e7a-6ffb95d183df
                   rel: describedby
+    collectionUpsertItems:
+      type: object
+      properties:
+        items:
```
boecklic, line 63: after discussion: rename to `features` since in the GET payload `features` is used as well (hence it's consistently confusing since in the path we have `../items`)

boecklic (approved):

PR #488, opened 2024-12-19
fixed deleting expired items that have pending asset uploads

boecklic:


app/stac_api/management/commands/dummy_asset_upload.py
```diff
@@ -0,0 +1,117 @@
...
+from stac_api.utils import get_sha256_multihash
+
+logger = logging.getLogger(__name__)
+
+
+class DummyAssetUploadHandler(CommandHandler):
+
+    def __init__(self, *args, **kwargs):
+
+        s3_bucket = kwargs.pop('s3_bucket', AVAILABLE_S3_BUCKETS.legacy)
```
schtibe, line 22: So this seems as a preparation for chosing another bucket than legacy? I don't quite follow where this `kwargs` would be set, as there is no argument parse option

app/stac_api/management/commands/dummy_asset_upload.py
```diff
@@ -0,0 +1,117 @@
...
+            f"{settings.AWS_SETTINGS[s3_bucket.name]['S3_BUCKET_NAME']}"
+        )
+        super().__init__(*args, **kwargs)
+        self.uploader = MultipartUpload(s3_bucket)
+
+    def start(self):
+        self.print_success(f"Starting upload for {self.options['asset_id']}")
+        asset = Asset.objects.filter(name=self.options['asset_id']).first()
+
+        key = f"{asset.item.collection.name}/{asset.item.name}/{asset.name}"
```
schtibe, line 34: FYI there's a function for that: `def get_asset_path(item, asset_name):` in `utils.py`

app/stac_api/management/commands/dummy_asset_upload.py
```diff
@@ -0,0 +1,117 @@
...
+            content_encoding=None
+        )
+        AssetUpload.objects.get_or_create(
+            asset=asset, upload_id=upload_id, number_parts=1, md5_parts={"1": "ASDF"}
+        )
+        self.print_success(f" -- upload_id: {upload_id}")
+
+    def list(self):
+        for upload in AssetUpload.objects.filter(status=BaseAssetUpload.Status.IN_PROGRESS):
+            print(f"> {upload.upload_id} (asset: {upload.asset.name})")
```
schtibe, line 54: Maybe also `self.print`

schtibe (approved):
Looks good! Some minor details, nothing that is blocking.

app/stac_api/management/commands/dummy_asset_upload.py
```diff
@@ -0,0 +1,117 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class DummyAssetUploadHandler(CommandHandler):
+
+    def __init__(self, *args, **kwargs):
+
+        s3_bucket = kwargs.pop('s3_bucket', AVAILABLE_S3_BUCKETS.legacy)
+        print(
```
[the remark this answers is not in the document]
boecklic: Yeah, this is probably purely used for dev purpose, so it doesn't really matter. But I've changed it.

app/stac_api/management/commands/dummy_asset_upload.py
boecklic: thanks for the hint, using the function now

app/stac_api/management/commands/dummy_asset_upload.py
boecklic: Currently the command is not meant to really upload sth, it's meant to just allow to manipulate AssetUpload objects. I need a valid bucket configuration to initiate the `MultipartUpload` class, but it doesn't really matter which one. I'll add a comment but leave it like this for the moment if you're ok with that.

PR #490, opened 2025-01-06
PB-986 Cleanup obsolete SQL management scripts

schtibe:
The SQL scripts aren't useful anymore. Some don't even work, as the DB structure has changed. 
The Database is now managed via Terraform, thus there's no need to have DB management scripts around.
Also removed the secrets.yml file, as this was necessary for the SQL scripts


scripts/sql/samples/load_random_data.sql
boecklic, line 1: Should we 'resque' this one somehow? E.g. transfer it into a mgmt command (executing raw sql), as existing commands to create dummy data are really slow for larger amount of objects (this sql script was created at the time in context of perf tests).

boecklic (commented):
Yay! :scissors:

boecklic (approved):

PR #492, opened 2025-01-09
Remove csrf check for api gateway authentication

benschs:
Issue is that SessionAuthentication always checks csrf, if we don't use another authentication class. Before we were using the TokenAuthentication that had the Authoritation header set, so no csrf check was done. Now with the api gateway we remove the Authorization Header that would solve csrf, so django still tries to check for csrf tokens and fails.

Do you see issues if we always remove csrf check for users authenticated by apigateway? Otherwise we would need to find some other way to differentiate between "normal" users and "machine" users.

### DIFF
--- a/app/middleware/apigw.py
+++ b/app/middleware/apigw.py
@@ -12,10 +12,10 @@ class ApiGatewayMiddleware(PersistentRemoteUserMiddleware):
         whether it was able to authenticate the user. If it could not
         authenticate the user, the value of the header as seen on the wire is a
         single whitespace. An hexdump looks like this:
-    
+
             47 65 6f 61 64 6d 69 6e 5f 75 73 65 72 6e 61 6d 65 3a 20 0d 0a
             Geoadmin-Username:...
-    
+
         This doesn't seem possible to reproduce with curl. It is possible to
         reproduce with wget. It is unclear whether that technically counts as an
         empty value or a whitespace. It is also possible that AWS change their
@@ -29,4 +29,6 @@ class ApiGatewayMiddleware(PersistentRemoteUserMiddleware):
         apigw_auth = request.META.get("HTTP_GEOADMIN_AUTHENTICATED", "false").lower() == "true"
         if not apigw_auth and self.header in request.META:
             del request.META[self.header]
+        else:
+            setattr(request, '_dont_enforce_csrf_checks', True)
         return super().process_request(request)


schtibe:
So from our discussions, IIRC, I thought the crucial part would be to differentiate between API requests and Django-Admin requests. Would that still be a possibility? Or won't the Django-Admin request work because the Auth Header is removed :thinking:  

(BTW. Is now all traffic going through API Gateway regardless of whether it's API requests or not?)

adk-swisstopo:
> Is now all traffic going through API Gateway regardless of whether it's API requests or not?

Yes, in Dev, all API and admin traffic goes through API Gateway. The plan is to do the same for Int and Prod.
If it makes it easier to handle admin and API traffic differently, we can consider changing that.

adk-swisstopo:
As per discussion:
* API Gateway does not remove the Authorization header so I think this is the wrong fix;
* medium/long term we probably want to split between human-oriented endpoints (admin UI) and machine-oriented endpoints (the rest of /api/stac/) at the Cloudfront level or earlier
* we can verify whether a request is authenticated by API Gateway with the Geoadmin-Authenticated header (but if we do that, we probably want the ApiGatewayMiddleware to expose that header in a nicer Django-style way)

schtibe:
> As per discussion:
>     * medium/long term we probably want to split between human-oriented endpoints (admin UI) and machine-oriented endpoints (the rest of /api/stac/) at the Cloudfront level or earlier

In all honesty, the fact that the admin is reachable under `/api/stac` feels sort of wrong anyways..?

adk-swisstopo:
> > As per discussion:
> > * medium/long term we probably want to split between human-oriented endpoints (admin UI) and machine-oriented endpoints (the rest of /api/stac/) at the Cloudfront level or earlier
> 
> In all honesty, the fact that the admin is reachable under `/api/stac` feels sort of wrong anyways..?

Changing the namespace is sort of orthogonal. It would also impact users in a way that decoupling the the Admin UI from the API endpoints would not (or at least not necessarily). That said, I agree the Admin UI and the "get-token" endpoints should probably be separate from the rest of the API endpoints.

adk-swisstopo:
I think I have a better fix in https://github.com/geoadmin/service-stac/commit/73e846c38e1b13a88c350edcd7062473f4303ddb
Waiting for https://github.com/geoadmin/service-stac/pull/501 to be merged before opening the PR.

PR #493, opened 2025-01-15
PB-1321 adapted fieldnames and spec to comply to forecast extension

boecklic:
Adapted the fieldnames and the spec to comply to the final version of the forecast extension.

The way we add extensions to the `stac_extensions` list is a bit hacky, but probably good enough for the moment. Once we'll have more collection-dependent extensions, we'll have to come up with a better, more generic solution.

Also fixed a bug in tests.


app/stac_api/serializers/item.py
```diff
@@ -380,7 +378,15 @@ def get_type(self, obj):
         return 'Feature'
 
     def get_stac_extensions(self, obj):
-        return []
+        extensions = [
+            # Extension provides schema for the 'expires' timestamp
+            "https://stac-extensions.github.io/timestamps/v1.1.0/schema.json"
+        ]
+        # IMPROVEMENT: This could be improved if there are other extensions coming by
+        # keeping the information on collection object itself
```
asteiner-swisstopo, line 386: I'm not sure I understand what you want to say here. If I get the intention right, maybe this would be clearer?

> We may want to improve this when other extensions are added that keep information about the collection object itself.

app/tests/tests_10/test_serializer.py
```diff
@@ -587,21 +588,23 @@ def test_deserialization_works_as_expected_for_valid_forecast_data(self):
...
         }
 
         serializer = ItemsPropertiesSerializer(data=data)
         self.assertTrue(serializer.is_valid())
 
-        self.assertTrue(
+        self.assertEqual(
             serializer.validated_data["forecast_reference_datetime"],
-            datetime(year=2024, month=11, day=19, hour=16, minute=15)
+            datetime(year=2024, month=11, day=19, hour=16, minute=15, tzinfo=timezone.utc)
```
asteiner-swisstopo, line 600: I assume this is the thing that bug you mentioned? If the default timezone is not UTC, I don't understand why this did not break before. Can you explain please? :upside_down_face:

spec/components/schemas.yaml
```diff
@@ -1093,20 +1095,22 @@ components:
...
         forecast:duration:
           $ref: "#/components/schemas/duration"
-        forecast:param:
+        forecast:variable:
           description: >-
-            Name of the model parameter that corresponds to the data
-          example: T
+            Name of the model variable that corresponds to the data. The variables
+            should correspond to the
+            [CF Standard Names](https://cfconventions.org/Data/cf-standard-names/current/build/cf-standard-name-table.html),
```
asteiner-swisstopo, line 1102: Great to have the link :+1:

spec/components/schemas.yaml
```diff
@@ -1591,6 +1595,13 @@ components:
       title: STAC version
       type: string
       readOnly: true
+    stac_extensions:
+      type: array
+      items:
+        type: string
+      description: List of relevant extensions for an object
+      example:
+        - https://stac-extensions.github.io/forecast/v0.2.0/schema.json
```
asteiner-swisstopo, line 1604: Is it intentional that this is field is now there in the specs but there is no place where it is already used?

asteiner-swisstopo (approved):
Looks good to me, thanks. Just a few questions for my understanding.

app/stac_api/serializers/item.py
boecklic: In principle we only want to have the extensions in this list that are relevant for that specific entity (be it collection, item, asset). So to resolve this in a general way we would have to keep track of which extensions are actually used by which collection/item/asset, which is overkill. The other extreme is to have everywhere the same extensions, whether they are used or not, which isn't ideal as well. A 'middle' way could be to keep track of this on collection level, so we (or the provider) knows which collection actually uses which extension and we could derive the list here based on that information instead of hardcoding based on collection name.

app/tests/tests_10/test_serializer.py
boecklic: The bug is actually that everywhere `assertTrue()` was used, so the two values were not compared but just had to evaluate to true (which is the case for almost everything as long as it's not `None` or `False`).

spec/components/schemas.yaml
boecklic: It is used, see https://github.com/geoadmin/service-stac/pull/493/files#diff-d6b0cd96c896d9730759fdbfdf6faa523fc9e3e479de2eef7aad584f15098cabR388, but doens't appear in the spec otherwise

PR #496, opened 2025-01-20
Re-enable spec consistency check.

adk-swisstopo:
This was disabled in https://github.com/geoadmin/service-stac/pull/389/commits/32439dd1c342b97029c0b6b42516562c58829f38

In https://github.com/geoadmin/service-stac/commit/611d339a2d7d3993152cdaf039cef4d670539c43 we have updated yq to 4.44.1 while our laptops are running 4.44.5. This should be close enough to resolve the problem.

- [ ] [wait until the spec is consistent again](https://github.com/geoadmin/service-stac/pull/495)


ltshb (approved):
Good but if this don't help we should consider stop using yq to merge the spec and use Redocly, @boecklic I thought we once decided to use redocly for the spec when we would go to STAC v1.0 ?

adk-swisstopo:
> Good but if this don't help we should consider stop using yq to merge the spec and use Redocly, @boecklic I thought we once decided to use redocly for the spec when we would go to STAC v1.0 ?

I haven't heard of that plan. I am not familiar with redocly/redoc. I feel like it should be OK to decouple v1 launch from the tools we use to generate the documentation although doing both at the same time would be a little nicer for users. Regardless, I think the real fix here is for the output to be generated and stored as an immutable build artefact and not be checked into the same git repo as the source.

adk-swisstopo:
There is still a diff so the problem is not with yq.

adk-swisstopo:
The diff suggests some fields are being reordered. Looks like yq output is non-deterministic, maybe due to eval-all or ireduce.

adk-swisstopo:
Actually, it's our Makefile that is non-deterministic. In CodeBuild:
```
docker run --rm -v /codebuild/output/src3763/src/github.com/geoadmin/service-stac/spec:/workdir 974517877189.dkr.ecr.eu-central-1.amazonaws.com/external/openapi/yq:4.44.1 eval-all '. as $item ireduce ({}; . *nd $item )' openapi.yaml components/headers.yaml components/parameters.yaml components/responses.yaml components/schemas.yaml | \
sed -E '/\ef:/s/"\..*?(#.*?)"/"\1"/' > static/spec/v1/openapi.yaml
```
On my laptop:
```
docker run --rm -v /home/ltkya/src/service-stac/spec:/workdir mikefarah/yq:4.45.1 eval-all '. as $item ireduce ({}; . *nd $item )' openapi.yaml components/responses.yaml components/schemas.yaml components/parameters.yaml components/headers.yaml | \
sed -E '/\ef:/s/"\..*?(#.*?)"/"\1"/' > static/spec/v1/openapi.yaml
```

Notice how the order of the input files changes. Sorting $^ resolves the problem.

PR #499, opened 2025-01-23
using GET range request instead of HEAD to check external asset reach…

boecklic:
…ability


benschs:
@boecklic Added check for content length.
Added unit tests using the [responses](https://pypi.org/project/responses/) lib to mock the external asset url response.

benschs (approved):

PR #502, opened 2025-01-23
PB-1366: make new auth work for write requests.

adk-swisstopo:
Django is configured to use CSRF mitigation tokens. This only works for clients that are web browsers. Typical API clients are not web browsers, do not keep track of cookies and fail CSRF mitigation checks. The old token authentication method works around that by disabling CSRF mitigation when that authentication method is used. With this change, we do the same for the new token authentication method.

In PB-1009 we implemented the new authentication method for the Admin UI. This change does the same for the API endpoints and add tests exercising it.

The API endpoints are rest_framework views that rely on authentication methods defined by the REST framework which is different from the Admin UI authentication plumbing. To support both, we need to have modules for both even though they basically do the same thing.

Specifically:
* create `api_gateway_authentication` as `rest_framework` authentication module
* create `api_gateway_middleware` as generic Django authentication module (for Admin UI)
* abstract the common parts into a new `api_gateway` module
* configure both authentication modules in the debug build
* add tests exercising a PUT API endpoint for both v0.9 and v1, including CSRF checks

This is all flag-guarded behind FEATURE_AUTH_ENABLE_APIGW.


boecklic (approved):
Very nice!

benschs (approved):

msom (approved):

PR #503, opened 2025-01-27
New Release v1.31.0 - #minor

adk-swisstopo:
The main purpose of this new release it to allow the new authentication method to be used in INT and PROD for PB-1284. This also includes changes required for PB-1372 "allow GET range request to check external asset href".


ltshb (commented):
Can existing user still access the admin page and API the way they use to do it ? Or do they need new credentials ? If they can't access anymore admin page and/or API with their current credentials I would create a major release (add `#major` in the PR title) and we need to do some coordination for the deployment.

adk-swisstopo:
> Can existing user still access the admin page and API the way they use to do it ? Or do they need new credentials ? If they can't access anymore admin page and/or API with their current credentials I would create a major release (add `#major` in the PR title) and we need to do some coordination for the deployment.

The changes in this PR do not actually change the way authentication works. That will be done when the FEATURE_AUTH_ENABLE_APIGW flag is enabled (which happens at Kubernetes level). Even then, the old authentication mechanism keep working in the same way. Hence I believe this qualifies as a minor release.

ltshb (approved):

boecklic (approved):

PR #506, opened 2025-01-28
PB-1281: Disable old authentication for STAC v1

msom:
Disables Basic / Token / Session Auth for STAC v1 if `FEATURE_AUTH_RESTRICT_V1` is set.

Together with `FEATURE_AUTH_ENABLE_APIGW` authentication is as followed:

| Authentication | STAC v0.9                       | STAC v1.0                       |
| ---------- | ------------------------- | ------------------------- |
| Basic      | yes                       | FEATURE_AUTH_RESTRICT_V1  |
| Token      | yes                       | FEATURE_AUTH_RESTRICT_V1  |
| Session    | yes                       | FEATURE_AUTH_RESTRICT_V1  |
| ApiGateway | FEATURE_AUTH_ENABLE_APIGW | FEATURE_AUTH_ENABLE_APIGW |

### DIFF
--- a/app/tests/tests_10/test_asset_upload_endpoint.py
+++ b/app/tests/tests_10/test_asset_upload_endpoint.py
@@ -9,6 +9,9 @@ from urllib import parse
 from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.test import Client
+from django.test import override_settings
+
+from rest_framework.authtoken.models import Token
 
 from stac_api.models.item import Asset
 from stac_api.models.item import AssetUpload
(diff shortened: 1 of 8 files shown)


msom:
On a sidenote: When testing the authentication endpoints I noticed that the unit tests do a login for the search endpoint. I removed this login since the search endpoint is public.

benschs (approved):

app/tests/tests_10/test_asset_upload_endpoint.py
```diff
@@ -1310,3 +1313,71 @@ def test_create_multipart_upload_on_external_asset(self):
         )
 
         self.assertStatusCode(400, response)
+
+
+@override_settings(FEATURE_AUTH_RESTRICT_V1=True)
+class AssetUploadDisabledAuthenticationEndpointTestCase(AssetUploadBaseTest):
```
adk-swisstopo, line 1319: This seems brittle as a small mistake in the test code means we might not test what we think we are testing (e.g. if authentication always fails for reason unrelated to the flag). Can you please add a test verifying session, old token and basic authentication do work when using the same code with FEATURE_AUTH_RESTRICT_V1=False? If those tests already exist, they should use the same code (not copy/paste).

app/tests/tests_10/test_search_endpoint.py
adk-swisstopo, line 1: That's unrelated. I suggest moving this change into a separate PR.

adk-swisstopo (changes requested):

adk-swisstopo (approved):

PR #507, opened 2025-01-28
Do not login when testing search endpoint

msom:
Unit tests should not do a login for the search endpoint since it's public.


adk-swisstopo (changes requested):
Looks good. Can you do the same for v0.9 please? :)

msom:
> Looks good. Can you do the same for v0.9 please? :)

Sure, totally missed that.

adk-swisstopo:
> > Looks good. Can you do the same for v0.9 please? :)
> 
> Sure, totally missed that.

The vagaries of copy/paste :)

PR #508, opened 2025-01-29
PB-1406: Make cognito authenticated users superuser

msom:
Currently `RemoteUserBackend` creates an authenticated user if it not already exists (`create_unknown_user` is set by default). I suggest we generally set the superuser flag for the cognito authenticated users until we implement proper authorization. The other possibility would be to set `create_unknown_user` and create the users by hand (auth-attempts for non-existing users will then return 401 instead of 403).

I also updated the tests for v1 to use the new authentication method, as this is what we want to use there so it doesn't make much sense to use another authentication method in the test.

What do you think?


msom:
Might be a bit early, let's discuss this next week...

PR #509, opened 2025-01-31
PB-1091 Fix file size updater

schtibe:
Change the implementation of the file size update management command so that it uses the s3 backend to determine the file size of an object.

This is needed because the file path can actually differ from the asset name, as is the case with for instance
https://data.geo.admin.ch/api/stac/v0.9/collections/ch.swisstopo.bahnen-winter/items/bahnen-winter/assets/additional-files_bahnen-winter_2056.zip

### DIFF
--- a/app/stac_api/management/commands/update_asset_file_size.py
+++ b/app/stac_api/management/commands/update_asset_file_size.py
@@ -1,17 +1,11 @@
 import logging
 
-from botocore.exceptions import ClientError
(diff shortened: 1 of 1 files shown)


benschs (approved):

boecklic:
@schtibe how can this happen? Imho this should never ever be the case. Is this due to manual interactions on the bucket?

PR #510, opened 2025-02-04
WIP Feat presign url creator

schtibe:
Just adding this here as a reference on how to generate presigned URLs


boecklic (approved):
so this is intended to be mainly used locally during development I guess?

schtibe:
> so this is intended to be mainly used locally during development I guess?

I was using this when I started exploring the possibilities of a jump-off upload solution for the Django admin. 

If you think this could be useful to keep around, it would need a little fix: the asset name isn't unique, and this script would fail if there were two assets with the same name. It should be that collection + item + assets needs to be specified in the command call.

PR #511, opened 2025-02-04
use new auth in v1.0 tests

msom:


app/tests/tests_10/test_asset_upload_endpoint.py
```diff
@@ -52,7 +52,7 @@ class AssetUploadBaseTest(StacBaseTestCase, S3TestMixin):
     @mock_s3_asset_file
     def setUp(self):  # pylint: disable=invalid-name
         self.client = Client()
```
adk-swisstopo, line 54: Instead of adding the headers=self.auth_headers line to every test, would it make sense to just do Client(headers=get_auth_headers()) in setUp()? Or do we have tests where the headers should not be set or set differently?

msom: There are a few instances where there is no authorization in a test of a class but generally this might work, I'll try.

app/tests/tests_10/test_generic_api.py
```diff
@@ -252,7 +253,7 @@ def test_get_precondition(self):
                 self.assertStatusCode(412, response4)
 
     def test_put_precondition(self):
-        client_login(self.client)
+        auth_headers = get_auth_headers()
```
adk-swisstopo, line 256: I think you can use the setUp() trick for this class too? OTOH I can't immediately find where self.client is defined for this class so maybe it's more complicated?

msom: you can't overwrite self.client since the get test requires no auth, but I added a local client

adk-swisstopo (approved):
Sorry for the delay, I ended up writing [a tool](https://github.com/adk-swisstopo/python-unidiff/blob/shunk/bin/shunk) to make this kind of reviews easier.

PR #512, opened 2025-02-04
PB-1279 Batch create endpoint for items/assets

asteiner-swisstopo:
Add a new POST endpoint to `collections/{collectionId}/items` that lets you create multiple Items together with their assets at in one go.

Explicitly handled HTTP return status:

- 201 Created: Successful creation
  (!) Even for empty list of items
- 400 Bad Request:
   - Payload malformed
   - Item exists already
   - Over 100 items
   - No or empty HTTP header parameter "Idempotency-Key"
- 404 Not Found: Collection not found

Example payload:

```json
{
  "features": [
    {
      "id": "feature-1",
      "links": [...],
      "assets": {
        "asset-1-1.txt": {...},
        "asset-1-2.txt": {...},
      },
      "geometry": {...},
      "properties": {...},
      "stac_extensions": [...]
    },
    {
      "id": "feature-2",
      "links": [...],
      "assets": {
        "asset-2-1.txt": {...},
        "asset-2-2.txt": {...},
      },
      "geometry": {...},
      "properties": {...},
      "stac_extensions": [...]
    }
  ]
}
```

The API specs:

1. Build with
   ```sh
   make build-specs
   ```

2. Serve with
   
   ```sh
   make serve-specs
   ```

3. In the API specs, navigate to Data Management > "Bulk create features":

    http://0.0.0.0:8090/apitransactional.html#tag/Data-Management/operation/bulkCreateItems

pb-1279-api-specs-bulkCreateItems

Design decisions:

1. **Only allow bulk creation, not updating.** This is because creation is probably the main use case. Including updates would be possible but it would make the code even more complicated for no real performance gain over individual requests. For the creation of several objects at once we can make use of Django's [`bulk_create()`](https://docs.djangoproject.com/en/5.1/ref/models/querysets/#bulk-create). For the update, however, we cannot just use [`bulk_update()`](https://docs.djangoproject.com/en/5.1/ref/models/querysets/#bulk-update) as we don't necessarily update the same field for all items. So we would have to resort to an iteration, finding out what actually changed. We agreed to leave out the update functionality and check with MeteoSchweiz how this fits their workflow in practice.
2. **Assets as part of an item are set as writeable**, before they were read-only. So before you could only create them individually, now you can create them as part of an item. This affects other endpoints but we agreed that this would not be a problem.
3. **Header parameter `"Idempotency-Key"`**: We require it even though it's not used currently in order to have a more stable interface.
4. **Response format**: `{"code": 1234, "description": "..."}` as it's given in the other API spec examples.

Implementation details:

1. Serializer `ItemListSerializer`
   1. `to_representation()`: Not sure if I'm doing something wrong but for a nested serializer you apparently need to explicitly tell it to wrap the items into a dictionary.
   2. `update()`: The linter forced me to implement this. As we currently don't support updating, I mark this with a `NotImplementedError`.
   3. `validate()`: I wanted to do this a "[field-level validation](https://www.django-rest-framework.org/api-guide/serializers/#field-level-validation)". So only validate the "features" field with something like `validate_features()` instead of a whole `validate()`. But as far as I understand, this does not allow me to call `super().validate_features()`. So I still use `validate()`.
2. Smaller changes:
   1. The update/creation of links is copied from somewhere similar but I must admit that I don't fully see through this. Please review carefully.
   2. `validate_json_payload()`: For reasons I don't understand the `initial_data` (holding the raw input payload) is not properly set for a nested serializer, like `ItemListSerializer` is. Resorting to the dedicated `Serializer.get_initial()` helped for the new serializer but it broke others. In [the DRF docs](https://www.django-rest-framework.org/api-guide/serializers/#accessing-the-initial-data-and-instance) it says "If the data keyword argument is not passed then the .initial_data attribute will not exist.". So we check for that.
   3. `ItemsUnImplementedEndpointTestCase`: This was introduced explictly in [BGDIINF_SB-1725](https://jira.swisstopo.ch/browse/BGDIINF_SB-1725) in 2021. Felt safe to remove now that this endpoint exists.
3. API Specs: Had to be adapted since [PB-1279](https://jira.swisstopo.ch/browse/PB-1279) because we changed some details.
   1. Move the endpoint from `collections/{collectionId}` to `collections/{collectionId}/items`. So it mirrors the "[Fetch features](https://sys-data.dev.bgdi.ch/api/stac/static/spec/v1/apitransactional.html#tag/Data/operation/getFeatures)" GET endpoint with a payload like `{"features": [item1, item2, ...]}`.
   2. Include "assets": Change `components.schemas.itemAssets.readOnly` to `false`. This affects many other endpoints. Curiously, the change is not visible in any of their payload examples because the examples are hard-coded at endpoint level and don't rely on their components.
   3. Adapt the ID of the endpoint from `collectionUpsertItems` to `collectionCreateItems` as we no longer update.

CC @schtibe

### DIFF
--- a/app/stac_api/serializers/item.py
+++ b/app/stac_api/serializers/item.py
@@ -515,3 +515,11 @@ class ItemListSerializer(serializers.Serializer):
         '''Convert to a dict like `{"features": [item1, item2, ...]}`.'''
         items_serialized = [ItemSerializer(item, context=self.context).data for item in instance]
         return {"features": items_serialized}
+
+    @override
+    def validate(self, attrs):
+        max_n_items = 100
+        if len(attrs["features"]) > max_n_items:
+            raise serializers.ValidationError({"features": f"More than {max_n_items} features"})
+
+        return super().validate(attrs)
--- a/app/stac_api/serializers/item.py
+++ b/app/stac_api/serializers/item.py
@@ -475,13 +475,6 @@ class ItemListSerializer(serializers.Serializer):
         '''Create items in bulk from the given list of items.'''
         collection = validated_data["collection"]
 
-        update_fields = [
-            field.name
-            for field in Item._meta.get_fields()
-            if not field.is_relation and not field.primary_key
-        ]
-
-        existing_items = []
         new_items = []
         links_data_list = []
         for item_in in validated_data["features"]:
@@ -489,21 +482,7 @@ class ItemListSerializer(serializers.Serializer):
             links_data_list.append(links_data)
 
             item = Item(**item_in, collection=collection)
-
-            existing_item_query = Item.objects.filter(collection=collection, name=item_in["name"])
-            if existing_item_query.exists():
-                # Convert QuerySet to Item for bulk_update()
-                existing_item = list(existing_item_query.in_bulk().values())[0]
-
-                for field in update_fields:
-                    if field in item_in:
-                        setattr(existing_item, field, item_in[field])
-                existing_items.append(existing_item)
-            else:
-                new_items.append(item)
-
-        if existing_items:
-            Item.objects.bulk_update(existing_items, update_fields)
+            new_items.append(item)
 
         items_created = Item.objects.bulk_create(new_items)
 
--- a/app/stac_api/serializers/item.py
+++ b/app/stac_api/serializers/item.py
@@ -1,6 +1,7 @@
 import copy
 import logging
 from datetime import timedelta
+from typing import override
 
 from django.core.exceptions import ValidationError as CoreValidationError
 from django.utils.translation import gettext_lazy as _
@@ -471,6 +472,7 @@ class ItemListSerializer(serializers.Serializer):
     # with the payload of the getFeatures endpoint.
     features = ItemSerializer(many=True)
 
+    @override
     def create(self, validated_data):
         '''Create items in bulk from the given list of items.'''
         collection = validated_data["collection"]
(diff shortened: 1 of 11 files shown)


benschs (approved):
Cool, nice work.

app/stac_api/views/item.py
```diff
@@ -144,6 +147,35 @@ def list(self, request, *args, **kwargs):
     def get(self, request, *args, **kwargs):
         return self.list(request, *args, **kwargs)
 
+    def post(self, request, *args, **kwargs):
+        try:
+            idempotency_key_param = "Idempotency-Key"
```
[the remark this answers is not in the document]
asteiner-swisstopo: Good idea, will do. Thanks!

app/stac_api/serializers/item.py
```diff
@@ -462,3 +463,63 @@ def validate(self, attrs):
...
+            assets_per_item[item_name] = item_in.pop('assets', [])
+
+            item = Item(**item_in, collection=collection)
+            items.append(item)
+
+        items_created = Item.objects.bulk_create(items)
+
+        assets = []
+        for item in items_created:
+            update_or_create_links(
```
boecklic, line 495: did you consider converting this to a 'bulk' version as well?

boecklic (approved):
Very nice!

app/stac_api/serializers/item.py
asteiner-swisstopo: Good point, I forgot!

As we don't update anymore, we can do it the same way as for the assets and create them in bulk.

PR #514, opened 2025-02-11
PB-1440: allow setting SESSION_* settings via environment variables.

adk-swisstopo:
We want to tweak those but we don't need to hardcode the exact values for all environments immediately. Another change will update the values in DEV, INT then PROD. Afterwards we can consider setting these values by default in STAC itself.

This also removes the specific values we set for the dev build.


app/config/settings_prod.py
```diff
@@ -351,3 +351,10 @@ def get_logging_config():
 DISALLOWED_EXTERNAL_ASSET_URL_SCHEMES = env.list(
     'DISALLOWED_EXTERNAL_ASSET_URL_SCHEMES', default=['http']
 )
+
+# These are the default values from Django as per https://docs.djangoproject.com/en/5.1/ref/settings/
+# We add them here so they can be changed through environment variables.
+SESSION_EXPIRE_AT_BROWSER_CLOSE = env('SESSION_EXPIRE_AT_BROWSER_CLOSE', default=False)
+SESSION_COOKIE_AGE = env('SESSION_COOKIE_AGE', default=60 * 60 * 24 * 7 * 2)
+SESSION_COOKIE_SAMESITE = env('SESSION_COOKIE_SAMESITE', default='Lax')
+SESSION_COOKIE_SECURE = env('SESSION_COOKIE_SECURE', default=False)
```
boecklic, line 360: don't you want to set the defaults to what we want to have in prod?

boecklic (approved):

app/config/settings_prod.py
msom: I would also leave the defaults like they were, the django defaults don't neccessarily make sense...

Defaulting `SESSION_COOKIE_SECURE` to `False` doesn't sound like a good idea.

Defaulting `SESSION_COOKIE_SECURE` to `Lax` when it already works with `strict` also doesn't sound like a good idea.

msom (changes requested):
Not sure about these defaults.

app/config/settings_prod.py
adk-swisstopo: The current state is already 2 weeks, Lax and "insecure". This just makes the default explicit. We will change them via flags at Kubernetes level which is easier to rollback or tweak further than if we need to make a new release every time.

We can set the default to what we want in prod eventually but then it means setting the flags in dev, int, prod upfront while they have no actual effect, which I think would be more confusing. E.g. we could set SESSION_COOKIE_SECURE=True in settings_prod.py and False in {dev,int,prod}/kustomization.yaml. But then updating kustomization.yaml to set it to True won't actually do anything until the new STAC release is rolled out.

msom (approved):

PR #515, opened 2025-02-11
PB-1406: Automatically promote remote users to superusers

msom:
Automatically promote users authenticated via the API gateway middleware to superusers.

Both GeoadminHeadersAuthForPutEndpointTestCase classes have been updated to check that the authentication works for:
- existing superusers
- existing users
  - if validation succeeds, they are promoted to superusers
  - if validation fails, they are not promoted to superusers
- non-existing but authenticated users
  - if validation succeeds, the user is created as superuser
  - if validation fails, no user is created

Normal tests using the new authentication have been adapted to not previously create the API user.


app/middleware/api_gateway_middleware.py
```diff
@@ -29,4 +29,8 @@ def authenticate(self, request, remote_user):
         if not settings.FEATURE_AUTH_ENABLE_APIGW:
             return None
 
-        return super().authenticate(request, remote_user)
+        user = super().authenticate(request, remote_user)
```
adk-swisstopo, line 32: I feel like a comment explaining why it's OK for now would be nice.

adk-swisstopo (approved):

app/middleware/api_gateway_middleware.py
msom: I've added a comment (and also fixed the docstring)

PR #516, opened 2025-02-12
PB-1279: Fix test

msom:
CC @boecklic 

Turns out this only happens during tests because of django wrapping the whole test in a transaction:

```
START TRANSACTION
    SAVE ITEM --> IntegrityError because of existing unique constraint
    SAVE SESSION --> DatabaseError because of an error occured in the current transaction
END TRANSACTION
```

Using a `TransactionalTestCase` instead of a `TestCase` solves the issue.

Posting outside works as expected:
```bash
curl \
  --include \
  --request POST \
  --header "Geoadmin-Username: apiuser" \
  --header "Geoadmin-Authenticated: true" \
  --header "Idempotency-Key: abc-123" \
  --url http://localhost:8000/api/stac/v1/collections/test/items \
  --header 'Content-Type: application/json' \
  --data '{
    "features": [{
        "id": "test.geojson",
        "assets": [{
            "id": "asset-1.txt",
            "title": "My title 1",
            "description": "My description 1",
            "type": "text/plain",
            "href": "asset-1",
            "roles": ["myrole"],
            "geoadmin:variant": "komb",
            "geoadmin:lang": "de",
            "proj:epsg": 2056,
            "gsd": 2.5
        }],
        "links": [{
            "href": "https://www.example.com/described-by-1",
            "rel": "describedBy",
            "title": "My title 1",
            "hreflang": "en"
        }],
        "geometry": {
            "type": "Point",
            "coordinates": [1.1, 1.2]
        },
        "properties": {"datetime": "2018-02-12T23:20:50Z"}
    }]
}'
```

Results in:
```bash
{"code":400,"description":"duplicate key value violates unique constraint \"stac_api_item_collection_id_name_78fbc154_uniq\"\nDETAIL:  Key (collection_id, name)=(267, test.geojson) already exists.\n"}
```

### DIFF
--- a/app/tests/tests_10/test_items_endpoint.py
+++ b/app/tests/tests_10/test_items_endpoint.py
@@ -809,13 +809,12 @@ class ItemsUpdateEndpointTestCase(StacBaseTestCase):
 
 
 @override_settings(FEATURE_AUTH_ENABLE_APIGW=True)
-class ItemsBulkCreateEndpointTestCase(StacBaseTestCase):
+class ItemsBulkCreateEndpointTestCase(StacBaseTransactionTestCase):
 
-    @classmethod
-    def setUpTestData(cls):
-        cls.factory = Factory()
-        cls.collection = cls.factory.create_collection_sample(db_create=True)
-        cls.payload = {
+    def setUp(self):
+        self.factory = Factory()
+        self.collection = self.factory.create_collection_sample(db_create=True)
+        self.payload = {
             "features": [
                 {
                     "id": "item-1",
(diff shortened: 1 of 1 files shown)


app/tests/tests_10/test_items_endpoint.py
```diff
@@ -809,13 +809,12 @@ def test_item_atomic_upsert_update_500(self):
 
 
 @override_settings(FEATURE_AUTH_ENABLE_APIGW=True)
-class ItemsBulkCreateEndpointTestCase(StacBaseTestCase):
+class ItemsBulkCreateEndpointTestCase(StacBaseTransactionTestCase):
```
asteiner-swisstopo, line 812: Might he helpful to add a comment explaining why we use `StacBaseTransactionTestCase` instead of `StacBaseTestCase` like everywhere else.

asteiner-swisstopo (approved):
I see, nice find - thanks!

The only thing I would suggest to add is a short explanation in the code on why this need to be a `StacBaseTransactionTestCase`.

PR #518, opened 2025-02-12
PB-1440: add typing to SESSION_* environment variables.

adk-swisstopo:
This is not strictly required but it is a little nicer and that helpfully triggers a new build to test the fix from https://github.com/geoadmin/infra-terraform-bgdi/pull/762


msom (commented):
:+1: 

BTW you can also use `env.int` etc.

msom (approved):

adk-swisstopo:
> 👍
> 
> BTW you can also use `env.int` etc.

I did use .env.local to test locally. For dev/int/prod this seems less relevant as we want to be able to change these flags without rebuilding a Docker image.

PR #519, opened 2025-02-20
PB-1454: Disable ApiGatewayMiddleware for API calls

msom:
`ApiGatewayMiddleware` creates a new session for every valid header it finds, which allows admin UI users to login by providing once a valid header and then provide the session cookie.

For API request, authentication is done by `ApiGatewayAuthentication`, but `ApiGatewayMiddleware` still runs on every request and creates therefore a new session with every new request (with a valid header).

This PR disables `ApiGatewayMiddleware` for all API calls (`/api/stac/v`).


adk-swisstopo (approved):
I wonder whether that should also be guarded by a feature flag.

msom:
> I wonder whether that should also be guarded by a feature flag.

More likely a bug flag...

boecklic (approved):

PR #520, opened 2025-02-25
update all paths from v0.9 to v1

boecklic:
We still have many references (mostly in examples) to `v0.9` in the spec. They are all replaced with v1 with this PR


ltshb:
Is replaced by #531

PR #522, opened 2025-02-27
PB-1469: use v0.9 for large uploads from admin UI.

adk-swisstopo:
In PB-1229 we started routing all STAC traffic through API Gateway which restricts the size of requests. In PB-1331 we updated the admin UI upload page to use S3 URLs for upload such that large requests don't go through API Gateway any more. This uses v1 API endpoints from Javascript. In PB-1281 we disabled session authentication for v1 endpoints, thus breaking admin UI upload page. In PB-1380 we opened v1 in PROD.

We prefer to not re-enable session authentication for v1 because disabling it later will be harder as (more) users would start relying on it. Current workaround involves uploading assets on behalf of users with our own script using the API.

This change updates the Javascript to use v0.9 API endpoints which still accept session-based authentication. This is a temporary workaround until we figure out a better long term fix that does not rely on v0.9.

I verified that, in conjunction with https://github.com/geoadmin/infra-terraform-bgdi/pull/795, I am able to upload files with the admin UI.

### DIFF
--- a/app/stac_api/templates/js/admin/upload.js
+++ b/app/stac_api/templates/js/admin/upload.js
@@ -126,7 +126,8 @@ async function createPresigned(md5, multi, file) {
                         'md5': md5,
                     }
                 ],
-                'file:checksum': multi,
+                'file:checksum': multi,  // v1
+                'checksum:multihash': multi,  // v0.9
             }),
         });
         if (!response.ok) {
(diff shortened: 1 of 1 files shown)


adk-swisstopo:
@msom @asteiner-swisstopo @boecklic I think that's all we need for the "quick fix" unless you know something is missing? I will try to test it later today.

adk-swisstopo:
With this change, the POST fails like this:
```
400
description | Object { "checksum:multihash": […] }
checksum:multihash | [ "This field is required." ]
0 | "This field is required."
```

adk-swisstopo:
Proposed fix for the CORS issue in https://github.com/geoadmin/infra-terraform-bgdi/pull/795

adk-swisstopo:
@sami-nouidri-swisstopo FYI as you are working on something in that area as part of PB-1434.

boecklic (approved):
:+1:

PR #523, opened 2025-02-27
PB-1434 : removal of the 'file' attribute & small file uploads from admin UI

sami-nouidri-swisstopo:
After discussion with @boecklic , we decided to remove the 'file' attribute from the admin UI, so as to have all file uploads go through the large file upload API calls, therefore eliminating the wrong asset filename problem. 

It turns out many tests relied on that 'file' attribute to upload a file and checks its content/checksum. We decided it was best to adjust the tests in order to use the large file upload, instead of modifying the admin UI to pass the tests.

Following a discussion with @schtibe, we think the best approach is to remove the file checks in the test_admin suite. These tests mainly revolve around testing admin page functionality, and since file upload is no longer part of it, it doesn't make sense for us to test for it in this context. In addition, there are API file upload tests already present in other test suites. 

As such, my initial approach of writing a custom class for file uploads in admin has been dropped.


schtibe (commented):
A small issue here: if the asset is flagged to be hosted externally (`is_external` on the Model), then we need the input field for the file to see and set the URL: 
image

Also, you can now rename the Button "Upload Large File" to "Upload File", as it is now not only meant for large files. Maybe, the deluxe solution would be if there's a hint where the file field used to be that tells the user to use this link now? (in case it's not "external").

schtibe:
Also, there used to be a link to the file: 
image

Maybe you can just disable the input and adjust the warning?

boecklic:
As mentioned on slack, I don't think we need this widget or the field at all. We have the `href` attribute which is sufficient.

sami-nouidri-swisstopo:
> As mentioned on slack, I don't think we need this widget or the field at all. We have the `href` attribute which is sufficient.

The widget is there as I haven't found a solution to display the file value without it looking like a text field. It is possible to use a field (forms.CharField) in read-only mode for internal assets, but I feel this looks worst than the custom "label" widget. Do we go for that option to avoid adding more code?

Charfield in read-only mode : 
image

Custom widget : 
image



Of course, it may be that I missed something and that I can also make it a label without a custom widget?

app/stac_api/widgets/label_widget.py
```diff
@@ -0,0 +1,11 @@
+from django import forms
+from django.utils.safestring import mark_safe
+
+
+class LabelWidget(forms.TextInput):
```
schtibe, line 5: I'm a bit confused about why this inherits `TextInput`. This should always display a text, shouldn't it? Or is there a case where `if value` isn't met?

app/stac_api/widgets/label_widget.py
```diff
@@ -0,0 +1,11 @@
...
+    """Custom widget to display a label without input"""
+
+    def render(self, name, value, attrs=None, renderer=None):
+        if value:
+            return mark_safe(f'{value}')
```
schtibe, line 10: If you do this
`return mark_safe(f'<div class="readonly">{value}</div>')`
then the value will be aligned properly :)

schtibe (approved):
Nice. I just have a small comment on the new Widget.

schtibe:
> > As mentioned on slack, I don't think we need this widget or the field at all. We have the `href` attribute which is sufficient.
> 
> The widget is there as I haven't found a solution to display the file value without it looking like a text field. It is possible to use a field (forms.CharField) in read-only mode for internal assets, but I feel this looks worst than the custom "label" widget. Do we go for that option to avoid adding more code?
> 
> Charfield in read-only mode : image
> 
> Custom widget : image
> 
> Of course, it may be that I missed something and that I can also make it a label without a custom widget?

You could potentially do it the other way around:
Drop the `href` field and in the custom widget you do this:
`return build_asset_href(self.request, instance.file.name)`

Although the challenge might be to figure out how to pass `request` and `instance`

schtibe:
I also had a look at the git commit messages. You could greatly improve them, maybe see [this article](https://cbea.ms/git-commit/) as a starting point.
Also, have a look into `rebasing` in git. Most easy way would be `git commit --amend`. I can give you a intro some time if you want, too.

sami-nouidri-swisstopo:
> > > As mentioned on slack, I don't think we need this widget or the field at all. We have the `href` attribute which is sufficient.
> > 
> > 
> > The widget is there as I haven't found a solution to display the file value without it looking like a text field. It is possible to use a field (forms.CharField) in read-only mode for internal assets, but I feel this looks worst than the custom "label" widget. Do we go for that option to avoid adding more code?
> > Charfield in read-only mode : image
> > Custom widget : image
> > Of course, it may be that I missed something and that I can also make it a label without a custom widget?
> 
> You could potentially do it the other way around altogether: Drop the `href` field and in the custom widget you do this: `return build_asset_href(self.request, instance.file.name)`
> 
> Although the challenge might be to figure out how to pass `request` and `instance`

I fear this may make it more complicated and less easily readable but I'm not sure

PR #525, opened 2025-03-03
Added vscode setting to run unittest from vscode

ltshb:
Those workspace settings are required in order to make use of vscode unittest tab. This helps debugging application.


schtibe (approved):
Nice!

PR #526, opened 2025-03-03
PB-929 : added Cognito username logging for requests

sami-nouidri-swisstopo:
The user's username is now logged whenever a request is made. This works for all authenticated requests which use the Geoadmin-Username (Cognito) header.

Preview : 
`'user' : { 'username' : 'apiuser' }`

### DIFF
--- a/app/config/logging-cfg-local.yml
+++ b/app/config/logging-cfg-local.yml
@@ -51,6 +51,7 @@ filters:
       - request.method
       - request.headers
       - request.META
+      - request.user.username
     exclude_keys:
       - request.headers.Authorization
       - request.headers.Proxy-Authorization


boecklic:
You probably have to check with @ltflb-bgdi next week to make sure this field is actually considered in Elasticsearch parsing

ltflb-bgdi:
@sami-nouidri-swisstopo Nice. Could this somehow translated to a [ecs compatible](https://www.elastic.co/guide/en/ecs/current/ecs-user.html#field-user-name) field? E.g. user.name.

sami-nouidri-swisstopo:
> Note that this only enable the log for local deployment to enable it on the DEV, INT and PROD staging you have to change the logging configuration on infra-kubernetes

see https://github.com/geoadmin/infra-kubernetes/pull/740 for that change

sami-nouidri-swisstopo:
> @sami-nouidri-swisstopo Nice. Could this somehow translated to a [ecs compatible](https://www.elastic.co/guide/en/ecs/current/ecs-user.html#field-user-name) field? E.g. user.name.

I think this could be done by modifying the JsonDjangoRequest but I'm not entirely sure. I don't know if it's worth pursuing considering these logs are local only, and won't be displayed in Kibana.

ltshb:
> > @sami-nouidri-swisstopo Nice. Could this somehow translated to a [ecs compatible](https://www.elastic.co/guide/en/ecs/current/ecs-user.html#field-user-name) field? E.g. user.name.
> 
> I think this could be done by modifying the JsonDjangoRequest but I'm not entirely sure. I don't know if it's worth pursuing considering these logs are local only, and won't be displayed in Kibana.

Logging config has two part, the one `JsonDjangoRequest` simply configure django to add a json compatible json information to the log record, what you changed here. The second part is the `JsonFormatter` which format the log record into a json and decide the mapping between the json output keys and the record keys. So there is no code modification required just configuration. But as said this concern only the localhost usage of STAC so it can be merged like this. For ECS field to have them on Kibana this needs to be done via kubernetes

PR #527, opened 2025-03-04
PB-1126: Removed the update_interval field from item and collections model

ltshb:
Due to performance issue we removed the `update_interval` field on the item and collection level and replace it by a single filed `cache_control_header` on the collection level.

Now use the previously new cache_control_header field on the collection model
to overwrite the default cache settings of assets, items and collections
endpoints. Note that endpoints that aggregate several collections like the
search endpoint and the collections list endpoint have now caching disabled
to avoid any caching issues and to keep the code simple.

⚠️  This should merge only when #528 has been released !


boecklic (approved):
Yay, code is getting simpler :+1: nice!

PR #528, opened 2025-03-04
PB-1126: Added cache_control_header field to collection model

ltshb:
This new field will replace later on the update_interval field. But we add
it separately so that we can pre-configured collections that need it up front
before removing the update_interval field that is set via API call.


app/stac_api/models/collection.py
```diff
@@ -105,6 +105,15 @@ class Meta:
                     'protocol://domain values for the external asset url validation')
     )
 
+    cache_control_header = models.CharField(
```
boecklic, line 108: should we add some validation function for what's entered here?

boecklic (approved):

PR #529, opened 2025-03-05
PB-1469: fix charset conversion for large file upload.

adk-swisstopo:
There are multiple problems preventing non-ASCII files from being uploaded correctly. Specifically:
* the content of the file was converted to UTF-8 before upload;
* the content of the file was converted to UTF-8 before MD5 hashing (for S3);
* the content of the file was converted to UTF-8 before SHA-256 hashing (for STAC).

None of these conversion was appropriate and resulted in mangled file and incorrect hashes. This PR removes them.


ltshb (approved):
👍🏼

PR #530, opened 2025-03-05
 PB-1486: Made the Collection Asset(s) endpoint(s) public in openapi spec

ltshb:
The endpoints collections/{collection}/assets and
    collections/{collection}/assets/{asset} where only part of the transactional
    API spec. Now they are also in the public spec.
    
    Also fixed some small issues in the collection asset spec.

### DIFF
--- a/spec/components/parameters.yaml
+++ b/spec/components/parameters.yaml
@@ -1,6 +1,13 @@
 openapi: 3.0.3
 components:
   parameters:
+    assetId:
+      name: assetId
+      in: path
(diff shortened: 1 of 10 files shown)


adk-swisstopo (approved):
This adds many references to v0.9. I assume you will update them as part of https://github.com/geoadmin/service-stac/pull/531 ?

ltshb:
> This adds many references to v0.9. I assume you will update them as part of #531 ?

Yes #531 is based on top of this PR and solve all v0.9

PR #532, opened 2025-03-06
Make large uploads a little more useable.

adk-swisstopo:
Specifically:
* actually attempt to abort abandoned uploads before starting a new one (does not actually work well depending on cache state but better than not working at all);
* document the actual size limit (~512MB).

This is still far from ideal but strictly before than the current state.

### DIFF
--- a/app/templates/uploadtemplate.html
+++ b/app/templates/uploadtemplate.html
@@ -20,7 +20,7 @@
 
 <div class="box info">
     &#9432; This file upload will use the multipart form upload API in the background.
-    The file will be uploaded to s3 using a presigned url. Max. file size is 5GB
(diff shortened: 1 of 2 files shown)


app/templates/uploadtemplate.html
```diff
@@ -20,7 +20,7 @@ <h1>Upload file for {{ collection_name }}/{{ item_name }}/{{ asset_name }}</h1>
 
 <div class="box info">
     &#9432; This file upload will use the multipart form upload API in the background.
-    The file will be uploaded to s3 using a presigned url. Max. file size is 5GB
+    The file will be uploaded to S3 using a presigned url. Maximum file size is 536,870,903 bytes (approximately 512 MB).
```
boecklic, line 23: :smile:

boecklic (approved):

PR #534, opened 2025-03-10
Allow to use docker compose up directly

ltshb:
When entering manually docker compose up it uses random ports and DB name.

So now it uses the same as the .env.default file.


sami-nouidri-swisstopo (approved):
LGTM

PR #536, opened 2025-03-10
Fixed E2E tests for the localhost

ltshb:
Those session settings are required in order to have our E2E tests from
infra-e2e-tests repos to pass when running them using the localhost setup.

See also https://github.com/geoadmin/infra-e2e-tests/pull/344


sami-nouidri-swisstopo (approved):
LGTM

PR #537, opened 2025-03-10
PB-1495 Remove date comparison for property `expires`

asteiner-swisstopo:
In the forecast extension, `datetime` is the forecast date. So `datetime` can be anywhere in the future and `expires` might be before that. So comparing these two dates makes no sense. Same for `end_datetime`.

The `expires` field is part of the Timestamps extension, so it might be used in different contexts where this check would actually make sense. But ultimately, it is the responsibility of the data owner that the data is consistent. So we remove the check to reduce complexity on our side.

:information_source:  The definition of the fields according to the [forecast extension docs](https://github.com/stac-extensions/forecast?tab=readme-ov-file#additional-fields-from-other-extensions):
> - **datetime**: REQUIRED. The forecast datetime. It follows the definition in the [STAC Common Metdata](https://github.com/radiantearth/stac-spec/blob/master/item-spec/common-metadata.md#date-and-time). If the forecast is not only for a specific instance in time but instead is for a certain period, you should use start_datetime and end_datetime and set datetime either to reflect the start_datetime (recommended) or to null.
> - **expires**: The datetime until the forecast is valid or gets superseded by a new forecast. It follows the definition in the [Timestamps Extension](https://github.com/stac-extensions/timestamps#item-properties-fields).

### DIFF
--- a/app/stac_api/validators.py
+++ b/app/stac_api/validators.py
@@ -2,6 +2,7 @@ import logging
 import re
 from collections import namedtuple
 from datetime import datetime
+from datetime import timezone
 from urllib.parse import urlparse
 
 import multihash
@@ -413,21 +414,18 @@ def validate_datetime_format(date_string):
 
 
 def validate_item_properties_datetimes(
-    properties_datetime, properties_start_datetime, properties_end_datetime, properties_expires
+    properties_datetime, properties_start_datetime, properties_end_datetime
 ):
     '''
     Validate the dependencies between the item datetime properties.
     This makes sure that either only the properties.datetime is set or
     both properties.start_datetime and properties.end_datetime are set.
-    If properties.expires is set, the value must be greater than
-    the properties.datetime or the properties.end_datetime.
     Raises:
         django.core.exceptions.ValidationError
     '''
     properties_datetime = validate_datetime_format(properties_datetime)
     properties_start_datetime = validate_datetime_format(properties_start_datetime)
     properties_end_datetime = validate_datetime_format(properties_end_datetime)
-    properties_expires = validate_datetime_format(properties_expires)
 
     if properties_datetime is not None:
         if (properties_start_datetime is not None or properties_end_datetime is not None):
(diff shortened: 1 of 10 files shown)


app/stac_api/validators.py
```diff
@@ -427,6 +427,18 @@ def validate_item_properties_datetimes(
...
+    #
+    #     expires ≥ datetime
+    #
+    # or
+    #
+    #     expires ≥ end_datetime
+    #
+    # because `expires` is only used in the forecast STAC extension.
+    # For forecast data, `datetime` can be anywhere in the future, so
+    # it can be after the expiration date of the forecast data itself.
```
boecklic, line 441: I'd remove that comment, I think it will just confuse in the future, expires is not conceptually related to forecast, MCH is just the first ones making use of it

app/stac_api/validators.py
```diff
@@ -435,10 +447,6 @@ def validate_item_properties_datetimes(
                 '(start_datetime, end_datetime)'
             logger.error(message)
             raise ValidationError(_(message), code='invalid')
-        if properties_expires is not None:
-            if properties_expires < properties_datetime:
-                message = "Property expires can't refer to a date earlier than property datetime"
-                raise ValidationError(_(message), code='invalid')
```
boecklic, line 441: we could replace that with a meaningful check: having an `expires` timestamp in the past doesn't make sense because it would immediately be deleted again, so this is probably a case we don't want to allow.

boecklic (approved):
instead of just removing the check completely, we could think about adapting it to sth that makes more sense than what we have currently

app/stac_api/validators.py
asteiner-swisstopo: Yeah why not.

I tried that and unfortunately it makes some existing tests more complicated: They create objects that are supposed to be expired. This does not work anymore with this validation.

My workaround was to create objects with a timestamp "now() + 10ms", then wait 20ms until they expire themselves. It's quite ugly and affects several tests.

Please let me know if you have a better idea.

boecklic: Hmm, you're right, this smells like resulting in flaky tests... You could maybe try to patch `datetime`? Create the objects in setup with an expire date in the future, then patch the datetime for the function where you test that it's correctly expired.

asteiner-swisstopo: Great idea, wasn't aware of this patching functionality - thanks!

I now create an object at "now() + 1h" and then I advance the clock by 2h by patching `django.utils.timezone.now()`:
```py
with patch.object(timezone, "now", return_value=timezone.now() + timedelta(hours=2)):
```

boecklic: Very nice :ok_hand: !

PR #540, opened 2025-03-12
PB-1434 : Add back 'file' field to Collections and CollectionAssets

sami-nouidri-swisstopo:
The 'file' field has been added back to the Collections and CollectionsAsset form after I removed in the previous PR.

UPDATE : I also found a better implementation for removing the upload button, which involves simply setting the 'file' field as read-only by default, and only making it editable for external assets. This also solves the missing href problem we had with the custom widget. 

LabelWidget has therefore been removed.

### DIFF
--- a/app/stac_api/admin.py
+++ b/app/stac_api/admin.py
@@ -607,7 +607,7 @@ class AssetAdmin(admin.ModelAdmin):
         if obj:
             # Don't allow to modify Asset name and media type, because they are tightly coupled
             # with the asset data file. Changing them require to re-upload the data.
-            # As file upload through admin has been disabled, the field is read-only unless the field is external.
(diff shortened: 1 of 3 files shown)


ltshb:
@sami-nouidri-swisstopo there is a ticket for that see PB-1515 you can assign to yourself

ltshb (approved):
Looks good if it works ? Did you test it ?

sami-nouidri-swisstopo:
> @sami-nouidri-swisstopo there is a ticket for that see [PB-1515](https://swissgeoplatform.atlassian.net/browse/PB-1515) you can assign to yourself

This is slightly different, it's more about keeping the 'file' field consistent between Assets and CollectionAssets like so : 
image

I'm happy to take on that other ticket as well, I'll do it in a different branch/PR

ltshb (approved):

PR #541, opened 2025-03-12
PB-1492: Made all optional item properties nullable in PATCH

ltshb:
This way we can remove them with a PATCH.

### DIFF
--- a/app/tests/tests_09/test_items_endpoint.py
+++ b/app/tests/tests_09/test_items_endpoint.py
@@ -1,3 +1,4 @@
+# pylint: disable=too-many-lines
 import logging
 from datetime import datetime
 from datetime import timedelta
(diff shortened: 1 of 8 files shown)


app/tests/tests_09/test_items_endpoint.py
```diff
@@ -675,6 +676,65 @@ def test_item_endpoint_patch(self):
...
+                "forecast:variable": None,
+                "forecast:perturbed": None
+            },
+        }
+        path = f'/{STAC_BASE_V}/collections/{self.collection["name"]}/items/{self.item["name"]}'
+        response = self.client.patch(path, data=data, content_type="application/json")
+        json_data = response.json()
+        self.assertStatusCode(200, response)
+        self.assertEqual(self.item['name'], json_data['id'])
+        self.assertNotIn("title", json_data['properties'].keys())
```
adk-swisstopo, line 717: Perhaps something like this would be a little nicer?

for field in data['properties']:
  self.assertNotIn(field, json_data['properties'])

adk-swisstopo (approved):

app/tests/tests_09/test_items_endpoint.py
ltshb: @adk-swisstopo Programmatically yes but having the assert listed make the assertions more declarative an easier to read and to debug in case of assertion failure as you can directly jump to the assertion that failed. Most of the time in case of unittest if you try to be clever and encapsulate too much repetitive code in loop or functions, it make the code easy to maintain and change but on the otherhand when an assertion failed it is harder then to debug and know what has exaclty failed.

ltclm:
thank you @ltshb

PR #545, opened 2025-03-17
PB-1482 : Add is_external field to collection assets

sami-nouidri-swisstopo:
The 'is_external' field has been added in order to manage collection assets with external hrefs.


app/stac_api/migrations/0065_collectionasset_is_external.py
```diff
@@ -0,0 +1,21 @@
+# Generated by Django 5.0.11 on 2025-03-17 13:39
+
+from django.db import migrations
+from django.db import models
```
ltshb, line 4: Migrations should be merged together, we do only one migration per PR usually.

app/stac_api/migrations/0066_merge_20250319_0934.py
```diff
@@ -0,0 +1,13 @@
...
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('stac_api', '0065_collectionasset_is_external'),
+        ('stac_api', '0065_remove_asset_add_del_asset_item_update_interval_trigger_and_more'),
+    ]
```
ltshb, line 11: Seems not so correct here, your migration should be the 66 not the 65. So you need to revert the migration, then delete those two files and re-create the migrations so that you have only one file with the correct 66 number, see readme about down migrating.

app/tests/tests_10/test_collection_assets_endpoint.py
```diff
@@ -316,29 +316,15 @@ def test_asset_endpoint_patch_put_href(self):
         asset_sample = self.asset.copy()
 
         put_payload = asset_sample.get_json('put')
-        put_payload['href'] = 'https://testserver/non-existing-asset'
-        patch_payload = {'href': 'https://testserver/non-existing-asset'}
+        put_payload['href'] = 'https://testserver/external-asset'
```
ltshb, line 319: Test doesn't seems to be fully complete in this class we should have

-`test_asset_endpoint_patch_put_href_to_external_href`. Actual test case like you modified it, meaning the current asset is internal and is modified to external. Making also sure that before the update you have an internal href.
- `test_asset_endpoint_patch_put_external_href` similar as above the the asset should already be an external asset where you modify the href, so making sure that before the put/patch you have href A and after you have href B.

ltshb (commented):
You should also add a test case to create a new collection asset with an external asset. Eventually also an admin interface test case for new external collection asset

app/tests/tests_10/test_collection_assets_endpoint.py
```diff
@@ -310,35 +310,46 @@ def setUp(self):  # pylint: disable=invalid-name
...
-            description['href'][0],
-            msg="Unexpected error message"
-        )
+        self.assertStatusCode(200, response)
+
+    def test_asset_endpoint_patch_put_external_href(self):
+        collection_name = self.collection['name']
+        asset_name = 'new-external-asset.tiff'
+        asset_sample = self.asset.copy()
+        asset_sample['id'] = asset_name
```
ltshb, line 339: Id name should not be needed here

app/tests/tests_10/test_collection_assets_endpoint.py
```diff
@@ -310,35 +310,46 @@ def setUp(self):  # pylint: disable=invalid-name
...
-            "Found read-only property in payload",
-            description['href'][0],
-            msg="Unexpected error message"
-        )
+        self.assertStatusCode(200, response)
+
+    def test_asset_endpoint_patch_put_external_href(self):
+        collection_name = self.collection['name']
+        asset_name = 'new-external-asset.tiff'
+        asset_sample = self.asset.copy()
```
ltshb, line 338: The test is not updating an existing external asset but creating a new asset and should not be part of this test class which is about updating. So you need to check in the setUp() method of the class if you already have an external asset and if not add one there, then in this test you take the external asset, first assert that it is external, then update it and assert that it has been updated.

app/tests/tests_10/test_collection_assets_endpoint.py
```diff
@@ -310,35 +310,46 @@ def setUp(self):  # pylint: disable=invalid-name
...
         collection_name = self.collection['name']
         asset_name = self.asset['name']
         asset_sample = self.asset.copy()
 
+        path = f'/{STAC_BASE_V}/collections/{collection_name}/assets/{asset_name}'
+
+        #verify the asset is internal before patching it
+        response = self.client.get(path)
+        self.assertStatusCode(200, response)
+        self.assertFalse(response.json()['is_external'])
```
ltshb, line 323: I'm really concern about this, it means that `is_external` is returned as part of the response but is not specified in the API, @boecklic is it intended ?

sami-nouidri-swisstopo: At first, I thought of asserting whether it was external or not by parsing the href field for an additional url (for example two http strings), but eventually found it cleaner to return is_external as part of the response.

boecklic: @ltshb @sami-nouidri-swisstopo nope, `is_external` should **not** be part of the response. Behaviour should be exactly like for item assets. 
The logic is as follows:
if on collection level the flag is set that allows to have external hrefs, `PUT/PATCH` calls including a `href` field will be accepted, otherwise denied. So we should rather test that this flag works correctly

app/tests/tests_10/test_collection_assets_endpoint.py
```diff
@@ -307,38 +309,71 @@ def setUp(self):  # pylint: disable=invalid-name
...
+        collection_name = self.collection_external.name
+        asset_name = self.internal_asset['name']
+        asset_sample = self.internal_asset.copy()
+
+        path = f'/{STAC_BASE_V}/collections/{collection_name}/assets/{asset_name}'
+
+        #verify the asset is internal before patching it
+        response = self.client.get(path)
+        self.assertStatusCode(200, response)
+        self.assertIn(TEST_SERVER, response.json()['href'])
```
ltshb, line 340: A message in assert would be nice here, `self.asserIn(..., msg="...")`

app/tests/tests_10/test_collection_assets_endpoint.py
```diff
@@ -307,38 +309,71 @@ def setUp(self):  # pylint: disable=invalid-name
...
 
         put_payload = asset_sample.get_json('put')
-        put_payload['href'] = 'https://testserver/non-existing-asset'
-        patch_payload = {'href': 'https://testserver/non-existing-asset'}
+        put_payload['href'] = settings.EXTERNAL_TEST_ASSET_URL
+        patch_payload = {'href': settings.EXTERNAL_TEST_ASSET_URL_2}
+
+        response = self.client.put(path, data=put_payload, content_type="application/json")
+        self.assertStatusCode(200, response)
+        self.assertIn(put_payload['href'], response.json()['href'])
```
ltshb, line 348: Here it should `assertEqual` both href must exactly match !

app/tests/tests_10/test_collection_assets_endpoint.py
```diff
@@ -307,38 +309,71 @@ def setUp(self):  # pylint: disable=invalid-name
...
-        self.assertStatusCode(400, response)
-        description = response.json()['description']
-        self.assertIn('href', description, msg=f'Unexpected field error {description}')
-        self.assertEqual(
-            "Found read-only property in payload",
-            description['href'][0],
-            msg="Unexpected error message"
-        )
+        self.assertStatusCode(200, response)
+        self.assertIn(patch_payload['href'], response.json()['href'])
```
ltshb, line 352: Same as previous comment use exact match

app/tests/tests_10/test_collection_assets_endpoint.py
```diff
@@ -307,38 +309,71 @@ def setUp(self):  # pylint: disable=invalid-name
...
+        )
+
         self.client = Client(headers=get_auth_headers())
         self.maxDiff = None  # pylint: disable=invalid-name
 
-    def test_asset_endpoint_patch_put_href(self):
-        collection_name = self.collection['name']
-        asset_name = self.asset['name']
-        asset_sample = self.asset.copy()
+    def test_asset_endpoint_put_patch_href_to_external_href(self):
```
ltshb, line 330: for clarity `test_asset_endpoint_put_patch_internal_href_to_external_href`

ltshb (changes requested):
The new tests case should be pretty simple and straight forward and implemented here to be sure that this PR is correct.

sami-nouidri-swisstopo:
I've made the changes requested, and added a new `test_external_collection_assets_endpoint` test suite based on `test_external_assets_endpoint`.

app/tests/tests_10/test_external_collection_assets_endpoint.py
```diff
@@ -0,0 +1,373 @@
...
+        # create the asset, which isn't allowed
+        response = self.client.put(
+            reverse_version(
+                'collection-asset-detail', args=[collection.name, collection_asset_data['id']]
+            ),
+            data=collection_asset_data,
+            content_type="application/json"
+        )
+
+        self.assertStatusCode(400, response)
```
ltshb, line 49: Test also that the error message is the one that say that external asset are not allowed

app/tests/tests_10/test_external_collection_assets_endpoint.py
```diff
@@ -0,0 +1,373 @@
...
+            content_type='application/json',
+            adding_headers={'Content-Length': '3'},
+            match=[responses.matchers.header_matcher({"Range": "bytes=0-2"})]
+        )
+
+        collection_asset_data = {
+            'id': 'clouds.jpg',
+            'description': 'High in the sky',
+            'type': 'image/jpeg',  # specify the file explicitly
+            # 'href': settings.EXTERNAL_TEST_ASSET_URL
```
ltshb, line 97: clean up

app/tests/tests_10/test_external_collection_assets_endpoint.py
```diff
@@ -0,0 +1,373 @@
...
+        # Mock response of external asset url
+        responses.add(
+            method=responses.GET,
+            url=external_test_asset_url,
+            body='som',
+            status=200,
+            content_type='application/json',
+            adding_headers={'Content-Length': '3'},
+            match=[responses.matchers.header_matcher({"Range": "bytes=0-2"})]
+        )
```
ltshb, line 91: Is the mocking is really needed for the call bellow to work ? Why ? Maybe a comment would be good here

app/tests/tests_10/test_external_collection_assets_endpoint.py
```diff
@@ -0,0 +1,373 @@
...
+            collection=collection, sample='external-asset', db_create=True
+        )
+
+        response = self.client.delete(
+            reverse_version(
+                'collection-asset-detail', args=[collection.name, collection_asset.attr_name]
+            )
+        )
+
+        self.assertStatusCode(200, response)
```
ltshb, line 373: Test if the object is really deleted from the DB

app/tests/tests_10/test_external_collection_assets_endpoint.py
```diff
@@ -0,0 +1,373 @@
...
+        # create the asset with an existing one
+        response = self.client.put(
+            reverse_version(
+                'collection-asset-detail', args=[collection.name, collection_asset_data['id']]
+            ),
+            data=collection_asset_data,
+            content_type="application/json"
+        )
+
+        description = response.json()['description']
```
ltshb, line 352: Assert the 400 http

ltshb: Also at the end of the test check that no asset has been created on the DB

app/tests/tests_10/test_external_collection_assets_endpoint.py
```diff
@@ -0,0 +1,373 @@
...
```
ltshb, line 320: Assert status code

Also at the end of the test check that no asset has been created on the DB

app/tests/tests_10/test_external_collection_assets_endpoint.py
```diff
@@ -0,0 +1,373 @@
...
+            content_type="application/json"
+        )
+
+        self.assertStatusCode(400, response)
+        description = response.json()['description']
+        self.assertIn('href', description, msg=f'Unexpected field error {description}')
+
+        self.assertEqual(
+            "Invalid URL provided", description['href'][0], msg="Unexpected error message"
+        )
```
ltshb, line 281: Also at the end of the test check that no asset has been created on the DB

app/tests/tests_10/test_external_collection_assets_endpoint.py
```diff
@@ -0,0 +1,373 @@
...
+
+        self.assertStatusCode(400, response)
+        description = response.json()['description']
+        self.assertIn('href', description, msg=f'Unexpected field error {description}')
+        self.assertIn(
+            'Provided URL returns bad content',
+            description['href'],
+            msg=f'Unexpected field error {description}'
+        )
```
ltshb, line 205: Also at the end of the test check that no asset has been created on the DB

app/tests/tests_10/test_external_collection_assets_endpoint.py
```diff
@@ -0,0 +1,373 @@
...
+
+        self.assertStatusCode(400, response)
+        description = response.json()['description']
+        self.assertIn('href', description, msg=f'Unexpected field error {description}')
+        self.assertIn(
+            'Provided URL is unreachable',
+            description['href'],
+            msg=f'Unexpected field error {description}'
+        )
```
ltshb, line 161: Also at the end of the test check that no asset has been created on the DB

ltshb (commented):
Very nice, just some comment see in code.
Also I would put all test case that test failures (invalid input) together in a class to better organize the test code.

ltshb (approved):
💯

PR #546, opened 2025-03-18
PB-1525: Make size of logged request and response content configurable

msom:
This PR make the size of the logged request and response content configurable with environment variables. Also adds tests for the logging middleware.


ltshb (approved):
👍🏼  Very good idea for the unittests 👍🏼

PR #547, opened 2025-03-20
PB-1531: Avoid unnecessary writes to the database

msom:
Avoid unnecessary writes to the database when using the new authentication.

Might not be related to the server errors (which I think are caused by pods waiting for db connections because of some reason), but it's unnecessary the way it's currently anyway.


ltshb (approved):

boecklic (approved):
Nice :+1:

adk-swisstopo (approved):

PR #548, opened 2025-03-20
PB-1519 Issue with asset href extra prefix

asteiner-swisstopo:
This removes the prefix `https://sys-data.int.bgdi.ch` from the `href` field of assets uploaded via [the bulk upload](https://sys-data.int.bgdi.ch/api/stac/static/spec/v1/apitransactional.html#tag/Data-Management/operation/bulkCreateItems).

This happened because there was no `is_external` flag set when validating the `href` field from a payload. If this flag is not set, the validation builds it's own URL. This URL is prefixed with our internal hostname, e.g. `https://sys-data.int.bgdi.ch` for INT or `https://testserver` for local development. So ended up having all `href` prefixed with the `https://sys-data.int.bgdi.ch`.

The special treatment of assets' `href` was introduced in [PB-35 external asset URL](https://swissgeoplatform.atlassian.net/browse/PB-35), that is why I set @schtibe as reviewer.

To see how the validation of `href` works one can follow the example of the [putAsset endpoint](https://sys-data.int.bgdi.ch/api/stac/static/spec/v1/apitransactional.html#tag/Data-Management/operation/putAsset). Here, the asset's `href`` is set to one of our own URLs if it's not defined in the payload.

There are two changes now for the bulk upload:
1. For the bulk upload, assume that every asset points to an external resource. That is just to get early feedback before changing too much. We could also add a conditional branch to construct the URL as for putAsset.
2. Add validation of the `href` fields, just as it's done for the `putAsset` endpoint. See `validators.validate_href_url()`: This checks, among others, if the URL is white listed and reachable. **Pinging the URL probably introduces a notable performance penalty - do we really want that? @boecklic**

As I changed the relatively central `AssetBaseSerializer`, I am not 100% confident that I have not introduced accidental changes elsewhere. The tests are green but I am not sure that the asset's `href` field is checked for appropriately everywhere.

### DIFF
--- a/app/stac_api/serializers/item.py
+++ b/app/stac_api/serializers/item.py
@@ -27,6 +27,7 @@ from stac_api.validators import validate_asset_name_with_media_type
 from stac_api.validators import validate_expires
 from stac_api.validators import validate_geoadmin_variant
 from stac_api.validators import validate_href_url
+from stac_api.validators import validate_href_reachability
 from stac_api.validators import validate_item_properties_datetimes
 from stac_api.validators import validate_name
 from stac_api.validators_serializer import validate_json_payload
@@ -266,6 +267,8 @@ class AssetBaseSerializer(NonNullModelSerializer, UpsertModelSerializerMixin):
 
             try:
                 validate_href_url(attrs['file'], collection)
+                if self.context.get("validate_href_reachability", True):
+                    validate_href_reachability(attrs['file'], collection)
             except CoreValidationError as e:
                 errors = {'href': e.message}
                 raise serializers.ValidationError(code='payload', detail=errors)
--- a/app/stac_api/serializers/item.py
+++ b/app/stac_api/serializers/item.py
@@ -26,8 +26,8 @@ from stac_api.validators import validate_asset_name
 from stac_api.validators import validate_asset_name_with_media_type
 from stac_api.validators import validate_expires
 from stac_api.validators import validate_geoadmin_variant
-from stac_api.validators import validate_href_url
 from stac_api.validators import validate_href_reachability
+from stac_api.validators import validate_href_url
 from stac_api.validators import validate_item_properties_datetimes
 from stac_api.validators import validate_name
 from stac_api.validators_serializer import validate_json_payload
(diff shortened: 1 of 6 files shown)


boecklic (commented):
Very nice!
I share your concern regarding performance: If we do it in series then it certainly has a too severe performance impact. I'm unsure if we can easily (and should) try to implement a parallelized way to do the checks or if that brings more trouble than it solves... I still think that it would be nice to only accept URLs that actually point to sth. but maybe it's better to split this into two parts: make actually calling the external URL in `validate_href_url` configurable, disable it for now in the batch endpoint to unblock MCH but try to implement a parallelized / async check. 
I quick hack using asyncio can check 100 URLS in ~0.4-0.5 seconds (see attached file, just replace `.txt` with `.py`), but I'm not sure if we can easily integrate such an approach into the current structure.
[checkurl.txt](https://github.com/user-attachments/files/19375886/checkurl.txt)

asteiner-swisstopo:
Hi @boecklic

I agree: Let's make it work without the reachability check for the moment and add that later.

Thanks for the investigation with asyncio, that looks interesting! :sparkles: 

I found it quite difficult to add a flag controlling the behaviour into the DRF serializers, so I resorted to the `context` field. Please have a look at [my latest commit](https://github.com/geoadmin/service-stac/pull/548/commits/6fe74de4466435be44e0013fb1792588f34a700a) for the details.

@schtibe As discussed, feel free to also have a look at it now :eyes:

app/stac_api/serializers/item.py
```diff
@@ -241,6 +242,37 @@ def validate_type(self, value):
...
+                    extra={
+                        'collection': self.collection, 'attrs': attrs
+                    }
+                )
+                errors = {'href': _("Found read-only property in payload")}
+                raise serializers.ValidationError(code="payload", detail=errors)
+
+            try:
+                validate_href_url(attrs['file'], collection)
+                if self.context.get("validate_href_reachability", True):
```
schtibe, line 270: Maybe just add a little comment here why we would omit the check, so that a future reader can connect this to the bulk upload

schtibe (approved):
LGTM!

PR #551, opened 2025-03-25
PB-1515 : add big asset upload to the collection asset

sami-nouidri-swisstopo:
This adds the big asset upload to collection assets. As discussed with @boecklic, we have added collection assets to v0.9 in order to mitigate potential API Gateway issues.

For testing, I uploaded files to existing collection assets, and was able to retrieve from the s3 bucket successfully, which indicates the upload went through (locally and on DEV). E2E tests are also successful.

Note : this needs to be rebased on this [PR](https://github.com/geoadmin/service-stac/pull/545) before merging

- [x] E2E Successful

### DIFF
--- a/app/stac_api/admin.py
+++ b/app/stac_api/admin.py
@@ -476,17 +476,6 @@ class CollectionAssetAdmin(AssetUploadAdminMixin, admin.ModelAdmin):
         self.request = request  # pylint: disable=attribute-defined-outside-init
         return super().get_form(request, obj, **kwargs)
 
-    def get_urls(self):
-        urls = super().get_urls()
-        my_urls = [
-            path(
-                "<path:object_id>/change/upload/",
-                self.admin_site.admin_view(self.upload_view),
-                name=f'{self.model._meta.app_label}_{self.model._meta.model_name}_upload',
-            )
-        ]
-        return my_urls + urls
-
     def href(self, instance):
         return build_asset_href(self.request, instance.file.name)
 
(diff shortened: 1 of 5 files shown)


app/stac_api/templates/js/admin/upload.js
```diff
@@ -6,6 +6,10 @@ function setStatus(text) {
 }
 
 function getAssetUploadUrlPrefix(collection, item, asset) {
+    if (item === '') {
+        //no item means we're dealing with a collection asset
```
asteiner-swisstopo, line 10: How about expressing this in a variable name rather than a comment?

Something like

```js
is_collection_asset = item === ''
if (is_collection_asset) {
   ....
}
```

app/stac_api/admin.py
```diff
@@ -492,6 +503,40 @@ def get_fieldsets(self, request, obj=None):
...
+        # Generate the transfer URL
+        property_upload_url = reverse(
+            f'admin:{self.model._meta.app_label}_{self.model._meta.model_name}_upload',
+            args=[object_id],
+        )
+
+        # Add the property upload URL to the extra context
+        extra_context['property_upload_url'] = property_upload_url
+
+        return super().change_view(request, object_id, form_url, extra_context=extra_context)
```
asteiner-swisstopo, line 538: I see that these two functions, including the `get_urls()` above, are copy-paste of the same functions in `AssetAdmin` (except `Asset` <-> `CollectionAsset`). Have you had a look at how to avoid this duplication?

asteiner-swisstopo (approved):
I must admit that I don't really follow what most of this new code does but then I see that it's copy-paste of the existing asset upload. So looks good to me.

Would be nice if there was less duplication, maybe we can still find a better solution there.

adk-swisstopo (commented):
Once you have addressed the comments, please update the PR description with details of how you verified it works.

sami-nouidri-swisstopo:
> The admin UI doesn't allow to create a new collection asset (Error `Invalid media type ""`).

I've made the changes requested, and tested the big asset upload feature, which seems to work correctly. 

I've however encountered a peculiar case of flakyness, where **sometimes** the end-to-end tests failed when using the `debug` docker image on the k8s cluster. I could not replicate this behavior locally, and the tests pass when using the docker `prod` image.

Here's the error that would sometimes occur : 
`AssertionError: 500 not found in [200, 404] : ClientError('An error occurred (AccessDenied) when calling the DeleteObject operation: User: arn:aws:iam::839910802816:user/service-stac-tech-dev is not authorized to perform: s3:DeleteObject on resource: "arn:aws:s3:::service-stac-managed-dev-swisstopo/test-e2e-asset-upload-f2c004cd-71cc-48cb-a8f2-aadbfc5b7fd3/test-e2e-asset-upload-64857f86-108f-469f-9996-ba9c5d4d3412/test-e2e-asset-upload-0db0740d-a7d3-4fa0-8785-28454d62ae00.txt" because no identity-based policy allows the s3:DeleteObject action')`

adk-swisstopo:
> **sometimes** the end-to-end tests failed when using the `debug` docker image on the k8s cluster

Sounds like PB-1249. AFAICT the debug image was never meant to pass the end-to-end test.

PR #554, opened 2025-03-27
Fixes various CSS and JS issues

msom:
Fixes various JavaScript error messages caused by using invalid DOM elements on various pages:
- api/stac/admin/stac_api/collection/add/
- api/stac/admin/stac_api/collection/{ID}/change/
- api/stac/admin/stac_api/item/add/
- api/stac/admin/stac_api/item/{ID}/change/
- api/stac/admin/stac_api/asset/
- api/stac/admin/stac_api/asset/add/
- api/stac/admin/stac_api/asset/{ID}/change/
- api/stac/admin/stac_api/collectionasset/add/
- api/stac/admin/stac_api/collectionasset/3/change/

Github makes it really hard to recognize but the changes are all only wrapping the existing code in an additional `if ({element})`-check.

Also fixes the search bar hover text being white color on white background.

Not sure if these functionalities are even still in use.

### DIFF
--- a/app/stac_api/templates/style/hover.css
+++ b/app/stac_api/templates/style/hover.css
@@ -14,6 +14,7 @@
     top:4em;
     left:70px;
     z-index:1000;
+    color: black;
     background-color: #DDD;
     padding: 5px;
     border-radius: 4px;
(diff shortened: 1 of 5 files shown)


ltshb (approved):

app/stac_api/templates/style/hover.css
```diff
@@ -14,6 +14,7 @@
     top:4em;
     left:70px;
     z-index:1000;
+    color: black;
```
ltshb, line 17: @msom I just test it the current behavior on sys-data.dev.bgdi.ch and it seems that if your browser is set to darkmode, this will greatly help as the background of the search help text is white in this case and make the help text non readable, on non darkmode the background is grey and the text black and is perfectly readable. This change should not affect non dark mode, but maybe double check if you haven't yet

PR #557, opened 2025-04-08
PB-1531: Update packages

msom:
Updates `gunicorn` as it had [deadlock issues in the prior version](https://docs.gunicorn.org/en/stable/news.html#id1).

This also resolves the following dependabot alerts:
- https://github.com/geoadmin/service-stac/security/dependabot/75
- https://github.com/geoadmin/service-stac/security/dependabot/76
- https://github.com/geoadmin/service-stac/security/dependabot/77
- https://github.com/geoadmin/service-stac/security/dependabot/78
- https://github.com/geoadmin/service-stac/security/dependabot/79


adk-swisstopo (approved):
In principle, that looks good. However I am not sure about the actual process to upgrade these dependencies in our environments.

I also think we should switch to Django 5.2 now that it's released as it's the latest LTS version and 5.0 is going to be unsupported: https://www.djangoproject.com/download/#supported-versions But that can be done separately.

PR #559, opened 2025-04-10
Upgrade gevent.

adk-swisstopo:
There is no specific indication it will resolve our issue but the problem is related to gevent/urllib3 interaction so we might just as well try. We are already using the latest urllib3 release.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -25,7 +25,7 @@ pytest-dotenv = "*"
 pytest-cov = "*"
 
 [packages]
-gevent = "~=24.2.1"
+gevent = "~=24.11.1"
 gunicorn = "~=23.0.0"
 psycopg2-binary = "~=2.9.9"
 numpy = "~=1.26.4"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "beaf4cede1fadab83a4d1fcd94089b7f8a78ded05fb094517285a48deae8b642"
+            "sha256": "7cef900595f2faa69bb22ff2e441b279a44078c3986597dc306a930db933a825"
         },
         "pipfile-spec": 6,
         "requires": {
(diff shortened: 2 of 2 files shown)


adk-swisstopo:
This failure is due to the boto3 upgrade:
```
FAIL: test_asset_upload_list_parts (tests.tests_10.test_collection_asset_upload_endpoint.CollectionAssetUploadListPartsEndpointTestCase.test_asset_upload_list_parts)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/root/.pyenv/versions/3.12.8/lib/python3.12/unittest/case.py", line 58, in testPartExecutor
    yield
  File "/root/.pyenv/versions/3.12.8/lib/python3.12/unittest/case.py", line 634, in run
    self._callTestMethod(testMethod)
    ^^^^^^^^^^^^^^^^^
  File "/root/.pyenv/versions/3.12.8/lib/python3.12/unittest/case.py", line 589, in _callTestMethod
    if method() is not None:
    ^^^^^^^^^^^^^^^^^
  File "/codebuild/output/src3503/src/github.com/geoadmin/service-stac/app/tests/tests_10/test_collection_asset_upload_endpoint.py", line 1333, in test_asset_upload_list_parts
    self.assertEqual(size, self.asset.file_size)
    ^^^^^^^^^^^^^^^^^
  File "/root/.pyenv/versions/3.12.8/lib/python3.12/unittest/case.py", line 885, in assertEqual
    assertion_func(first, second, msg=msg)
    ^^^^^^^^^^^^^^^^^
  File "/root/.pyenv/versions/3.12.8/lib/python3.12/unittest/case.py", line 878, in _baseAssertEqual
    raise self.failureException(msg)
    ^^^^^^^^^^^^^^^^^
AssertionError: 20971520 != 20971864
```

adk-swisstopo:
Dropping boto3 upgrade from this PR as fixing the test is not trivial.

PR #561, opened 2025-04-15
PB-1531: document the workaround and enable it by default.

adk-swisstopo:


adk-swisstopo:
> I think we want these flags (if they stay for longer) be defined like the other ones in settings_prod.

Sorry, I only saw your comment after merging...

I think we want to keep the flag because we will want to remove that workaround whenever gevent will have fixed their bug.

I don't think we can generally import Django settings from within wsgi.py because (a) monkey patching needs to happen before ~everything else (b) wsgi.py is used in at least two different ways, testing that early loading stuff properly and automatically is sort of hard and there are [signs](https://github.com/geoadmin/service-stac/blob/12bb373eff0a10a5719affd57b815d2a9b7d8eff/app/wsgi.py#L47) that messing with settings upsets at least one of these two ways.

That said, we can probably make this early loading stuff cleaner and I will happily review your PR.

PR #565, opened 2025-04-16
PB-1575: make remove_expired_items more memory-efficient, probably.

adk-swisstopo:
The cron jobs running this command often runs out of memory in production. The calls to [`all()`](https://docs.djangoproject.com/en/5.2/ref/models/querysets/#all) are likely consuming a relatively large amount of memory. This change replaces them with [iterators](https://docs.djangoproject.com/en/5.2/ref/models/querysets/#iterator) which are supposed to be cheaper.

Tests still pass but I have not verified actual memory usage before/after. I propose to just push as is and monitor the situation. If it resolves the problem, great. Otherwise we can still benchmark or consider other actions later.

### DIFF
--- a/app/stac_api/management/commands/remove_expired_items.py
+++ b/app/stac_api/management/commands/remove_expired_items.py
@@ -38,7 +38,7 @@ class Handler(CommandHandler):
                     "These are likely stale, so we'll abort them"
                 )
                 uploads_in_progress.update(status=BaseAssetUpload.Status.ABORTED)
-            self.delete(item.assets.all(), 'assets')
+            self.delete(item.assets.get_queryset(), 'assets')
             self.delete(item, 'item')
             if not self.options['dry_run']:
                 self.print_success(


app/stac_api/management/commands/remove_expired_items.py
```diff
@@ -25,31 +25,31 @@ def run(self):
...
             )
             if uploads_in_progress.count() > 0:
                 self.print_warning(
                     "WARNING: There are still pending asset uploads for expired items. "
                     "These are likely stale, so we'll abort them"
                 )
                 uploads_in_progress.update(status=BaseAssetUpload.Status.ABORTED)
-            assets_length = len(assets)
-            self.delete(assets, 'assets')
+            self.delete(item.assets.all(), 'assets')
```
benschs, line 41: Would it be possible to delete the `item.assets` without a call to `all()`?

benschs (approved):
Cool. Was not aware querysets could be used like this. :+1:

app/stac_api/management/commands/remove_expired_items.py
adk-swisstopo: I found a way.

adk-swisstopo:
Comparing memory usage of `app/tests/tests_10/test_remove_expired_items.test_remove_item` with `tracemalloc` (which I discovered today thanks to https://github.com/hxu296/tariff/pull/68 ) I see 0.1% improvement in memory usage over two runs before and after (4 runs total). It does not look much but given the test is tiny (one item, two assets) I only expect to see a "saving" of at most one asset so that's consistent.

PR #568, opened 2025-04-22
PB-1575: make test larger and more flexible.

adk-swisstopo:
The remove_expired_items cronjob frequently runs out of memory in INT and PROD. As we are considering improvements, we need a way to easily test them and profile memory usage. This change adds a `remove_many_items` method that allows easily changing how many test items are considered. We also add two tests exercising that new method with 10 items and 10 assets each with and without dry run.

From there we can easily get a very basic memory profile with something like this:
```
/bin/time -f %M pipenv run python3 app/manage.py test --verbosity=2 tests.tests_10.test_remove_expired_items
```

We will also probably tweak the exact number of items and assets later to match more realistic scenarios.

### DIFF
--- a/app/tests/tests_10/test_remove_expired_items.py
+++ b/app/tests/tests_10/test_remove_expired_items.py
@@ -32,7 +32,7 @@ class RemoveExpiredItems(TestCase):
         return out.getvalue()
 
     @mock_s3_asset_file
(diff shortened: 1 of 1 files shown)


adk-swisstopo:
Superseded by https://github.com/geoadmin/service-stac/pull/569

PR #569, opened 2025-04-24
PB-1575: many improvements to remove_expired_items and its tests.

adk-swisstopo:
See the commits for details.

TL;DR:
* increase code coverage
* make it easier to add meaningful tests
* add basic profiling test
* increase test setup performance by enough that we can actually run tests with more than 100 items
* increase remove_expired_items performance by ~60% (for 1000 items)
* add safety to avoid accidentally the whole database

The actual problem might still not be full resolved but at this point the bottleneck is likely to be on the database side.


benschs (approved):
Cool

PR #570, opened 2025-04-29
PB-1615: Fix change form template variable lookup error

msom:
We currently overwrite the `change_form` template globally and check for `property_upload_url`, which is only defined for a very few admin UI views (and even there only for part of the cases). Although django renders the template without showing an error, it pollutes the logs with a stack trace.

This PR introduces a template tag to get the context variable without triggering tracebacks if it is not defined.

### DIFF
--- a/app/stac_api/templatetags/context_variables.py
+++ b/app/stac_api/templatetags/context_variables.py
@@ -1,0 +1,20 @@
+from django import template
+
+register = template.Library()
+
+
+@register.simple_tag(takes_context=True)
+def get_context_variable(context, name, default):
+    """Retrieve a context variable from the template context, returning a default value if the
+    variable is not defined.
+
+    Accessing undefined variables will still render the template, but using this function prevents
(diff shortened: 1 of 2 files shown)


app/stac_api/templatetags/context_variables.py
```diff
@@ -0,0 +1,17 @@
+from django import template
+
+register = template.Library()
+
+
+@register.simple_tag(takes_context=True)
+def get_context_variable(context, name, default):
+    """Retrieve a context variable from the template context, returning a default value if the
+    variable is not defined.
```
adk-swisstopo, line 9: Please explain in that function documentation why this is useful/needed.

adk-swisstopo (approved):
https://forum.djangoproject.com/t/proper-way-of-checking-variables-existing-in-templates/25678 suggests there exceptions are routine and best ignored. Without much Django experience I agree it nicer to avoid throwing the exception in the first place.

PR #571, opened 2025-05-05
PB-1616: Revert "PB-1615: Fix change form template variable lookup error"

adk-swisstopo:
I think https://github.com/geoadmin/infra-kubernetes/pull/801 makes https://github.com/geoadmin/service-stac/pull/570 obsolete. So let's revert it.


adk-swisstopo:
> I think it's bad practice (and a pain in the a** if you need to debug) to leave unnecessary error messages and stack traces in the log but I don't mind reverting this change.

I tend to agree but with https://github.com/geoadmin/infra-kubernetes/pull/801 we gave up on that and decided to filter out these messages instead as that's apparently the way to do it in Django. So I think keeping this extra code for nothing is worse.

PR #572, opened 2025-05-05
PB-1635 Hide expired items in search

asteiner-swisstopo:
This hides expired items in the `/search` endpoints. That this wasn't done already only became apparent when a bug prevented the cron job `cron-delete-expired` to clean up expired items properly, see PB-1575.

There are two search endpoints affected by this change:
1. [GET /search](https://sys-data.dev.bgdi.ch/api/stac/static/spec/v1/apitransactional.html#tag/STAC/operation/getSearchSTAC): Simple filtering
2. [POST /search](https://sys-data.dev.bgdi.ch/api/stac/static/spec/v1/apitransactional.html#tag/STAC/operation/postSearchSTAC): Advanced filtering in the payload.

We hide an item if has property "expires" set to a time in the past, so in principle the same as in these endpoints:
- [GET /collections/{collectionId}/items](https://sys-data.dev.bgdi.ch/api/stac/static/spec/v1/apitransactional.html#tag/Data/operation/getFeatures): Fetch features
- [GET /collections/{collectionId}/items/{featureId}](https://sys-data.dev.bgdi.ch/api/stac/static/spec/v1/apitransactional.html#tag/Data/operation/getFeature): Fetch a single feature

To discuss:
1. There is a small difference to the Item endpoints. I first check if the field is defined and then compare timestamps, not vice versa:
   - Items endpoints: `Q(properties_expires__gte=timezone.now()) | Q(properties_expires=None)`
   - Search endpoints: `Q(properties_expires=None) | Q(properties_expires__gte=timezone.now())`
   
   I don't see why you would check for `None` only after... Is this a bug?
2. To check whether an expiry date is in the past, the reference time is the time at which the query is processed on our server. As storage and filtering is done on the same server, I don't see problems with time zones.

### DIFF
--- a/app/config/settings_prod.py
+++ b/app/config/settings_prod.py
@@ -365,9 +365,7 @@ DISALLOWED_EXTERNAL_ASSET_URL_SCHEMES = env.list(
 )
 
 # Feature flag to hide/show expired items in the search endpoint
-FEATURE_HIDE_EXPIRED_ITEMS_IN_SEARCH_ENABLED = env(
-    'FEATURE_HIDE_EXPIRED_ITEMS_IN_SEARCH_ENABLED', bool, default=True
-)
+HIDE_EXPIRED_ITEMS_IN_SEARCH = env('HIDE_EXPIRED_ITEMS_IN_SEARCH', bool, default=True)
 
 # These are the default values from Django as per
 # https://docs.djangoproject.com/en/5.1/ref/settings/
--- a/app/config/settings_prod.py
+++ b/app/config/settings_prod.py
@@ -364,6 +364,11 @@ DISALLOWED_EXTERNAL_ASSET_URL_SCHEMES = env.list(
     'DISALLOWED_EXTERNAL_ASSET_URL_SCHEMES', default=['http']
 )
 
+# Feature flag to hide/show expired items in the search endpoint
+FEATURE_HIDE_EXPIRED_ITEMS_IN_SEARCH_ENABLED = env(
+    'FEATURE_HIDE_EXPIRED_ITEMS_IN_SEARCH_ENABLED', bool, default=True
+)
+
 # These are the default values from Django as per
 # https://docs.djangoproject.com/en/5.1/ref/settings/
 # We add them here so they can be changed through environment variables.
--- a/app/stac_api/views/filters.py
+++ b/app/stac_api/views/filters.py
@@ -1,0 +1,9 @@
+from django.db.models import Q
+from django.utils import timezone
+
+
+def create_is_active_filter():
+    """
+    Create a filter to check if the item is not expired.
+    """
+    return Q(properties_expires__gte=timezone.now()) | Q(properties_expires=None)
--- a/app/tests/tests_10/test_search_endpoint.py
+++ b/app/tests/tests_10/test_search_endpoint.py
@@ -514,6 +514,13 @@ class SearchEndpointTestCaseTwo(StacBaseTestCase):
         self.factory.create_item_sample(
             self.collection, name='item-expired', db_create=True, properties_expires=tomorrow
         )
+        in_a_week = timezone.now() + timedelta(days=7)
+        self.factory.create_item_sample(
+            self.collection,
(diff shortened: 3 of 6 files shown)


adk-swisstopo:
> I don't see why you would check for None only after... Is this a bug?

Why would the order matter?
That said, maybe there should be a common get_visible_items method that both endpoints call?

adk-swisstopo:
While I think this looks good, I generally think it's nicer to guard this kind of behaviour change behind a flag we can control from an environment variable. This allows to roll back easily in case it causes a problem. See FEATURE_AUTH_ENABLE_APIGW for an example.

boecklic (approved):

asteiner-swisstopo:
Hey @adk-swisstopo,

Thanks for those comments. As discussed:
1. You are right about the order: Does not matter as it's translated into an SQL query, it's not evaluated in Python.
2. Extract common functionality: I did that, at least the simple part of it. The function name is a bit clunky but probably already better to have that in one place instead of 4.
3. Feature flag: Makes sense, I have added flag `FEATURE_HIDE_EXPIRED_ITEMS_IN_SEARCH_ENABLED `. I add a ticket to remove that again at some point :sweat_smile: 

Can you have another look please? :innocent:

app/config/settings_prod.py
```diff
@@ -364,6 +364,11 @@ def get_logging_config():
     'DISALLOWED_EXTERNAL_ASSET_URL_SCHEMES', default=['http']
 )
 
+# Feature flag to hide/show expired items in the search endpoint
+FEATURE_HIDE_EXPIRED_ITEMS_IN_SEARCH_ENABLED = env(
```
adk-swisstopo, line 368: "_ENABLED" is redundant.

app/tests/tests_10/test_search_endpoint.py
```diff
@@ -507,6 +509,34 @@ def test_datetime_invalid_format_query_get(self):
...
+            self.collection, name='item-expired', db_create=True, properties_expires=tomorrow
+        )
+
+        after_tomorrow = timezone.now() + timedelta(days=2)
+        with patch.object(timezone, "now", return_value=after_tomorrow):
+            response = self.client.get(self.path)
+
+        self.assertStatusCode(200, response)
+        feature_ids = [feature["id"] for feature in response.json()['features']]
+        self.assertNotIn('item-expired', feature_ids)
```
adk-swisstopo, line 524: Can you also please test that unexpired feature(s) do show up? As it is, I think the tests will pass if we never return any item that has a non-NULL value in properties_expires.

app/stac_api/views/filters.py
adk-swisstopo, line 1: I am sad we don't already have something for this. It feels like a bigger refactoring would be nice (collection__published=True has the same problem for example). But that's out of scope for this PR.

app/config/settings_prod.py
asteiner-swisstopo: Not if you read it like this: `FEATURE_X_ENABLED` where X is `HIDE_EXPIRED_ITEMS_IN_SEARCH`.

But I see that it is confusing. So let me shorten it to this:

```txt
HIDE_EXPIRED_ITEMS_IN_SEARCH
```

app/tests/tests_10/test_search_endpoint.py
adk-swisstopo: I have seen no other test that tests that. If there is already one, it's probably fine.

app/config/settings_prod.py
adk-swisstopo: It looks like our style guide does not cover this. We should probably come up with an explicit naming convention for feature flags specifically and maybe for Python global variables in general.

Meanwhile I am fine with FEATURE_HIDE_EXPIRED_ITEMS_IN_SEARCH_ENABLED, FEATURE_HIDE_EXPIRED_ITEMS_IN_SEARCH and HIDE_EXPIRED_ITEMS_IN_SEARCH.

adk-swisstopo (approved):

PR #573, opened 2025-05-07
PB-1575: don't attempt to delete externally hosted S3 objects.

adk-swisstopo:
Every time we delete an `Asset` or `CollectionAsset`, we make a call to the corresponding S3 `DeleteObjects` method. For objects that are hosted externally (i.e. not in an S3 bucket we control), this call always fails silently and we can just as well not do it.

With this change, we stop attempting to call `DeleteObjects` for externally hosted `Assets` and `CollectionAssets`, thus speeding up deletion.


benschs (approved):

boecklic (approved):
:+1:

PR #574, opened 2025-05-07
PB-1575: make deletion batch size configurable.

adk-swisstopo:
One `delete()` per object is very slow. One `delete()` for all the objects consumes a lot of memory and fails in an all-or-nothing mode (i.e. when it fails mid-way, nothing gets deleted).

With this change, we split the objects into batches of configurable size and run `delete()` over each batch. It consumes a ~constant amount of memory, is faster than having a batch size of 1 and is more robust. I.e. when it fails mid-way, it will most likely have deleted at least some objects and the next run has that much less work to do.

There will be cases where expired items are partially deleted (i.e. some of their assets are deleted but not all). I don't expect that to be a problem for our use case.

I tested this manually on my laptop with a old copy of the INT database also hosted on my laptop. I set `shm_size: 128m` for the Posgresql Docker container. I ran a full vacuum after loading the data but before running `remove_expired_items`. The dataset contains 2 622 231 expired assets (out of 4 613 212 total) and 2 622 251 expired items (out of 4 010 706 total). 1 247 448 of these expired assets are marked as "not external", meaning there is also a call to S3 for each of them. The S3 endpoint is minio also running on my laptop and not actually hosting any file (i.e. every `DeleteObjects` returned `204 No Content`).

|test name         |max RSS                         |wallclock duration  |kernel CPU time|user CPU time |
|-------------------------|-------------------------------|----------------------------------------------------|-----------------------|-------------------------|
|*batch_size=5,000,000          | >= 6 522 268 Kb           |   > 5h37m18s                                         | >166.76 s         | >1 645.88 s     |
|batch_size=10,000                 |208 960 Kb                     |2h21m57s                                             |   217.98 s |           1 802.70 s|
|**no OFFSET, batch_size=10,000| 208 636 Kb|2h16m43s|211.92 s|1 774.26 s|
|***batch_size=1                           |>= 129 800 Kb            |~120 days (estimated)               |?                    | ?                  |

*interrupted after 5h37. It was still deleting assets and did not have time to proceed to items. Nothing was deleted at all as no transaction completed.
**Last minute [change](https://github.com/geoadmin/service-stac/pull/574/commits/fbf64e705bfab9461425002d222c1084c3e35353) based on [this discussion](https://forum.djangoproject.com/t/incremental-bulk-deletion/40833/2).
*** interrupted after 14 minutes and 402 assets deleted.

In Kubernetes I expect similar memory usage. Duration may vary in either direction as most of the time is spent waiting on other processes (database and file storage).


boecklic (approved):
Nice :ok_hand:

PR #578, opened 2025-05-19
PB-1590: item search conformance - #minor

benschs:
Add item search-conformance. We still are missing a couple of points to be 100% conform to the spec, however for our use cases it is close enough and still helpful to add the item-search to our conformance list.
Missing conformance:
- Return Content-Type Header is not `application/geo+json`. This would be a breaking change
- Filter by 3d bbox (bbox with 6 values). Not a trivial change.

Changes:
- If limit is larger than the configured max_limit, use max_limit instead of returning status 400
  - According to https://github.com/radiantearth/stac-api-spec/blob/release/v1.0.0/item-search/README.md#query-parameter-table
- Datetime is case insensitive
- Filter by intersect in GET request, as is the case with a POST request
- If bbox has 6 values, return status 400 and custom error message informing that we do not support 3d bboxes
- Add item-search conformance link (https://api.stacspec.org/v1.0.0/item-search) to landing page for v1


app/tests/tests_10/test_generic_api.py
```diff
@@ -100,10 +100,7 @@ def test_invalid_limit_query(self):
                                  msg='Unexpected error message')
 
                 response = self.client.get(f"/{STAC_BASE_V}/{endpoint}?limit=1000")
-                self.assertStatusCode(400, response)
-                self.assertEqual(['limit query parameter too big, must be in range 1..100'],
-                                 response.json()['description'],
-                                 msg='Unexpected error message')
+                self.assertStatusCode(200, response)
```
ltshb, line 103: I would add a comment here that the limit is corrected to the max, and maybe check that the limit is corrected also to the configured max.

app/tests/tests_09/test_generic_api.py
```diff
@@ -100,10 +100,7 @@ def test_invalid_limit_query(self):
                                  msg='Unexpected error message')
 
                 response = self.client.get(f"/{STAC_BASE_V}/{endpoint}?limit=1000")
-                self.assertStatusCode(400, response)
-                self.assertEqual(['limit query parameter too big, must be in range 1..100'],
-                                 response.json()['description'],
-                                 msg='Unexpected error message')
+                self.assertStatusCode(200, response)
```
ltshb, line 103: Same here check that the result don't have more than the max allowed limit and add a comment.

ltshb (approved):
Nice just some minors improvement in unittest.

benschs:
Thanks @ltshb. I could not find a way to set the MAX_PAGE_SIZE for the test case as it is part of the Rest Framework, so I added a test creating more than 100 items.

app/tests/tests_10/test_generic_api.py
```diff
@@ -99,8 +99,29 @@ def test_invalid_limit_query(self):
...
+            self.assertStatusCode(200, response)
+            # test setup creates 3 collections
+            self.assertEqual(len(response.json()['collections']), 3)
+        endpoint = f'collections/{self.collections[0]["name"]}/items'
+        with self.subTest(endpoint=endpoint):
+            # If limit is greater than max_limit, we expect max_limit to be used.
+            response = self.client.get(f"/{STAC_BASE_V}/{endpoint}?limit=1000")
+            self.assertStatusCode(200, response)
+            page_1 = response.json()
+            self.assertEqual(len(page_1['features']), 100)
```
hansmannj, line 117: dito

app/tests/tests_09/test_generic_api.py
```diff
@@ -99,8 +99,29 @@ def test_invalid_limit_query(self):
...
+            self.assertStatusCode(200, response)
+            # test setup creates 3 collections
+            self.assertEqual(len(response.json()['collections']), 3)
+        endpoint = f'collections/{self.collections[0]["name"]}/items'
+        with self.subTest(endpoint=endpoint):
+            # If limit is greater than max_limit, we expect max_limit to be used.
+            response = self.client.get(f"/{STAC_BASE_V}/{endpoint}?limit=1000")
+            self.assertStatusCode(200, response)
+            page_1 = response.json()
+            self.assertEqual(len(page_1['features']), 100)
```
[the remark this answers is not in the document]
benschs: Not sure about a message. I found the [context extension](https://github.com/stac-api-extensions/context) that allows to set the limit requested and actually returned. I haven't found an easy/standard way of setting other messages.

benschs:
I somehow messed up the rebase / cherry-pick to the master branch. It seemed easier to create a new branch and PR: https://github.com/geoadmin/service-stac/pull/579
Sorry for the inconvenience :(

PR #579, opened 2025-05-20
PB-1590: Added item-search conformance - #minor

benschs:
Add item search-conformance. We still are missing a couple of points to be 100% conform to the spec, however for our use cases it is close enough and still helpful to add the item-search to our conformance list.
Missing conformance:
- Return Content-Type Header is not `application/geo+json`. This would be a breaking change
- Filter by 3d bbox (bbox with 6 values). Not a trivial change.

Changes:
- If limit is larger than the configured max_limit, use max_limit instead of returning status 400
  - According to https://github.com/radiantearth/stac-api-spec/blob/release/v1.0.0/item-search/README.md#query-parameter-table
- Datetime is case insensitive
- Filter by intersect in GET request, as is the case with a POST request
- If bbox has 6 values, return status 501 and custom error message informing that we do not support 3d bboxes
- Add item-search conformance link (https://api.stacspec.org/v1.0.0/item-search) to landing page for v1

### DIFF
--- a/app/stac_api/utils.py
+++ b/app/stac_api/utils.py
@@ -308,6 +308,8 @@ def harmonize_post_get_for_search(request):
             query_param['ids'] = query_param['ids'].split(',')  # to array
         if 'collections' in query_param:
             query_param['collections'] = query_param['collections'].split(',')  # to array
+        if 'intersects' in query_param:
+            query_param['intersects'] = json.loads(query_param['intersects'])
 
         # Forecast properties can only be filtered with method POST.
         # Decision was made as `:` need to be url encoded and (at least for now) we do not need to
--- a/app/stac_api/utils.py
+++ b/app/stac_api/utils.py
@@ -51,7 +51,7 @@ def isoformat(date_time):
 def fromisoformat(date_time):
     '''Return a datetime object from a isoformated datetime string
     '''
-    return datetime.fromisoformat(date_time.replace('Z', '+00:00'))
+    return datetime.fromisoformat(date_time.upper().replace('Z', '+00:00'))
 
 
 def utc_aware(date_time):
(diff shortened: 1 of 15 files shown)


app/stac_api/utils.py
```diff
@@ -456,6 +458,8 @@ def geometry_from_bbox(bbox):
         ValueError, IndexError, GDALException
     '''
     list_bbox_values = bbox.split(',')
+    if len(list_bbox_values) == 6:
+        raise ValueError('3-dimensional bbox is currently not supported')
```
hansmannj, line 462: what response status code would be connected with this exception?
Guess we should make sure (also with a corresponding unit test for this case?) to get a 501 "not implemented", according to https://datatracker.ietf.org/doc/html/rfc7231#section-6.6.2 and also after discussions with https://github.com/m-mohr

app/tests/tests_09/test_generic_api.py
```diff
@@ -99,11 +99,29 @@ def test_invalid_limit_query(self):
...
+            self.assertStatusCode(200, response)
+            # test setup creates 3 collections
+            self.assertEqual(len(response.json()['collections']), 3)
+        endpoint = f'collections/{self.collections[0]["name"]}/items'
+        with self.subTest(endpoint=endpoint):
+            # If limit is greater than max_limit, we expect max_limit to be used.
+            response = self.client.get(f"/{STAC_BASE_V}/{endpoint}?limit=1000")
+            self.assertStatusCode(200, response)
+            page_1 = response.json()
+            self.assertEqual(len(page_1['features']), 100)
```
hansmannj, line 117: probably a `msg` for this case might make sense

hansmannj (approved):
Nice, thanks! :rocket:
Only some minor comments :wink:

app/stac_api/utils.py
benschs: I added a new exception to return status code 501.

app/tests/tests_09/test_generic_api.py
benschs: Sorry I initially misunderstood you. :sweat_smile:
I added a message to the assert.

PR #582, opened 2025-05-22
PB-1704: Handle application/geo+json requests

msom:
Adds a new DRF renderer that accepts `geojson` request, either via `Accept: application/geo+json` header or using the `format=geojson` query parameter, and returns JSON as usual but `Content-Type: application/geo+json`. Default/unspecified is still `application/json`.


hansmannj (approved):
:rocket:
Nice, thanks a lot!

hansmannj:
Just tested the search on https://sys-data.dev.bgdi.ch/#/search and the new adaptions seem to make the new STAC browser happy.
Searching seems to work now, nice! :tada:

PR #593, opened 2025-06-30
fix: update Access-Control-Allow-Headers for edge using `Accept-Language` header longer 128

Rdataflow:
currently stac-service in edge with strict mode enabled refuses to load from other pages due to CORS error.

**repro of error case**
1. use edge in strict mode (i.e. federal default setup)
2. open https://radiantearth.github.io/stac-browser/#/external/data.geo.admin.ch/api/stac/v0.9/?.language=en
3. see error message 
image
4. check console log
`
Access to XMLHttpRequest at 'https://data.geo.admin.ch/api/stac/v0.9/' from origin 'https://radiantearth.github.io' has been blocked by CORS policy: Request header field accept-language is not allowed by Access-Control-Allow-Headers in preflight response.
`

**measures to enable edge in strict mode**
- [ ] Allow all the occuring request headers 
  - [x] i.e. `Accept-Language` etc. (this PR)
  - [ ] more may be necessary
- [ ] alternatively: allow `*`

HTH to enable data.geo.admin.ch STAC-service for broader internal & public use 😄

### DIFF
--- a/app/middleware/cors.py
+++ b/app/middleware/cors.py
@@ -33,6 +33,6 @@ class CORSHeadersMiddleware:
         if request.path in (f'/{STAC_BASE}/v0.9/search', f'/{STAC_BASE}/v1/search'):
             allow_methods.append('POST')
         response['Access-Control-Allow-Methods'] = ','.join(allow_methods)
-        response['Access-Control-Allow-Headers'] = 'Content-Type,Accept'
+        response['Access-Control-Allow-Headers'] = 'Content-Type,Accept,Accept-Language'
 
         return response
--- a/app/tests/tests_09/base_test.py
+++ b/app/tests/tests_09/base_test.py
@@ -129,7 +129,7 @@ class StacTestMixin:
 
     def assertCors(self, response):  # pylint: disable=invalid-name
         for header, value in {
-            'Access-Control-Allow-Headers': 'Content-Type,Accept',
+            'Access-Control-Allow-Headers': 'Content-Type,Accept,Accept-Language',
             'Access-Control-Allow-Methods': 'GET,HEAD',
             'Access-Control-Allow-Origin': '*'
         }.items():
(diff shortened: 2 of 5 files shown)


app/middleware/cors.py
```diff
@@ -33,6 +33,6 @@ def __call__(self, request):
         if request.path in (f'/{STAC_BASE}/v0.9/search', f'/{STAC_BASE}/v1/search'):
             allow_methods.append('POST')
         response['Access-Control-Allow-Methods'] = ','.join(allow_methods)
-        response['Access-Control-Allow-Headers'] = 'Content-Type,Accept'
+        response['Access-Control-Allow-Headers'] = 'Content-Type,Access,Accept-Language'
```
boecklic, line 36: :+1: for adding `Accept-Language`, but I don't understand why you suggest to remove `Accept` and add `Access`? I don't see `Access` header being sent by Edge (and to my knowledge it's not a standard/common header)

boecklic (changes requested):
Thanks for your PR.
I cannot reproduce what you describe on a default federal machine but we can safely add the requested header if this helps (taking the specs strictly https://developer.mozilla.org/en-US/docs/Glossary/CORS-safelisted_request_header this shouldn't be necessary, but if it solves your problem we'll add it).

Rdataflow:
@boecklic so sure let's try this PR
NB: if it helps I can offer a screen session 👍

Rdataflow:
**FTR**
default `Accept-Language` header is `en,de;q=0.7,de-DE;q=0.7,en-GB;q=0.7,en-US;q=0.7,de-CH;q=0.7,rm;q=0.7,ar;q=0.2,es;q=0.2,fr;q=0.2,it;q=0.2,ro;q=0.2,ja;q=0.2,pt;q=0.2,id;q=0.2`

this is longer than 128 thus you cannot rely on CORS simple request assumption [documented here](https://github.com/mdn/content/blob/84ba257fb5e776926bb6dda340b0aa88300246b3/files/en-us/glossary/cors-safelisted_request_header/index.md?plain=1#L30)

That's why proper CORS implementation is actually needed for such cases

boecklic:
@Rdataflow : could you pls sync your fork and rebase your branch on `develop`? We enforce branches to be up to date with the base branch before merging. I can merge then, thanks!

Rdataflow:
@boecklic rebase is done - HTH 👍

boecklic (approved):

Rdataflow:
@boecklic LGTM using dev endpoint 👍

PR #605, opened 2025-09-18
New Release v2.2.0 - #minor

adk-swisstopo:


adk-swisstopo:
> Didn't @ltshb mentioned something about filling out the PR description because of changelogs?

I don't see anything in https://github.com/geoadmin/doc-guidelines/blob/master/VERSIONING_RELEASE.md#new-semver-release
What kind of description is expected that would not be redundant with the list of commits?

PR #644, opened 2026-05-29
PB-1403: test and refactor the timestamp-updating trigger.

adk-swisstopo:
Updating any field of an Asset triggers an update of its timestamp as stored in the `updated` field. There is no test for that trigger.

In PB-1403 we want to change this behaviour and for the trigger to become conditional.

This change updates the trigger generator to allow it to be conditional. It also adds tests. There is no actual change in behaviour. A future change will make use of this new parametre.


benschs (approved):
👍

PR #648, opened 2026-06-03
PB-1403: Only update asset timestamp when the underlying file has changed.

adk-swisstopo:
The `updated` field of `Asset` objects is updated when any field is updated. This causes unecessary churn and confusion when hidden metadata are updated.

This change ensures the `updated` field is only updated when the `file` or `checksum_multihash` field is updated. This also adds a test to verify the timestamp is not updated when the description is updated.

This PR is branched from [another refactoring PR](https://github.com/geoadmin/service-stac/pull/647) which is being reviewed separately.


benschs (approved):

boecklic (approved):
👌

PR #649, opened 2026-06-03
PB-1403: only update Item timestamp when Asset timestamp is updated.

adk-swisstopo:
The `updated` field of `Item` object is updated whenever any associated `Asset` object is updated. This causes unecessary churn and confusion when hidden metadata are updated.

This change ensures the `Item.updated` field is only updated when the `Asset.updated` field is updated. This broke a test that assumed that updating `Asset.title` would update `Collection.updated`. I have modified the test to match the new behaviour.


benschs (approved):
👍

msom (approved):

adk-swisstopo:
This PR is incorrect. Withdrawing in favour of https://github.com/geoadmin/service-stac/pull/652 and https://github.com/geoadmin/service-stac/pull/653

PR #651, opened 2026-06-09
PB-1403: Always update Asset.etag when Asset is updated.

adk-swisstopo:
In https://github.com/geoadmin/service-stac/pull/648 we modified Asset triggers such that `etag` and `updated` fields are only updated when the underlying file is modified. However we later determined that `etag` should be updated when any field is updated.

This change splits the trigger into two: one for the timestamp and one for the ETag. We make the ETag trigger unconditional and update the tests accordingly.


boecklic (approved):
Nice 👌

PR #653, opened 2026-06-11
PB-1403: Only update parent timestamp when child timestamp is updated.

adk-swisstopo:
The triggers generated by `child_triggers` are meant to establish a child/parent relationship in which the timestamp and ETag of the parent is updated when the child is updated. It currently updates the timestamp whenever any field in the child is updated.

As part of PB-1403 we want to only update the timestamps when the underlying files are modified. In a previous change we already updated the children to only update their timestamp when the underlying file is modified. With this change we modify the trigger generator to always update the ETag but to only update the timestamp when the timestamp of the child is updated.

This changes the behaviour of the following relationships:
- Asset/Item
- Item/Collection
- CollectionAsset/Collection
- CollectionLink/Collection
- ItemLink/Item
- Provider/Collection

Most notably, the child in the last three relationships have no `updated` field. Consequently, they never update the timestamp of the parent. Only the ETag.

This change also updates the tests to reflect these changes in behaviour and improves documentation for the `child_triggers` and `auto_variables_triggers` trigger generators.

This PR is based off https://github.com/geoadmin/service-stac/pull/652 which is being reviewed separately.


boecklic (approved):
👌

PR #655, opened 2026-06-18
PB-1608 Reject assets in putFeature payload #patch

asteiner-swisstopo:
This fixes an inconsistency in the PUT endpoint to create/update a feature ([putFeature](https://sys-data.dev.bgdi.ch/api/stac/static/spec/v1/apitransactional.html#tag/Data-Management/operation/putFeature)). The API docs list "assets" as required for the request payload but the endpoint returns a 500 if it is actually included.

Assets in a `putFeature` payload passed validation but caused a 500 because update_or_create rejected it. Since assets are managed through the dedicated asset endpoints, they are now explicitly rejected at the serializer level with a 400.

To that end, a separate `ItemDetailSerializer` was created. We need a separate one from `ItemSerializer` because the POST endpoint `bulkCreateItems` still needs to be able to have assets in the payload.

The API docs are updated accordingly: "assets" is removed from the request body schema. As before, the request sample payload does not include "assets".

image

### DIFF
--- a/app/stac_api/serializers/item.py
+++ b/app/stac_api/serializers/item.py
@@ -467,6 +467,15 @@ class ItemSerializer(NonNullModelSerializer, UpsertModelSerializerMixin):
         return attrs
 
 
+class ItemDetailSerializer(ItemSerializer):
+    '''ItemSerializer variant used by the single-item detail view.
+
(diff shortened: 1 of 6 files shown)


boecklic (approved):
Thanks for fixing. Interesting that this hasn't caused troubles before...

PR #656, opened 2026-06-24
PB-2354: Add CF extension fields

asteiner-swisstopo:
This adds [CF extension](https://github.com/stac-extensions/cf) fields `cf:standard_names` and `unit` to a STAC `Item`.

Example:

```json
"properties": {
                ...
                "cf:standard_name": "air_temperature",
                "unit": "K",
            },
```

It follows the example of the forecast extension properties. So we add the fields to:

1. ... the model and the serializers, so it's available in the Item/Feature endpoints.
2. ... the admin UI.
3. ... the OpenAPI specs.

image

The more peculiar bits:

1. Add the fields to the [POST /search endpoint](https://data.geo.admin.ch/api/stac/static/spec/v1/apitransactional.html#tag/STAC/operation/postSearchSTAC). However, as for the forecast extension fields, they are excluded from the GET request because of the colon `:` character in the URL.
2. List the extension only if one of the two fields is present. For the forecast extension, this is only done for collection `ch.meteoschweiz.ogd-forecasting-icon`.
3. Add the fields to the e2e test input. Not strictly necessary as the fields are optional but I figured it makes sense to keep this example representative of the DB schema.

As discussed:

1. There is no additional validation like checking if `cf:standard_name` is from the [CF Standard Name Table](https://cfconventions.org/Data/cf-standard-names/current/build/cf-standard-name-table.html).
2. `cf:cell_methods` / ` description` are left out on purpose. 
3. As the list of properties in the OpenAPI specs is already quite long, we might want to group them by extension using the "Any of" feature. I propose we do this in the follow-up task [PB-2355](https://swissgeoplatform.atlassian.net/browse/PB-2355).

:question:  **To be clarified:** In https://github.com/geoadmin/service-stac/pull/653 a trigger [`update_collection_child_trigger_etag_only`](https://github.com/geoadmin/service-stac/blob/develop/app/stac_api/migrations/0076_remove_asset_add_item_child_trigger_and_more.py) was added to update the ETag if a field changed. I am not familiar with how this works, Do I have to add another trigger for the new properties?

[PB-2355]: https://swissgeoplatform.atlassian.net/browse/PB-2355?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/stac_api/models/item.py
```diff
@@ -197,6 +200,17 @@ class Meta:
         "\"unknown\" in case it's missing."
     )
 
+    cf_standard_name = models.CharField(
```
boecklic, line 203: Maybe add a comment that this two fields are related to the CF extension.

app/stac_api/serializers/item.py
```diff
@@ -377,6 +384,9 @@ def get_stac_extensions(self, obj):
         # keeping the information on collection object itself
         if obj.collection.name.startswith('ch.meteoschweiz.ogd-forecasting-icon'):
             extensions.append("https://stac-extensions.github.io/forecast/v0.2.0/schema.json")
+        # Add CF extension when CF properties are present
+        if obj.cf_standard_name is not None or obj.unit is not None:
+            extensions.append("https://stac-extensions.github.io/cf/v0.2.0/schema.json")
```
boecklic, line 389: as mentioned here: https://swissgeoplatform.atlassian.net/browse/PB-2355 I'd solve this in a generic way by carrying this information on collection level. I think this could simplify / generalize other parts as well.

boecklic (changes requested):
Thanks a lot. 
I'd generalize handling of extension, see comment below.

asteiner-swisstopo:
On hold until this is merged: https://github.com/geoadmin/service-stac/pull/658

asteiner-swisstopo:
Finally adapted to the changes made in #658: Can you have another look please @boecklic?

To test manually:

1. Run the STAC API server: `make setup && make serve`
2. In [the Django admin UI](http://127.0.0.1:8000/api/stac/admin/authtoken/tokenproxy/), add a token, here for example `b5c95f9615d610635c88226d03344fee8ea2ea67`.
3. Create a collection using that token:

   ```sh
   curl -X PUT \
     "http://127.0.0.1:8000/api/stac/v1/collections/cf-collection" \
     -H "Content-Type: application/json" \
     -H "Authorization: Token b5c95f9615d610635c88226d03344fee8ea2ea67" \
     -d '{
       "id": "cf-collection",
       "description": "A test collection",
       "license": "Test license"
     }'
   ```
4. Enable the CF extension in the Admin UI: 
   image

Then create a feature with the [putFeature](https://data.geo.admin.ch/api/stac/static/spec/v1/apitransactional.html#tag/Data-Management/operation/putFeature) endpoint (:warning: declare the CF extension in `stac_extensions`!):

```sh
curl -X PUT \
  "http://127.0.0.1:8000/api/stac/v1/collections/cf-collection/items/cf-item-1" \
  -H "Content-Type: application/json" \
  -H "Authorization: Token b5c95f9615d610635c88226d03344fee8ea2ea67" \
  -d '{
    "id": "cf-item-1",
    "geometry": {
      "type": "Point",
      "coordinates": [8.5417, 47.3769]
    },
    "properties": {
        "title": "Feature title",
        "cf:standard_name": "air_temperature",
        "unit": "K"
    },
    "links": [],
    "stac_extensions": ["https://stac-extensions.github.io/cf/v1.0.0/schema.json"]
  }'
```

Which can also be retrieved through the [POST /search endpoint](https://data.geo.admin.ch/api/stac/static/spec/v1/apitransactional.html#tag/STAC/operation/postSearchSTAC) as with the `forecast:*` properties:

```sh
curl -X POST \
  "http://127.0.0.1:8000/api/stac/v1/search" \
  -H "Content-Type: application/json" \
  -H "Authorization: Token b5c95f9615d610635c88226d03344fee8ea2ea67" \
  -d '{
    "cf:standard_name": "air_temperature"
  }'
```

asteiner-swisstopo:
From offline discussion with @boecklic regarding [POST /search](https://data.geo.admin.ch/api/stac/static/spec/v1/apitransactional.html#tag/STAC/operation/postSearchSTAC):

Currently, to query `/search` for a field in `properties`, you need to send a payload like this

```json
"query": {
    "title": {
       "eq": "CF Collection 1"
    }
}
```

For the other fields, you can just do

```json
{
    "datetime": "2018-02-12T00:00:00Z/2018-03-18T12:31:12Z"
}
```

This is in accordance with the [query extension](https://github.com/stac-api-extensions/query). However, currently "query" only accepts the following fields:
- title
- created
- updated

For the `forecast:*` properties, it works as for the other fields:

```json
{
    "forecast:variable": "ALB_RAD"
}
```

But this is a bug! It should work like this:

```json
"query": {
    "forecast:variable": {
       "eq": "ALB_RAD"
    }
}
```

So for the CF fields, we should do it correctly from the start: They need to be queried through the "query" field. We agreed to do that in the scope of this ticket.

Other possible follow-ups identified:

1. Also move the `forecast:*` properties into the "query" field for [POST /search](https://data.geo.admin.ch/api/stac/static/spec/v1/apitransactional.html#tag/STAC/operation/postSearchSTAC).
2. The [/conformance endpoint](https://data.geo.admin.ch/api/stac/v1/conformance) does not list the [query extension](https://github.com/stac-api-extensions/query) even though it is supported (albeit partly non-compliant).
3. The [query extensions docs](https://github.com/stac-api-extensions/query) say "It is recommended to implement the [Filter Extension](https://github.com/stac-api-extensions/filter) instead of the Query Extension.": We might want to do that in the course of the transition to SWISSGEO.

asteiner-swisstopo:
Update: To filter by a CF field, you need to go through the `"query"` parameter in the [POST /search endpoint](https://data.geo.admin.ch/api/stac/static/spec/v1/apitransactional.html#tag/STAC/operation/postSearchSTAC). For example:

```json
"query": {
    "cf:standard_name": {
       "eq": "air_temperature"
    }
}
```

This is now also reflected in the OpenAPI docs:
pb-2354-cf-extension-openapi-specs-search

:warning: In the course of implementing this, I found a bug in the "query" parameter. It only takes into account the first field passed to it, the other are ignored. So, for example, here the `"unit"` field is ignored:

```json
"query": {
    "cf:standard_name": {
       "eq": "air_temperature"
    },
    "unit": {
       "eq": "K"
    }
}
```

As agreed with @boecklic: We ignore this for the moment as POST /search does not seem to be used a lot. Instead, we fix this properly when moving service-stac to SWISSGEO by replacing the Query Extension with the Filter Extension. This is also what the Query extension [recommends](https://github.com/stac-api-extensions/query):

> It is recommended to implement the [Filter Extension](https://github.com/stac-api-extensions/filter) instead of the Query Extension. Filter Extension is more well-defined, more expressive, and uses the standardized CQL2 query language instead of the proprietary language defined here. There is no plan to deprecate this extension, but it is also unlikely to see any further refinement or changes.

Impact on the [geoadmin STAC Browser](https://github.com/geoadmin/stac-browser): Probably none - the Query Extension is not supported as far as I can see. The latest STAC Browser (not the geoadmin fork) is [compatible with CQL](https://github.com/radiantearth/stac-browser/blob/main/docs/code-generators.md#search-scenarios-and-cql-mode), so I understand that it rather supports the [Filter Extension](https://github.com/stac-api-extensions/filter) instead of the (deprecated) Query Extension.

boecklic (approved):
🎉

app/stac_api/models/item.py
asteiner-swisstopo: Good point, I added a short comment for both.

app/stac_api/serializers/item.py
asteiner-swisstopo: Addressed here: https://github.com/geoadmin/service-stac/pull/658

PR #661, opened 2026-08-20
chore: Fix flaky unit test

asteiner-swisstopo:
This test fails occasionally with

> Property expires can't be in the past.

because sometimes it may take more than 100ms between getting the current time and validating the item.

To mitigate that, I increased the delta from 100ms to 10m.


ltshb (commented):
Why 10min ? Sounds like a lot more than 100ms,  would it not be better to simply increase it a little bit, like 1-3s. 100ms is very short for python creation in DB but it should be able to do it in less than 1min.

PR #662, opened 2026-08-26
PB-2354: OpenAPI: Group properties by extension

asteiner-swisstopo:
Changed the schema of "properties" to group the fields by STAC extension:

- Core Properties
    - created
    - datetime
    - start_datetime
    - end_datetime
    - updated
    - title
- Timestamps extension
    - expires
- Forecast extension
    - forecast:reference_datetime
    - forecast:horizon
    - forecast:duration
    - forecast:variable
    - forecast:perturbed
- CF extension
    - cf:standard_name
    - unit

With the `anyOf`, SwaggerUI renders these groups into groups that can be clicked. That should make the understanding easier.

pb-2354-openapi-properties-grouped-by-extension

On the other side, technically the OpenAPI specs are now incorrect: As it is now, you could, for example, have this:

```json
{
    "properties":
    "expires": "2018-02-12T23:20:50Z"
}
```

while in fact some of the core properties are always mandatory and you would rather have something like this:

```json
{
    "properties":
    "expires": "2018-02-12T23:20:50Z",
}
```

Another sideeffect of this change is that the example payload only shows the core properties, not all properties anymore:

```json
"properties": {
    "start_datetime": "2018-02-12T23:20:50Z",
    "end_datetime": "2018-02-12T23:20:50Z",
    "title": "Feature title"
},
```

We decided to still go with the incorrect OpenAPI schema because it makes it more explicit.

### DIFF
--- a/spec/components/schemas.yaml
+++ b/spec/components/schemas.yaml
@@ -1130,41 +1130,39 @@ components:
                 minItems: 2
                 items:
                   type: number
-    itemProperties:
-      title: Properties
-      description: >-
-        Provides the core metadata fields plus extensions
-
-
-        The item's data timing information can be specified either with
-
-        * One datetime value in the field `datetime`
-
-        * A datetime range with a `start_datetime` and an `end_datetime`
-
-
-        One of the two is required.
-
-
-        **Note on STAC extensions:** When a STAC extension is listed in the `stac_extensions` field,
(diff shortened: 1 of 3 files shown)


boecklic (approved):
Not 100% ideal to have an "incorrect" spec, but I have no better idea right now, let's proceed with this.

PR #663, opened 2026-09-07
PB-1677: sort extension - Add sortby query param

asteiner-swisstopo:
This allows sorting items with a `sortby` query parameter as defined in the [sort extension](https://github.com/stac-api-extensions/sort):

```txt
/collections/my-collection/items?sortby=created,-title
```

The following properties are sortable:

- id
- collection
- datetime
- title
- created
- updated

As we don't support Collection Search, the only affected endpoints are:

- [GET `/collections/{collectionId}/items`](https://data.geo.admin.ch/api/stac/static/spec/v1/apitransactional.html#tag/Data/operation/getFeatures)
- [GET `/search`](https://data.geo.admin.ch/api/stac/static/spec/v1/apitransactional.html#tag/STAC/operation/getSearchSTAC) with simple filtering
- [POST `/search`](https://data.geo.admin.ch/api/stac/static/spec/v1/apitransactional.html#tag/STAC/operation/postSearchSTAC) with full-featured filtering

Queries in practice look like this:

```sh
# GET /collections/{collectionId}/items
curl -X GET \
  "http://127.0.0.1:8000/api/stac/v1/collections/test-collection/items?sortby=datetime"

# GET /search
curl -X GET \
  "http://127.0.0.1:8000/api/stac/v1/search?ids=test-feature-1,test-feature-2&sortby=datetime"

# POST /search
curl -X POST \
  "http://127.0.0.1:8000/api/stac/v1/search" \
  -H "Content-Type: application/json" \
  -H "Authorization: Token my-token" \
  -d '{
    "query": {
      "title": {
        "startsWith": "test-description-A"
      }
    },
    "sortby": [
      {
        "field": "datetime",
        "direction": "asc"
      }
    ]
  }'
```

Notes:

- Fields are referred to as, e.g., `created` and not `properties.created`. I would find `properties.created` better but the "query" parameter in the /search endpoint also refers to the properties without the `properties.` prefix. So I favored consistency.
- I extracted tests connected to a STAC extension from `test_search_endpoint.py` to `test_search_endpoint_extensions.py` because the file got too long.

Follow-up work:

- Compatibility with STAC Browser
   - A `/sortables` endpoint listing the sortable properties
   - Advertising the sort extension in the `/conformance` endpoint
- A `/sortables` endpoint that takes into account which `stac_extensions` are declared. For the moment, only core properties are supported.


app/stac_api/pagination.py
```diff
@@ -171,7 +171,25 @@ def get_page_size(self, request):
         )
 
 
-class GetPostCursorPagination(CursorPagination):
+class SortedCursorPagination(CursorPagination):
+    '''Pagination class that supports sorting via the sortby parameter'''
+    ordering = 'name'
```
boecklic, line 176: is this declaration and initialisation correct? below it's defined as an array (which by default is empty), here it's a string..?

app/stac_api/utils.py
```diff
@@ -629,3 +630,125 @@ def parse_cache_control_header(cache_control_header):
...
+            raise ValidationError(
+                f"Invalid sort direction '{direction}'. "
+                f"Allowed values are: 'asc', 'desc'"
+            )
+        internal_field = _resolve_sort_field(field_name, sortable_fields)
+        sort_fields.append((internal_field, is_ascending))
+    return sort_fields
+
+
+def _resolve_sort_field(field_name, sortable_fields):
```
boecklic, line 734: the other helper functions are not `_*` prefixed... why this one?

boecklic (approved):
Nice 🎉 

Adding `/sortables` will be in a separate PR I guess?

app/stac_api/pagination.py
asteiner-swisstopo: Good catch but it should be fine. According to the [DRF docs](https://www.django-rest-framework.org/api-guide/pagination/#details-and-limitations:~:text=ordering%20%3D%20%27name) both works:

> - `ordering` = This should be a string, or list of strings, indicating the field against which the cursor based pagination will be applied. For example: `ordering = 'slug'`. Defaults to `-created`. This value may also be overridden by using `OrderingFilter` on the view.

app/stac_api/utils.py
asteiner-swisstopo: Because it is only used within the same module.

According to [PEP8](https://peps.python.org/pep-0008/#naming-conventions):

> - `_single_leading_underscore`: weak “internal use” indicator. E.g. `from M import *` does not import objects whose names start with an underscore.

The other new functions are used elsewhere, so not just used "internally".

I think that is a common way to mark helper functions, so I keep it as it is.

PR #664, opened 2026-09-09
PB-1677: sort extension - Add /sortables endpoints

asteiner-swisstopo:
Follow-up of #663: This adds the `/sortables` endpoints of the [sort extension](https://github.com/stac-api-extensions/sort):
- `/sortables`: Global search
- `/collections/{collectionId}/items/sortables`: Item search on a specific collection

The response is of media type `application/schema+json`, for example:

```json
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "$id": "https://data.geo.admin.ch/api/stac/v1/sortables",
  "title": "Sortables",
  "type": "object",
  "properties": {
    "id": {
      "type": "string"
    },
    "collection": {
      "type": "string"
    },
    "datetime": {
      "type": "string",
      "format": "date-time"
    },
    "title": {
      "type": "string"
    },
    "created": {
      "type": "string",
      "format": "date-time"
    },
    "updated": {
      "type": "string",
      "format": "date-time"
    }
  },
  "additionalProperties": false
}
```

These endpoints are advertised in the links sections of the landing page (`/`) and of the collection (`/collections/{collectionId}`):

```json
"links": [
{
    "rel": "http://www.opengis.net/def/rel/ogc/1.0/sortables",
    "href": "http://127.0.0.1:8000/api/stac/v1/sortables",
    "type": "application/schema+json",
    "title": "Sortable fields for the sortby parameter"
},
```

Finally, the conformance classes of the sort extension were added as well to `/conformance`:

- For the `sortby` query param:
    - https://api.stacspec.org/v1.1.0/item-search#sort
    - https://api.stacspec.org/v1.1.0/ogcapi-features#sort
- For the `/sortables` endpoints:
    - https://api.stacspec.org/v1.1.0/item-search#sortables
    - http://www.opengis.net/spec/ogcapi-features-5/1.0/conf/sortables

## STAC Browser

All that enables [recent versions of the STAC Browser](https://github.com/radiantearth/stac-browser) (v5.1.0) use this to [dynamically adapt their sorting functionality](https://github.com/radiantearth/stac-browser/blob/main/src/components/SearchFilter.vue#L385):

image

Note that for  our old version of the STAC Browser ([v3.3.4](https://github.com/geoadmin/stac-browser/)), the `/sortables` endpoint is not taken into account. Instead a hardcoded list of properties [is assumed](https://github.com/geoadmin/stac-browser/blob/b85ab977359cc2648afed95703c33d2a29313124/src/components/SearchFilter.vue#L267-L274):

- `id`
- `properties.title`
- `properties.datetime`

So the sorting dropdown appears but it shows the wrong properties (and sends the wrong requests):

image

Therefor, we probably want to sync this with the update of the STAC browser being worked on in PB-2381 (CC @adk-swisstopo).

To be clarified: Our old STAC Browser shows for Search locally:

image

even though the landing page reports https://api.stacspec.org/v1.0.0/item-search in `conformsTo`. [On PROD](https://data.geo.admin.ch/browser/index.html#/search), however, Search is displayed as expected:

image

What is also confusing: It talks to the v0.9 version (/api/stac/v0.9) for the conformance.

## OpenAPI specs

I have added the new endpoints to a different category ("tag"): Sort Extension. This is what the sort extension suggests in their[ openapi.yaml](https://github.com/stac-api-extensions/sort/blob/main/openapi.yaml). I also adopted some descriptions from there.

image

The [specs for the /conformance endpoint](https://data.geo.admin.ch/api/stac/static/spec/v1/apitransactional.html#tag/Capabilities/operation/getConformanceDeclaration) were outdated: I added the missing conformance classes.

## Implementation details

Technical details:

1. **`SORTABLE_FIELDS`**: This just used to map only external field name (e.g., `collection`) to Django model name (e.g. `collection__name`). Now it also includes info for the `/sortables` JSON schema. This makes sure that the allowed fields for the `sortby` parameter are the same as the ones reported by `/sortables`.
2. **`properties.*` prefix**: In #663, we chose to omit the `properties.` prefix for the `sortby` query parameter. So we refer to `datetime`, not `properties.datetime`. This was because the "query" parameter of the [POST /search endpoint](https://data.geo.admin.ch/api/stac/static/spec/v1/apitransactional.html#tag/STAC/operation/postSearchSTAC) also omits the `properties.` prefix. Looking into the STAC Browser code, however, I see that the `properties.` prefix is the assumed default. I think then it would be better to follow the STAC Browser. The Query extension will anyway be replaced by the Filter extension (see PB-2392). I address this in a small follow-up PR.
3. **`SchemaJSONRenderer`**: As we advertise the /sortables endpoints with `"type": "application/schema+json"`, the STAC Browser also tries to fetch the
JSON schema with schema+json. However, the STAC API cannot cope with this Accept header (`Accept: application/schema+json`) and returns an error:

   ```log
   Request GET /api/stac/v1/collections/test-collection/sortables
   Response 406: {'detail': ErrorDetail(string='Could not satisfy the request Accept header.', code='not_acceptable')}
   ```

   Therefore, we now also accept and render media type "application/schema+json".

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,6 +60,7 @@ nose2-junit.xml
 # IDE config files
 *.sublime-*
 *.code*
+*.zed
 
 # Makefile
 .timestamps
--- a/app/stac_api/migrations/0083_add_sort_conformance_classes.py
+++ b/app/stac_api/migrations/0083_add_sort_conformance_classes.py
@@ -1,0 +1,35 @@
+from django.db import migrations
+
+SORT_CONFORMANCE_CLASSES = [
+    'https://api.stacspec.org/v1.1.0/item-search#sort',
+    'https://api.stacspec.org/v1.1.0/ogcapi-features#sort',
+    'https://api.stacspec.org/v1.1.0/item-search#sortables',
+    'http://www.opengis.net/spec/ogcapi-features-5/1.0/conf/sortables',
+]
+
+
+def update_conformance(apps, schema_editor):
+    LandingPage = apps.get_model("stac_api", "LandingPage")
+    lp = LandingPage.objects.get(version='v1')
+    for conformance_class in SORT_CONFORMANCE_CLASSES:
+        if conformance_class not in lp.conformsTo:
+            lp.conformsTo.append(conformance_class)
+    lp.save()
+
+
+def reverse_update_conformance(apps, schema_editor):
+    LandingPage = apps.get_model("stac_api", "LandingPage")
+    lp = LandingPage.objects.get(version='v1')
+    for conformance_class in SORT_CONFORMANCE_CLASSES:
+        if conformance_class in lp.conformsTo:
+            lp.conformsTo.remove(conformance_class)
+    lp.save()
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("stac_api", "0082_alter_collection_stac_extensions_enabled_and_more"),
+    ]
+
+    operations = [migrations.RunPython(update_conformance, reverse_update_conformance)]
--- a/app/stac_api/serializers/general.py
+++ b/app/stac_api/serializers/general.py
@@ -110,14 +110,14 @@ class LandingPageSerializer(serializers.ModelSerializer):
                 ("href", get_url(request, 'search-list')),
                 ("rel", "search"),
                 ("method", "GET"),
-                ("type", "application/json"),
+                ("type", "application/geo+json"),
                 ("title", "Search across feature collections"),
             ]),
             OrderedDict([
                 ("href", get_url(request, 'search-list')),
                 ("rel", "search"),
                 ("method", "POST"),
-                ("type", "application/json"),
+                ("type", "application/geo+json"),
                 ("title", "Search across feature collections"),
             ]),
         ]
--- a/app/stac_api/serializers/general.py
+++ b/app/stac_api/serializers/general.py
@@ -121,6 +121,12 @@ class LandingPageSerializer(serializers.ModelSerializer):
                 ("title", "Search across feature collections"),
             ]),
             OrderedDict([
+                ("rel", "http://www.opengis.net/def/rel/ogc/1.0/sortables"),
+                ("href", get_url(request, 'sortables')),
+                ("type", "application/schema+json"),
+                ("title", "Sortable fields for the sortby parameter"),
+            ]),
+            OrderedDict([
                 ("href", get_browser_url(request, 'browser-catalog')),
                 ("rel", "alternate"),
                 ("type", "text/html"),
--- a/app/stac_api/serializers/general.py
+++ b/app/stac_api/serializers/general.py
@@ -70,7 +70,7 @@ class LandingPageSerializer(serializers.ModelSerializer):
         # We use OrderedDict, although it is not necessary, because the default serializer/model for
         # links already uses OrderedDict, this way we keep consistency between auto link and user
         # link
-        representation['links'][:0] = [
+        links = [
             OrderedDict([
                 ('rel', 'root'),
                 ('href', get_url(request, 'landing-page')),
@@ -120,17 +120,23 @@ class LandingPageSerializer(serializers.ModelSerializer):
                 ("type", "application/json"),
                 ("title", "Search across feature collections"),
             ]),
-            OrderedDict([
-                ("rel", "http://www.opengis.net/def/rel/ogc/1.0/sortables"),
-                ("href", get_url(request, 'sortables')),
-                ("type", "application/schema+json"),
-                ("title", "Sortable fields for the sortby parameter"),
-            ]),
+        ]
+        if is_api_version_1(request):
+            links.append(
+                OrderedDict([
+                    ("rel", "http://www.opengis.net/def/rel/ogc/1.0/sortables"),
+                    ("href", get_url(request, 'sortables')),
+                    ("type", "application/schema+json"),
+                    ("title", "Sortable fields for the sortby parameter"),
+                ])
+            )
+        links.append(
             OrderedDict([
                 ("href", get_browser_url(request, 'browser-catalog')),
                 ("rel", "alternate"),
                 ("type", "text/html"),
                 ("title", "STAC Browser"),
-            ]),
-        ]
+            ])
+        )
+        representation['links'][:0] = links
         return representation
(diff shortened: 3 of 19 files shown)


adk-swisstopo:
s/PB-1550/PB-2381/

Also can you confirm which commit/version actually adds the feature you want? I think that was implemented in https://github.com/radiantearth/stac-browser/commit/eb415247b4b991372acbe1301311609ec210fd3a which is in v5.0.0? Is there anything specific we need in v5.1.0 that is not in v5.0.0? Not that I see any specific reason to hold back to v5.0.0 either.

asteiner-swisstopo:
> Also can you confirm which commit/version actually adds the feature you want? I think that was implemented in [radiantearth/stac-browser@eb41524](https://github.com/radiantearth/stac-browser/commit/eb415247b4b991372acbe1301311609ec210fd3a) which is in v5.0.0? Is there anything specific we need in v5.1.0 that is not in v5.0.0? Not that I see any specific reason to hold back to v5.0.0 either.

Yes, it's that one.

I am not sure I understand which version supports it:

- In [radiantearth/stac-browser@eb41524](https://github.com/radiantearth/stac-browser/commit/eb415247b4b991372acbe1301311609ec210fd3a) it is tagged with v5.1.0.
-  In the release notes of [v5.0.0-beta.1](https://github.com/radiantearth/stac-browser/releases/tag/v5.0.0-beta.1) it says "Support for Sortables"

I only tried v5.1.0.

boecklic (approved):
🎉 nice!

asteiner-swisstopo:
@boecklic Update on what was changed:

1. Changed Media type of the /search link relation from `application/json` to `application/geo+json`. This makes the item search work again in the STAC Browser (old and new). See [this commit message](https://github.com/geoadmin/service-stac/pull/664/commits/5025db33b084d7bda7ccbcb163563499f38a0f27) for why it works on PROD, it's quite a confusing situation. Maybe to be discussed with @hansmannj who tried to revert that browser-side fix but then didn't (in PB-1550).
2. Use `properties.datetime` instead of `datetime`, see [this commit](https://github.com/geoadmin/service-stac/pull/664/commits/c97a6681039342f7bb2caa25f48915e2b68d0568) for details. I initially had it in a separate branch but I think it's easier for you if everything is here. I deleted that branch.

What that enables together: Use item search + sorting in all STAC browser versions.
- The old version: The STAC API matches the defaults like `properties.datetime` in the STAC 
- The new version: The STAC API correctly advertises /search with `application/geo+json`

Old version:
image

:warning: You see here that sorting is available but it has no effect: It's not sorted in ascending date. I think this is a bug in the old STAC Browser: It [sorts client-side](https://github.com/geoadmin/stac-browser/blob/main/src/components/Items.vue#L133) because the view [never passes on](https://github.com/geoadmin/stac-browser/blob/main/src/views/Search.vue#L48) the apiFilters containing the sortby.

New version:
image

:white_check_mark:  So here it works as expected, the bug in the STAC Browser seems to be fixed.

That is one more reason to update the STAC Browser in PB-2381. Feel free to merge on your convenience, I am off next week :hugs:

boecklic:
> @boecklic Update on what was changed:
> 
>     1. Changed Media type of the /search link relation from `application/json` to `application/geo+json`. This makes the item search work again in the STAC Browser (old and new). See [this commit message](https://github.com/geoadmin/service-stac/pull/664/commits/5025db33b084d7bda7ccbcb163563499f38a0f27) for why it works on PROD, it's quite a confusing situation. Maybe to be discussed with @hansmannj who tried to revert that browser-side fix but then didn't (in [PB-1550](https://swissgeoplatform.atlassian.net/browse/PB-1550)).
> 

~~The reason for having `application/json` and not `application/geo+json` was that Cloudfront automatically compresses content with type `application/json` but not `application/geo+json`. 
We'll have to be careful with that change. I think in terms of impact on compression we can probably change that without major impact, these payloads are generally small anyway (and Cloudfront compresses only above 1MB anyway). But clients might rely on the fact the response type is currently `application/json`. What does the browser send as `Accept` header? Could we make the response `Content-Type` depend on the Accept header? (so if `Accept` header is `application/geo+json` the response `Content-Type` is `application/geo+json`, if the `Accept` header is `application/json` the response `Content-Type` is `application/json`  ? what do you think @adk-swisstopo ?~~

While basically true what I wrote above it's not what's actually changed here. The relevant change is that the content type advertised in the link to the `/search` endpoint is changed. The `Content-Type` of the payload of `/search` is `application/geo+json` already now. So I think it's all good. I'll merge this PR.

[PB-1550]: https://swissgeoplatform.atlassian.net/browse/PB-1550?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

hansmannj:
> > @boecklic Update on what was changed:
> > ```
> > 1. Changed Media type of the /search link relation from `application/json` to `application/geo+json`. This makes the item search work again in the STAC Browser (old and new). See [this commit message](https://github.com/geoadmin/service-stac/pull/664/commits/5025db33b084d7bda7ccbcb163563499f38a0f27) for why it works on PROD, it's quite a confusing situation. Maybe to be discussed with @hansmannj who tried to revert that browser-side fix but then didn't (in [PB-1550](https://swissgeoplatform.atlassian.net/browse/PB-1550)).
> > ```
> 
> The reason for having `application/json` and not `application/geo+json` was that Cloudfront automatically compresses content with type `application/json` but not `application/geo+json`. We'll have to be careful with that change. I think in terms of impact on compression we can probably change that without major impact, these payloads are generally small anyway (and Cloudfront compresses only above 1MB anyway). But clients might rely on the fact the response type is currently `application/json`. What does the browser send as `Accept` header? Could we make the response `Content-Type` depend on the Accept header? (so if `Accept` header is `application/geo+json` the response `Content-Type` is `application/geo+json`, if the `Accept` header is `application/json` the response `Content-Type` is `application/json` ? what do you think @adk-swisstopo ?

That sound's like a good idea IMHO

[PB-1550]: https://swissgeoplatform.atlassian.net/browse/PB-1550?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #666, opened 2026-09-14
PB-2400: replace deprecated minio with moto-server

boecklic:
Minio is deprecated and their docker images cannot be downloaded anymore. Moto-server is replacing minio to allow local development and testing again.


docker-compose.yml
```diff
@@ -15,38 +15,40 @@ services:
...
-      - type: bind
-        source: ${PWD}/.volumes/minio
-        target: /data
+  moto-server:
+    image: motoserver/moto
+    container_name: moto-server
     ports:
-      - 9090:${S3_PORT:-9000}
-      - 9001:9001
+      - "${S3_PORT:-9000}:${MOTO_PORT:-5000}"
```
msom: `S3_PORT` referred to the container port before, now to the host port. The S3 host port was 9090 before, now it defaults to 9000. `LEGACY_AWS_S3_ENDPOINT_URL` for example is still `http://127.0.0.1:9090`.

Do we even need the any of these, or should we just use `- "9090:5000" (host 9090, container 5000 always default)  and remove these. Or does the container require `MOTO_PORT`?

msom (changes requested):
I think the port is wrong, it should be 9090, not 9000.

docker-compose.yml
boecklic: I think the default `MOTO_PORT` is `5000`, so we can leave that out, I'd leave however the host port configurable.

msom (approved):

PR #658, opened 2026-07-16
PB-2355 Improve extension handling

asteiner-swisstopo:
## Introduction

In a nutshell, this PR adds:

1. Item field `stac_extensions` and collection field `stac_extensions_enabled`
2. Validation for `stac_extensions` and `properties` based on `stac_extensions`
3. Django admin UI fields and OpenAPI documentation.

The user can now specify the `stac_extensions` in the payload when creating an Item. As an admin, similarly to the `published` field, one can enable STAC extensions for Items in a certain Collection by adjusting `stac_extensions_enabled`.

The extensions we partly support currently are:

- [Timestamps extension](https://github.com/stac-extensions/timestamps): Only field `expires`
- [Forecast extension](https://github.com/stac-extensions/forecast):
   - `forecast:reference_datetime`: required
   - `forecast:horizon`
   - `forecast:duration`
   - `forecast:variable`
   - `forecast:perturbed`
- [CF extension](https://github.com/stac-extensions/cf): Soon available with `cf:standard_name` and `unit`, see PB-2354

## Admin UI

Each collection now has the option to enable STAC extensions:

stac-extensions-django-admin-stac-extensions-enabled

You can filter by that in the collection overview:

stac-extensions-django-admin-stac-extensions-enabled-filter

Selecting "None" deselects the others. It's possible to select multiple extensions; when more than one is checked, only collections that have **all** of the selected extensions are shown.

## Validation

This has most impact on how the API behaviour changes.

### Preliminary Investigation

We discussed enforcing extension schemas using the JSON Schema of the extension, e.g., for the timestamps extension: https://stac-extensions.github.io/timestamps/v1.1.0/schema.json. There are Python libraries for that like [jsonschema](https://python-jsonschema.readthedocs.io).

The only collection using the forecast extension is [ch.meteoschweiz.ogd-forecasting-icon-ch2](https://data.geo.admin.ch/browser/index.html#/collections/ch.meteoschweiz.ogd-forecasting-icon-ch2). All items in there are compliant with the [JSON schema v0.2.0](https://stac-extensions.github.io/forecast/v0.2.0/schema.json) except for one: [Item "ch"](https://data.geo.admin.ch/api/stac/v1/collections/ch.meteoschweiz.ogd-forecasting-icon-ch2/items/ch).

We concluded that we continue with our hand-made validation for flexibility and not to change too much at once.

### Validation Steps

The validation steps carried out are:

1. `stac_extensions`
   1. The value must be one of these:
   
      ```txt
[...]

PR #668, opened 2026-09-17
PB-2355: optimize data migration

boecklic:
The original data migration was inefficient for large datavolumes and was breaking on PROD. Although it's bad practice to modify already applied migration an optimized version is provided now as reference for future, similar data migrations.


adk-swisstopo:
Is that something we could have detected before reaching prod? Maybe INT should have a DB size congruent with PROD?

Is that something we could have prevented by generating more efficient migration scripts by default? Would that be a bug/feature request against upstream?

How was the migration completed eventually? With this modified script?

boecklic:
> Is that something we could have detected before reaching prod? Maybe INT should have a DB size congruent with PROD?

yeah, at some point we should think about having the same (or at least the same amount) of data on INT than on PROD, that would certainly have helped

> 
> Is that something we could have prevented by generating more efficient migration scripts by default? Would that be a bug/feature request against upstream?

"upstream" here are we ourselves ;-). Data migrations are not generated by Django, we create them ourselves...

> 
> How was the migration completed eventually? With this modified script?

I had to "fake apply" (option `--fake`) the migration to get the service up again and then do the data migration manually once the service was functioning properly again.

boecklic:
(btw: feel free to merge this once someone has approved)

adk-swisstopo:
> "upstream" here are we ourselves ;-). Data migrations are not generated by Django, we create them ourselves...

I think I got confused between schema [migrations](https://docs.djangoproject.com/en/6.1/topics/migrations/) which are generated by Django and data migration which we apparently write by hand? The relevant data migration was introduced in https://github.com/geoadmin/service-stac/pull/658/changes/199df2b3d6d54be13fdff07a7f62d1c6c01c7de1
The `Generated by Django` comment at the top makes it extra confusing as it's hard to tell what has been machine or human-generated. Modifying it after the fact (this PR) adds to the confusion, although I agree it's better than leaving it unmodified.

> > How was the migration completed eventually? With this modified script?
> 
> I had to "fake apply" (option `--fake`) the migration to get the service up again and then do the data migration manually once the service was functioning properly again.

I see. Should we document that procedure somewhere the next person who runs into such problem has a chance to find it? Our README.md has a [section](https://github.com/geoadmin/service-stac/blob/develop/README.md#migrate-db-with-django) about migrations has an FAQ in which it would fit nicely.

app/stac_api/migrations/0079_add_stac_extensions_to_meteoschweiz_collections.py
```diff
@@ -23,7 +23,16 @@ def update_stac_extensions(apps, schema_editor):
...
+    # Note: it's bad practice to modify already applied migrations, but the original version
+    # `for item in Item.objects.filter(forecast_reference_datetime__isnull=False):`
+    # is extremely inefficient for large datasets and breaks on
+    # PROD because it loads all items into memory at once.
+    # I leave it here as a reference for future data migrations and add an
+    # optimized version that uses an iterator.
+    # (see https://docs.djangoproject.com/en/6.1/ref/models/querysets/#iterator)
+    for item in Item.objects.exclude(stac_extensions__contains=[FORECAST_EXT]
+                                    ).filter(forecast_reference_datetime__isnull=False
+                                            ).iterator(chunk_size=1000):
```
adk-swisstopo, line 35: Any particular reason for specifying the `chunk_size`? The [default](https://docs.djangoproject.com/en/6.1/ref/models/querysets/#iterator) seems to be 2000.

adk-swisstopo (approved):
What is the testing strategy for this change? I understand that without a large enough database (currently only available in PROD) it's hard to tell whether it has the intended effect. I don't have any specific reason to think it won't resolve the problem but it's easy to run into another scaling issue we haven't noticed.

I think further optimisation might be possible by using `update`/`bulk_update`, which might even let us skip the Python loops completely. Although it might not be necessary.

